PR #23, opened 2021-03-02
chore(prettier): add .prettierrc file

with-heart:
This change adds a `.prettierrc` file to the project.

From looking at the existing code and #8, the current style seems to match `prettier` defaults except for `trailingComma` which is set to `"all"`. By specifying only that override, `prettier` will respect its default values for all other rules.

This allows `plugin:prettier/recommended` to know what to recommend, and prevents anyone with a global `.prettierrc` in their local env (like me 😁) from constantly overwriting the expected style.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "trailingComma": "all"
+}


nihaals:
I haven't yet looked at prettier's config much as I've been working on actually writing components but this should be a great first change as we decide on style later on. Right now everything's been done just with VSCode's prettier and all default config (afaik).

nihaals:
Thanks!

with-heart:
Yeap, just noticed that this should actually be all defaults and `trailingComma` shouldn't have a custom value. Will fix to just stick with the defaults in a bit :)

PR #31, opened 2021-03-05
chore(prettier): reset to entirely default config

with-heart:
Ran on #8 to validate that it matches exactly. Should've done that to start 😬


nihaals:
I haven't looked at all the prettier options but this one at a first glance I definitely prefer (`es5` over `all`)

nihaals:
Thanks!

PR #32, opened 2021-03-05
chore: add pre-commit hook for prettier + eslint

### LINKED ISSUE #30: Set up husky+lint-staged+prettier

with-heart:
Closes #30 

This change adds a pre-commit hook via `husky` and `lint-staged`, which enables the configured tasks to be run on staged changes. With this configuration, `prettier` and `eslint` are run on their respective files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 /node_modules/
 **/yarn-error.log
-.eslintcache
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules/
 **/yarn-error.log
+.eslintcache
--- a/package.json
+++ b/package.json
@@ -20,8 +20,8 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
-    "husky": "=4",
-    "lint-staged": ">=10",
+    "husky": "^4.3.8",
+    "lint-staged": "^10.5.4",
     "prettier": "^2.2.1",
     "typescript": "^4.2.3"
   },
--- a/package.json
+++ b/package.json
@@ -32,6 +32,6 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx,md,mdx,yml,json}": "prettier --write",
-    "*.{js,jsx,ts,tsx}": "eslint --cache --fix"
+    "*.{js,jsx,ts,tsx}": "eslint --cache --cache-location node_modules/.cache/.eslintcache --fix"
   }
 }
--- a/package.json
+++ b/package.json
@@ -20,7 +20,18 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
+    "husky": "=4",
+    "lint-staged": ">=10",
     "prettier": "^2.2.1",
     "typescript": "^4.2.3"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,md,mdx,yml,json}": "prettier --write",
+    "*.{js,jsx,ts,tsx}": "eslint --cache --fix"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2517,7 +2517,7 @@ human-signals@^1.1.1:
   resolved "https://registry.yarnpkg.com/human-signals/-/human-signals-1.1.1.tgz#c5b1cd14f50aeae09ab6c59fe63ba3395fe4dfa3"
   integrity sha512-SEQu7vl8KjNL2eoGBLF3+wAjpsNfA9XMlXAYj/3EdaNfAlxKthD1xjEQfGOUhllCGGJVNY34bRr6lPINhNjyZw==
 
-husky@=4:
+husky@^4.3.8:
   version "4.3.8"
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -20,7 +20,18 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
+    "husky": "=4",
+    "lint-staged": ">=10",
```
nihaals, line 24: What's the advantage of using `>=` instead of a compatible range (`^`)?

package.json
```diff
@@ -20,7 +20,18 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
+    "husky": "=4",
```
nihaals, line 23: Is there a reason to avoid husky 5?

.gitignore
```diff
@@ -1,2 +1,3 @@
 /node_modules/
 **/yarn-error.log
+.eslintcache
```
nihaals, line 3: Maybe we could follow CRA (https://github.com/facebook/create-react-app/pull/9977) and move it into node_modules? It would also be hidden from the IDE's file explorer if `node_modules` is already hidden (which it is in the default `.vscode/settings.json` in the repo)

nihaals (changes requested):

.gitignore
with-heart: I'm down with that!

package.json
with-heart: ~I'm not actually not sure why either of these are set up this way.~ I generated this using `npx mrm lint-staged` as recommended [in the `prettier` docs](https://prettier.io/docs/en/precommit.html) to save time. This issue explains what's going on, at least for `husky`: https://github.com/sapegin/mrm/issues/145

I'll switch `lint-staged` to use compatible range.

package.json
with-heart: See above comment

with-heart: Also fine with upgrading to v5 if we want, just haven't taken the time to investigate the new setup to determine how to automatically install the hooks for any new contributors.

nihaals: I think we can separate this into a separate issue (and PR in the future) since it's not a big deal and contributors are probably more familiar with v4 at the moment

package.json
```diff
@@ -20,7 +20,18 @@
...
     "prettier": "^2.2.1",
     "typescript": "^4.2.3"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,md,mdx,yml,json}": "prettier --write",
```
nihaals, line 34: `yaml` isn't included here. Is this intentional (and trying to set a standard of using `yml`?

nihaals (changes requested):
One thing I "like" (it might be just that I'm used to it, it can be annoying though when you're making multiple commits in one go by staging parts of files) when using [pre-commit](https://github.com/pre-commit/pre-commit) is that when a formatter makes a change, it aborts the commit and you have to stage the new changes before committing again, while this changes the staged file (not sure how this works, maybe just immediately staging the lines that were affected?) and doesn't neccesarily make it clear it made changes before letting you enter a commit message (assuming no `-m`). This causes some slightly weird things when staging only parts of files when you might want more control. I'm undecided if this should block this though, especially with husky being the standard tool for this (not sure if lint-staged is responsible for this behaviour or husky). My tests weren't super realistic though, this may be a non issue in a more standard file.

package.json
with-heart: Nope, I'll add `yaml`. Was just following the pattern that the only yaml files in the project currently are `yml` but they're both totally fine.

with-heart:
My personal workflow utilizes heavy partial staging for atomic commits, and I haven't noticed an issue with this on any of the projects that I contribute to with the same or similar setups. From what I could find in the `lint-staged` issues, they added support for partial staging quite some time ago. With that said, I've never directly verified that it works, so maybe it has been subtly committing the wrong things for a long time and I've just never noticed 🙃 Worth testing a little more!

MythicManiac:
I would feel more comfortable using [pre-commit](https://pre-commit.com/), as that's what I have experience with and know the ecosystem around (not to mention it's already used in the backend repo), but I'm also not opposed to using husky if it's what's better known in the javascript world.

That being said, I do have some questions left unanswered after skimming through husky docs:
- Can we run this on _all files_ in the repo? With pre-commit, you can do `pre-commit run --all-files`, which makes it easy to run checks on the CI runner.
- What are the hooks based on? Shell scripts? Javascript? Interestingly enough the documentation doesn't seem to mention how would you actually implement hooks if you needed to
- What kind of an ecosystem is there around husky? Are there useful hooks that we could easily depend on, or is it more common to roll your own?

One more thing that I find particularly annoying is if [this workaround](https://typicode.github.io/husky/#/?id=yarn-on-windows) is an actual requirement

nihaals:
> - Can we run this on _all files_ in the repo? With pre-commit, you can do `pre-commit run --all-files`, which makes it easy to run checks on the CI runner.

We could probably add a `lint` script to the `package.json` that just runs prettier and eslint. We could then add one for writing the changes and one which just runs as a check to support different use cases.

> One more thing that I find particularly annoying is if [this workaround](https://typicode.github.io/husky/#/?id=yarn-on-windows) is an actual requirement

Not sure if you can just use cmd or PowerShell but if not we could recommend WSL since that seems to be what a lot of devs on Windows are using now, especially with the WSL support in VSCode. Also #38 would also help but users wouldn't be able to use a git GUI other than VSCode's if they're used to it (although maybe they support using a custom shell?). Even then you could use the GUI, just not for actually committing. My workflow at the moment is a mix between `git add`, VSCode's VCS and GitKraken to stage changes then I always use `git commit` myself (partially due to my PGP key's passphrase being saved only when running it myself).

PR #37, opened 2021-03-05
chore: use preconstruct to manage monorepo

with-heart:
This PR makes a few changes to the project in order to use [`preconstruct`](https://preconstruct.tools/) for managing the monorepo. `preconstruct` simplifies a few things that are typically painful in a monorepo, and I wish we were using this for Chakra, so I thought I'd propose it early here. 

## Changes
- Moved some dependencies around
	- `react`/`react-dom` should exist in individual packages and not the base `package.json`
	- Base `package.json` is just a umbrella for the other packages and has no distinction between deps/devDeps, so we can just install all base packages as regular deps
- Added `preconstruct`
	- Configured to automatically link packages on `yarn install`
	- Responsible only for building and linking `thunderstore-components` atm
		- Adding new packages in the future is as simple as creating a new package, adding to `preconstruct.packages` in `package.json`, and running `preconstruct init`
	- Added `@preconstruct/next` plugin in `next.config.js` so next `build`/`dev` commands work the same regardless of whether we're using local dev link or prod build

### Local package links

Normally with a monorepo, you link packages for local dev, but the consumed package needs to be rebuilt anytime it changes in order for those changes to make it to consuming packages (like `nextjs` for example). This is a real hassle and requires using `watch`, which can be annoying if you have more than one consumed package you're working on at once.

`preconstruct dev` analyzes the project to link packages in the right way, and also generates a shim `dist` for each package it's responsible for that just points to the source folder, so there's no need to build the package for local dev work.

### Build process

`preconstruct build` handles the build process for all packages we make it responsible for, using a sane `rollup` config and `babel`. This takes a significant amount of effort out of the build configuration process, as we can simply install the `babel` plugins/presets/transforms we want to use and allow `preconstruct` to handle the rest. Adding new buildable packages requires no additional effort.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,5 +27,14 @@
   ],
   "rules": {
     "react/react-in-jsx-scope": "off"
-  }
+  },
+  "overrides": [
+    {
+      "files": "*.js",
+      // rule overrides for js config files (next.config.js, etc.)
+      "rules": {
+        "@typescript-eslint/no-var-requires": "off"
+      }
+    }
+  ]
 }
--- a/README.md
+++ b/README.md
@@ -1,0 +1,13 @@
+# Thunderstore UI
+
+## Monorepo Setup
+
+- [`yarn` workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) for
+  managing the packages in the monorepo (see `packages` key in base
+  `package.json` file)
+- [`preconstruct`](https://preconstruct.tools/) tool takes a lot of the pain
+  out of managing a monorepo
+  - code behaves the same in dev as it will in production
+  - packages linked locally via `preconstruct dev` (runs automatically in
+    `postinstall` hook) without needing to manually build anything
+  - `nextjs` workspace uses `@preconstruct/next` plugin to integrate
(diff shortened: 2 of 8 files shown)


.eslintrc.json
```diff
@@ -27,5 +27,14 @@
...
+  },
+  "overrides": [
+    {
+      "files": "*.js",
+      // rule overrides for js config files (next.config.js, etc.)
+      "rules": {
+        "@typescript-eslint/no-var-requires": "off"
+      }
+    }
+  ]
```
with-heart, line 39: Allows us to use `require` statements in `.js` files, which I imagine will mostly be configuration files for this project.

nihaals (changes requested):
I think we should have some developer documentation as based on just looking at the diff I wasn't confident on how to actually have the same setup I had before (running Next.js' dev server and have the components auto build). After reading https://github.com/thunderstore-io/thunderstore-ui/pull/37#issue-585793001, I guessed just doing `cd nextjs` and `yarn run dev` but this didn't work (after modifying `PackageUpload.tsx`, the change was not reflected).

PR #84, opened 2021-04-04
Restructure project and add Storybook

MythicManiac:


packages/components/src/components/PackageUpload.tsx
```diff
@@ -139,9 +139,15 @@ const PackageUploadForm: React.FC<PackageUploadFormProps> = ({
...
-                    )}
+                    {errors.communities &&
+                      errors.communities.map(
+                        (value, index) =>
+                          value && (
+                            <FormErrorMessage key={index}>
+                              {value.message}
+                            </FormErrorMessage>
+                          )
+                      )}
```
nihaals, line 150: This should check if `errors.communities.*.message` is an actual value

packages/nextjs/next.config.js
```diff
@@ -0,0 +1,14 @@
...
+        rule.include = [
+          ...rule.include,
+          // eslint-disable-next-line @typescript-eslint/no-var-requires
+          require("path").join(__dirname, ".."),
+        ];
+      }
+    });
+    return config;
+  },
+};
```
nihaals, line 14: Why is this needed? Some comments would help explain (I'm assuming it's related to importing from the components library)

packages/nextjs/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@thunderstore/nextjs",
+  "version": "0.1.0",
+  "private": true,
+  "repository": "https://github.com/thunderstore-io/thunderstore-ui/tree/master/packages/nextjs",
```
nihaals, line 5: From the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository):
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

A relevant example:
```json
{
  "repository": {
    "type": "git",
    "url": "https://github.com/facebook/react.git",
    "directory": "packages/react-dom"
  }
}
```

packages/nextjs/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@thunderstore/nextjs",
+  "version": "0.1.0",
```
nihaals, line 3: Is this intentionally `0.1.0` or should it be unversioned (`0.0.0`)?

packages/nextjs/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "build": "next build",
+    "start": "next start",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@thunderstore/components": "*",
+    "next": "^10.1.3",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
+  }
```
nihaals, line 17: Shouldn't some of the dev dependencies like `@types/react` be duplicated here?

package.json
```diff
@@ -2,25 +2,17 @@
   "private": true,
   "repository": "https://github.com/thunderstore-io/thunderstore-ui",
```
nihaals, line 3: From the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository):
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

This should probably end in `.git`

packages/components/package.json
```diff
@@ -1,13 +1,14 @@
 {
-  "name": "thunderstore-components",
+  "name": "@thunderstore/components",
   "version": "0.1.0",
   "description": "Shared components for Thunderstore",
-  "repository": "https://github.com/thunderstore-io/thunderstore-ui/tree/master/thunderstore-components",
-  "main": "./dist/index.js",
+  "repository": "https://github.com/thunderstore-io/thunderstore-ui/tree/master/packages/components",
```
nihaals, line 5: From the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository):
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

A relevant example:
```json
{
  "repository": {
    "type": "git",
    "url": "https://github.com/facebook/react.git",
    "directory": "packages/react-dom"
  }
}
```

packages/storybook/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@thunderstore/storybook",
+  "version": "0.1.0",
+  "private": true,
+  "repository": "https://github.com/thunderstore-io/thunderstore-ui/tree/master/packages/storybook",
```
nihaals, line 5: From the [docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository):
> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It's for computers.

A relevant example:
```json
{
  "repository": {
    "type": "git",
    "url": "https://github.com/facebook/react.git",
    "directory": "packages/react-dom"
  }
}
```

package.json
```diff
@@ -2,25 +2,17 @@
...
-    "thunderstore-components"
+    "packages/*"
   ],
-  "dependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
-  },
   "devDependencies": {
     "@microsoft/eslint-formatter-sarif": "^2.1.5",
-    "@types/node": "^14.14.37",
```
nihaals, line 14: This package isn't in any of the `package.json`s anymore

packages/components/src/components/PackageUpload.tsx
```diff
@@ -139,9 +139,15 @@ const PackageUploadForm: React.FC<PackageUploadFormProps> = ({
                       defaultValue={null}
                       control={control}
                     />
-                    {errors.communities?.message && (
-                      <FormErrorMessage>{errors.communities.message}</FormErrorMessage>
-                    )}
+                    {errors.communities &&
+                      errors.communities.map(
```
nihaals, line 143: This gives me `TypeError: errors.communities.map is not a function` if I press `Upload` without entering any values (as it is a string, not a list of strings)

packages/components/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const isDevelopment = process.env["NODE_ENV"] === "development";
+export const isProduction = process.env["NODE_ENV"] === "production";
```
nihaals, line 2: What's wrong with `process.env.NODE_ENV`?

packages/nextjs/pages/_app.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { RootWrapper, theme } from "thunderstore-components";
+import { RootWrapper, theme } from "@thunderstore/components";
 
 function MyApp({ Component, pageProps }: AppProps): JSX.Element {
   return (
     <RootWrapper
       theme={theme}
       thunderstoreProviderValue={{
-        apiUrl: process.env.NEXT_PUBLIC_API_URL || "https://thunderstore.io/api/",
+        apiUrl: process.env["NEXT_PUBLIC_API_URL"] || "https://thunderstore.io/api/",
```
nihaals, line 9: What's wrong with `process.env.NEXT_PUBLIC_API_URL`?

nihaals (changes requested):
The `dependabot.yml` file might be outdated now, I'm not sure if it checks yarn workspaces and updates those correctly, especially with some packages being in multiple `package.json`s

GitHub says that `ssri=6.0.1` is used which has a [vulnerability](https://github.com/advisories/GHSA-vx3p-948g-6vhq). There are also mentions of `8.0.1` so I'm not sure which version is actually being used, but should definitely make sure that it's not possible to install `<8.0.1`.

I haven't looked at the `tsconfig`s yet but some of the references to `types/` might be redundant.

`typings/` is also the standard name for a directory like that.

Overall great changes and sets good foundations for the final storybook setup (e.g. msw).

package.json
```diff
@@ -2,25 +2,17 @@
   "private": true,
   "repository": "https://github.com/thunderstore-io/thunderstore-ui",
   "workspaces": [
-    "nextjs",
```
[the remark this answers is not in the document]
MythicManiac: I don't think it really matters, keeps the configuration simpler to have everything under the same rules of structuring. If anything you could rename the `packages` to something like `projects`.

package.json
MythicManiac: all of the typechecks work just fine without it at least

packages/components/src/components/PackageUpload.tsx
MythicManiac: In that case the type definition for `errors.communities` was wrong and should be looked at. I just made it pass typechecks here.

packages/components/src/constants.ts
MythicManiac: lint rule requiring indexes to be accessed as indexes

packages/nextjs/pages/_app.tsx
MythicManiac: same as above

packages/storybook/package.json
MythicManiac: Iherited the pattern from the existing two projects, but I'll change this and those to just point to the root repo here

packages/components/src/components/PackageUpload.tsx
nihaals: Yeah that's why I left it at the time. I think we might need to make an issue on react-hook-form. It's either a list of strings or a string

packages/components/src/constants.ts
nihaals: Which lint rule? I don't think I saw an ESLint error

MythicManiac: some typescript strict flag

package.json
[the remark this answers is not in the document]
nihaals: I think `projects` sounds better since we're unlikely to have more libraries and the components are mostly for our own projects. I don't have a strong preference between the two though, it's pretty minor

tsconfig.build.json
```diff
@@ -0,0 +1,36 @@
...
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
+    "strictBindCallApply": true,
+    "strictPropertyInitialization": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+
+    "noUnusedLocals": true,
+    "noUnusedParameters": false,
```
nihaals, line 24: `noUnusedParameters` should be enabled since `ts(6133)` and `@typescript-eslint/no-unused-vars` enforce no unused parameters (and there isn't really a reason to allow it anyway)

packages/components/src/components/PackageUpload.tsx
nihaals: `PackageUploadFormInputs.communities` should be changed to `NestedValue<string[]>` as "non-primitive types" should be wrapped with `NestedValue` (the docs aren't very clear). This fixes the type for `errors`

package.json
nihaals: I think we can just add this if it's needed in the future for `getServerSideProps`

tsconfig.build.json
MythicManiac: sometimes you might have a callback where you for example need the 3rd parameter, and IMO it's nicer to just name them out instead of having `_, _, something` for example, which is why I figured this could be disabled.

anttimaki:
Replaced by PR #289 .

PR #285, opened 2021-11-29
Use Preconstruct to manage monorepo

anttimaki:
This PR intends to do the same things as PR #37. I initially started to work with it, rebasing it to current master etc., but I encountered one error after another and finally decided to start from the scratch.

Basically I followed [this tutorial](https://monorepo.guide/getting-started) and fixed any problems I detected. Some of them were already solved in the earlier PR, so the solutions look somewhat similar.

Closes #37


MythicManiac (approved):
Looks good to me, nice work 👍 

Should mention merging this closes #37 (in the PR description) so that indeed happens when this is merged.

PR #289, opened 2021-12-02
Use Storybook to create reusable UI components in isolation

anttimaki:
This PR does much of the same stuff as PR #84. Includes:

* Configuration for [Storybook](https://storybook.js.org/)
* Short description in README about using Storybook
* Simple exemplary stories for `Select`, `MultiSelect` and `Markdown` components

Included in PR #84, but **not** included here:

* Project directory restructuring (this was already done in PR #285)
* TypeScript config tuning (since it's not related to Storybook)

Closes PR #84.


apps/storybook/.babelrc.json
```diff
@@ -0,0 +1,61 @@
...
+    [
+      "babel-plugin-polyfill-corejs3",
+      {
+        "method": "usage-global",
+        "absoluteImports": "core-js",
+        "version": "3.19.1"
+      }
+    ]
+  ]
+}
```
MythicManiac, line 61: pre-commit not handling missing newlines in this repo? Should we perhaps add pre-commit or something similar (in another PR)?

MythicManiac (approved):
Looks good to me 👍 merge at will

apps/storybook/.babelrc.json
anttimaki: Well there is this long-forgotten PR: #32 .

There's also seems to be a pre-commit config in the repo, but I've not set it up yet. Should we use pre-commit as we do in our other repos and ignore that old PR?

MythicManiac: Yeah that sounds like a good idea

PR #304, opened 2021-12-07
Linking poc 2

anttimaki:


packages/components/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const thunderstoreLinkProps: ThunderstoreLinkProps = {
+  community: "",
+  package: "",
+};
+
+// Accepting any and all props is required to keep the linking
+// implementation-agnostic. E.g. an implementation with Chakra has all
+// kinds of custom properties.
+interface AnyProps {
+  [key: string]: string | number | boolean;
```
MythicManiac, line 30: `object` not needed?

MythicManiac: Moreover why not just `any`?

anttimaki: I live in constant fear of my immortal soul being eternally punished for invoking the forbidden `any`.

PR #332, opened 2022-01-24
Add package detail view

anttimaki:


anttimaki:
This PR contains quite a lot of TODOs:

- The page uses fake data
- Due to change in layout width, the `PackageActions` component's layout was changed, which leads in a gap below the `PackageHeader` component on full width page. Current idea is we might place an add banner there, but it's not implemented yet
- `PackageActions` currently only displays package's id string. User should be able to copy it to clipboard easily, but we're not sure how to implement this UI-wise yet
- `PackageActions` currently links to dependants view with an element that looks like button. This is probably not the UI we want, but it should be decided together with the id string copying mentioned above
- The links in `PackageRequirements` currently assume that each dependency is included in the "active" community's packages, but this is not guaranteed. As a fallback, we should link to community-agnostic package view, or select one of the communities where the package is listed and link there
- [Chakra UI's typings are currently broken](https://github.com/chakra-ui/chakra-ui/issues/5370), so we're using inline styles instead of `Text` components `variant` prop, which would be nice and DRY. The code should be updated once the problem is fixed.

packages/hooks/src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    // Set initial state, which will always be false for the initial,
+    // server-side rendering.
+    checkMatch();
+
+    // Update the state whenever the component is (re)rendered in browser.
+    window.addEventListener("resize", checkMatch);
+    return () => window.removeEventListener("resize", checkMatch);
+  }, []);
```
MythicManiac, line 21: This should include a dependency for `checkMatch`

MythicManiac (commented):
Overall looks good, just a minor technicality

packages/hooks/src/hooks/useMediaQuery.ts
anttimaki: The idea was to run the effect only once, hence no dependencies. Although, `checkMatch` being defined as it is, I'm not sure it will trigger reruns anyway. Based on a quick test, adding `checkMatch` to dependencies works as well, but it was hard to say whether this caused extra renders or not. Thoughts?

MythicManiac: It should functionally be the same, but since `checkMatch` is used in the effect, proper practice is to include it in the dependencies too (some linters even notify about this). I don't think it matters a lot in practice here, but  it is more future proof in case the hook evolves into something more complex.

anttimaki: [Fixed](https://github.com/thunderstore-io/thunderstore-ui/pull/332/commits/d7b58111663d7dd9efc13fb316bd521733f6aef7#diff-c603c0226e1474d16fae89b64097f6f7a19fff80b1c0dd728d6ccdf1c7cc2d6dR21)

PR #338, opened 2022-01-27
Update @chakra-ui/react to 1.8.0

anttimaki:
Commit 3c0e8e4 introduced a temporary version of @chakra-ui/react
which partially fixed type-related issues. The newest release includes
these fixes, and also fixes the problems with typing variant prop, so
related temporary fixes can now be removed from codebase.

However, this introduces a new issue, but it's such a minor one that
updating now seems reasonable:
https://github.com/chakra-ui/chakra-ui/issues/5453

### DIFF
--- a/apps/nextjs/package.json
+++ b/apps/nextjs/package.json
@@ -9,7 +9,7 @@
     "start": "next start"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


MythicManiac (approved):
Nice 👍

PR #345, opened 2022-02-03
Add Dockerfile for Storybook app

anttimaki:


anttimaki:
This approach uses `http-server` inside the container to serve the files. I'm not aware how efficient it is performance-wise, but I assume the storybook wouldn't be receiving that much traffic anyway?

Storybook's documentation recommends an external service for hosting, but I don't know if that's something we're interested in at this point.

> Once you've built your Storybook as a static web app, you can publish it to your web host. We recommend [Chromatic](https://www.chromatic.com/), a free(*) publishing service made for Storybook that documents, versions, and indexes your UI components securely in the cloud.

MythicManiac:
Regarding serving the files in the docker container, a fairly common practice is to use an nginx container to do it, so the docker build becomes two-phased, where first there's a builder image that builds the static files, and second they and an nginx config get copied to an nginx container.

anttimaki:
Updated the [second commit](https://github.com/thunderstore-io/thunderstore-ui/pull/345/commits/7a5ec5f6663c0652dd6d6f4048c278428c66a823) to use nginx to serve the static files in Storybook container.

MythicManiac (approved):
Nice work 👍

PR #370, opened 2022-03-07
Unify terminology between backend and UI

anttimaki:


MythicManiac:
I notice a lot of the changes here are related to local dummy data, would it be beneficial to implement a system specifically for generating dummy data and re-using that? Could help at least with (future) tests as well as storybook, and would allow organizing the code a bit better (moving dummy data generation to a separate module somewhere).

Perhaps it'd make the most sense to do that with the data provider pattern, and have a dummy data provider. Some considerations also should be taken for generating data deterministically (for tests and storybook), but for local dev even random works.

packages/components/src/components/LinkingProvider.tsx
```diff
@@ -71,14 +67,6 @@ export interface LinkLibrary {
   PackageDependants: (
     props: AnyProps & { community: string; package: string }
   ) => RE | null;
-  /** Link for downloading the package */
-  PackageDownload: (
-    props: AnyProps & { namespace: string; package: string; version: string }
-  ) => RE | null;
-  /** Link for installing the package via mod manager */
-  PackageInstall: (
```
MythicManiac, line 79: Is this removal a good idea? How do we implement the install button in the mod manager, where it probably shouldn't open a link but instead run the actual action?

apps/nextjs/utils/transforms/packageCard.ts
```diff
@@ -0,0 +1,43 @@
...
+  image_src: string | null;
+  last_updated: string;
+  nsfw: boolean;
+  package_name: string;
+  pinned: boolean;
+  rating_score: number;
+  team_name: string;
+}
+
+export type PackageCardPropsWithoutOnClick = Omit<
```
MythicManiac, line 21: This seems like a slightly odd type to have, perhaps necessary but it does stand out to me at a quick glance. Could the onClick be made optional to begin with on the base type?

packages/components/src/components/PackageActions.tsx
```diff
@@ -107,7 +95,10 @@ export const PackageActions: React.FC<PackageActionsProps> = (props) => {
           {dependencyString}
         </Button>
 
-        <PackageDependantsLink community={communityName} package={packageName}>
+        <PackageDependantsLink
+          community={communityIdentifier}
+          package={packageName}
```
MythicManiac, line 100: This is probably lacking namespace (unless it's implied in `packageName` alread, which I doubt). I'm expecting this might be addressed with the upcoming data provider PR (due to the backend also needing it), so if it's easier to address at the same time there, feel free to ignore for now in this PR.

MythicManiac (commented):
Overall looks good, just a few questions

apps/nextjs/utils/transforms/packageCard.ts
anttimaki: Sounds good to me.

packages/components/src/components/PackageActions.tsx
anttimaki: This was actually discussed in [a backend PR](https://github.com/thunderstore-io/Thunderstore/pull/507) but I'll comment here also for completeness sake: namespacing was overlooked altogether in the links and will be addressed in a separate PR.

anttimaki:
> I notice a lot of the changes here are related to local dummy data, would it be beneficial to implement a system specifically for generating dummy data and re-using that? Could help at least with (future) tests as well as storybook, and would allow organizing the code a bit better (moving dummy data generation to a separate module somewhere).

Perhaps having a separate system for dummy data generation would be a good idea, I'm not completely sure at this point yet (e.g. which is the recommended way for Storybook). Whatever I've written so far is mostly intended as a quick'n'dirty throwaway code.

MythicManiac:
> Perhaps having a separate system for dummy data generation would be a good idea, I'm not completely sure at this point yet (e.g. which is the recommended way for Storybook). Whatever I've written so far is mostly intended as a quick'n'dirty throwaway code.

We could do something super simple in this PR, e.g. by just moving the functions to a dedicated module for data generation, and then only later on think about the best way to generate the dummy data if necessary?

anttimaki:
Changes:

* In [this old commit](https://github.com/thunderstore-io/thunderstore-ui/pull/370/commits/3e09a2b0500c6eb6079f6ffe06de059b3053219f)
  * `PackageCardPropsWithoutOnClick` interface was removed since it wasn't needed anymore (the prop is now optional)
  * boolean fields "received" from "backend" have `is_` prefix in their names
* In [this old commit](https://github.com/thunderstore-io/thunderstore-ui/pull/370/commits/1f1a957beb37ba19c7b7e874f59655cb797001a3)
  * `PackageCard`'s `categoryOnClick` prop is now optional
  * Boolean props of `PackageCard` and `PinnedTag` components use `is` prefix in their names
* Added [a commit](https://github.com/thunderstore-io/thunderstore-ui/pull/370/commits/b1f9c2f4dca0670e7f48d8a43d01adc0bb39681d) to support displaying unlisted packages as dependencies
* Added [a commit](https://github.com/thunderstore-io/thunderstore-ui/pull/370/commits/9fc737c2782e00e9713aa4d7f946ca39087c8baa) to extract dummy data from page definition files to separate files

packages/components/src/components/LinkingProvider.tsx
MythicManiac: Let's figure this out in a separate PR

MythicManiac (approved):
Nice work 👍

PR #375, opened 2022-03-24
Data provider

anttimaki:
Draft for early discussion and feedback regarding data fetching:

* "Data provider" implemented in a separate package
* A class with typed fetching methods, to be used in `getServerSideProps` etc.
* A context provider wrapping said class, and a hook for using it in function components

Any thoughts on error handling (when/where) or e.g. data validation/type checking?


anttimaki:
A more thought-out, complete implementation. In addition to the earlier comment:

* Error handling is not implemented at this point
* The data provider uses Zod to validate the received data
* Data provider encapsulates the transforming of keys (e.g. from snake_case to camelCase)
* More thorough processing for query strings to make requests more cache-friendly

packages/dapper/src/typeguards.ts
```diff
@@ -0,0 +1,23 @@
+export const isNumberArray = (value: unknown): value is number[] => {
```
MythicManiac, line 1: Ironically, this will actually pass for `[NaN]`, since `typeof NaN` is also `"number"`. If intended, that's fine, just making sure.

packages/dapper/src/methods/communityPackageList.ts
```diff
@@ -0,0 +1,75 @@
+import { PackageCardProps, SelectOption } from "@thunderstore/components";
```
MythicManiac, line 1: Same here

packages/dapper/src/errors.ts
```diff
@@ -0,0 +1,15 @@
+export class DapperError extends Error {
```
MythicManiac, line 1: I remember there being some wonkiness when it comes to extending (and appropriately catching) custom errors in JS/TS. As error handling isn't implemented yet, I don't think it's a major issue, but just noting down for when we're adding error handling as we'll probably want to write tests for that.

packages/dapper/src/queryString.ts
```diff
@@ -0,0 +1,125 @@
...
+
+type QueryStringable = string | number | boolean | string[] | number[];
+
+interface QsItem<T extends QueryStringable> {
+  key: string;
+  value?: T;
+  impotent?: T;
+}
+
+interface DefinedQsItem<T extends QueryStringable> extends QsItem<T> {
```
MythicManiac, line 11: Minor note, but I feel like `RequiredQsItem` would be a more standard naming scheme.

packages/dapper/src/queryString.ts
```diff
@@ -0,0 +1,125 @@
...
+export const serializeQueryString = (allParams: QsArray): string => {
+  const keys = allParams.map((p) => encodeURIComponent(p.key));
+
+  if (keys.length != new Set(keys).size) {
+    throw new Error("Duplicate keys are not allowed. Use value array instead.");
+  }
+
+  return (
+    _filterEmptyAndImpotent(allParams)
+      // Force lower case to improve the changes of cache hits.
```
MythicManiac, line 48: Typo (`changes` -> `chances`), but also are we sure this is a valid way of handling query strings? More specifically, do we have guarantees our query parameters are in fact always case insensitive? Same for values, seems like a risky assumption to make.

packages/dapper/src/context.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import { Dapper } from "./dapper";
+
+const DapperContext = React.createContext<Dapper | null>(null);
+
+export const DapperProvider: React.FC<{ apiDomain: string }> = (props) => {
```
MythicManiac, line 7: Perhaps instead of taking in the `apiDomain` as a prop, we should take the `DapperInterface` instance to keep the context re-usable with other implementations too?

MythicManiac (commented):
Overall looks good, especially nice to have some more tests in this repo.

I've commented about a few issues, and I feel like most of them can & should be addressed in separate pull requests given the size of this one. In other words, none of them are blockers, but should probably be addressed.

On top of what I commented on, there's also:
- `useEffect` is not really nice to use when implementing components, providing a hook based data fetching API would be nicer to work with.
  - Example on how Apollo does it: https://www.apollographql.com/docs/react/data/queries/#executing-a-query
  - Example on how SWR does it: https://swr.vercel.app/examples/basic
- Use of suspenses
- Error handling (related to both of the above, should probably not be done before)
- Ensuring we support caching as intended (304 status codes, ETags, last-modified headers) when serving from NextJS

Some of the above are already planned I believe, but not everything. Let's make sure to everything down as separate issues before merging this one.

So in summary, I think we should delegate the comments into separate issues & merge this as is. Address comments in future separate PRs.

I'll mark the PR as approved once issues tracking the comments have been created on ClickUp.

packages/dapper/src/typeguards.ts
anttimaki: Added check for NaN and some simple test cases for type guards.

packages/dapper/src/transformers.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { packageCardSchema, packageCategorySchema } from "./schemas";
+
+export const transformCategories = (
+  categories: z.infer<typeof packageCategorySchema>[]
+): SelectOption[] => categories.map((c) => ({ value: c.slug, label: c.name }));
+
+export const transformPackageCards = (
+  packages: z.infer<typeof packageCardSchema>[]
+): PackageCardProps[] =>
```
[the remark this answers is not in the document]
anttimaki: I'm not sure I fully agree. I think controlled component such as `PackageCard` should be the one that defines which props/types it accepts/requires, and encapsulate this information. The component library should be usable without other dependencies, right?

I kinda agree that this should probably cut the other way around too, i.e. Dapper shouldn't depend on the interfaces defined in `@thunderstore/components`, either. But redeclaring the interface in Dapper feels like a poor choice too. I think I'd rather keep this as is (or maybe rename the interface).

packages/dapper/src/errors.ts
anttimaki: https://app.clickup.com/t/20413243/TS-307

packages/dapper/src/queryString.ts
anttimaki: I don't think it quite fits in this case? The interface is used when we've checked that the given value is not undefined. It's still not required as in a "required field", we've just filtered out those items where the value was not explicitly defined.

packages/dapper/src/queryString.ts
anttimaki: IDK, what kinda use cases do we have for supporting case insensitive values? I mean, sure we can not lower case the values if you feel like keeping up the option of possibly having case insensitive values outweights the possibility of cache misses. Doign the same for keys poses the additional problem of detecting duplicates/arrays.

Fixed the typos.

packages/dapper/src/context.tsx
anttimaki: Changed this as suggested.

anttimaki:
> * `useEffect` is not really nice to use when implementing components, providing a hook based data fetching API would be nicer to work with.

I actually spent some time trying to create a reusable custom hook, but it got quite complicated trying to combine

* props (`getServerSideProps`)
* default values read from GET parameters
* changes to filters (so many filters) by user actions
* infinite scrolling tracking
* strong typing

So I ended giving up on it. I'm not saying it can't be done, but I do believe it won't be worth the time investment.

That being said, I agree that it's a good idea to revisit the idea of a custom hook for other, simpler pages, like the front page where we just load some data and that's it.

anttimaki:
> * Use of suspenses

My memory is absolute garbage, but didn't we agree that mixing suspense with Next.js/SSR/data provider was not going to be a good idea?

anttimaki:
> * Error handling (related to both of the above, should probably not be done before)
> 
> * Ensuring we support caching as intended (304 status codes, ETags, last-modified headers) when serving from NextJS

https://app.clickup.com/t/20413243/TS-307 and https://app.clickup.com/t/20413243/TS-308 added.

packages/dapper/src/transformers.ts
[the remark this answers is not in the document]
MythicManiac: The problem I have with that approach is that now we're tightly coupling two packages together instead of them being standalone, effectively making it so the separation makes no sense.

If we consider it from a (library) consumer perspective, it'd make sense the data provider also determines the shape of the data provided. Here it's harder to draw the lines because we're both the consumer and the provider (and the backend API) all at once, but consider a 3rd party application relying on the data provider. Now they'd be also forced to rely on our component library even if they don't use react.

I'd suggest we either reverse the dependency relation, or we define a completely separate package for common data types. I'm not sure which approach is better at this point.

So in summary, even though it intuitively makes sense for our dev process to have the dependency this way around (since the API has been designed with the client in mind), it doesn't make sense from a separation of concerns perspective & potential future library consumers.

packages/dapper/src/queryString.ts
MythicManiac: Ah, that makes more sense. I got confused thinking from type hinting perspective, i.e. if I were to type a query param interface of which some properties are required and some optional. I think this way around the naming logic is sound 👍

packages/dapper/src/queryString.ts
MythicManiac: > what kinda use cases do we have for supporting case insensitive values?

I assume you mean **case-sensitive**, as the current proposed implementation would make them case-insensitive. For an example where case-insensitive values don't work, string based IDs that use lower & upper chars are fairly common (e.g. base64-encoded IDs which aren't too uncommon), which would break if we implicitly cast everything to lowercase.

In terms of caching, if our server is treating query params as case-sensitive, incorrect casings would lead to error pages and thus wouldn't be a problem for the cache.

So I believe we should not implicitly transform the values in any way since that might be significant to the functionality. Ordering is not significant (at least in any of the current APIs we have, and I'd imagine it's not going to be in future APIs either).

MythicManiac:
> My memory is absolute garbage, but didn't we agree that mixing suspense with Next.js/SSR/data provider was not going to be a good idea?

That might very well be so. As mentioned [here](https://github.com/thunderstore-io/thunderstore-ui/pull/386#discussion_r853591844), we might want to evaluate what kind of design constraints NextJS imposes on us currently, and if that makes sense. There might be a way to architecture things without NextJS becoming a blocker for features such as suspenses. Perhaps we can combine that evaluation when looking into suspense use in general, and make that a broader architecture planning topic? This also applies to the comment about hooks.

Beyond the comment about package dependencies, this looks good to me (though we do want to create a ticket about those hooks/suspenses/NextJS architecture implications)

packages/dapper/src/transformers.ts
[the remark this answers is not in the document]
anttimaki: Well, as I stated above, IMO reversing the dependency makes no sense to me and solves nothing, so I'll guess we should look into creating a separate package. I added [this card](https://app.clickup.com/t/20413243/TS-314) for it, please check that were on the same page about this task.

packages/dapper/src/queryString.ts
anttimaki: > I assume you mean case-sensitive

Yep yep yep.

Casing is no longer adjusted for neither keys nor the values. Also removed the related test case.

anttimaki:
> That might very well be so. As mentioned [here](https://github.com/thunderstore-io/thunderstore-ui/pull/386#discussion_r853591844), we might want to evaluate what kind of design constraints NextJS imposes on us currently, and if that makes sense. There might be a way to architecture things without NextJS becoming a blocker for features such as suspenses. Perhaps we can combine that evaluation when looking into suspense use in general, and make that a broader architecture planning topic? This also applies to the comment about hooks.

Added https://app.clickup.com/t/20413243/TS-315

packages/dapper/src/transformers.ts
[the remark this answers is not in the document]
MythicManiac: Looks good enough to me, I'll consider this solved for this PR and we can address the design later

MythicManiac (approved):
Nice work 👍

PR #378, opened 2022-03-30
@thunderstore/nextjs: update @preconstruct/next to 4.0.0

anttimaki:
The update fixes an issue that prevented us from using Next's new,
faster rust-based compiler. The Babel config that was used by the old
fallback compiler was removed.


MythicManiac (approved):
LGTM 👍

PR #386, opened 2022-04-14
Add infinite scroll to community's package list view

anttimaki:


anttimaki:
Related backend PR: https://github.com/thunderstore-io/Thunderstore/pull/552

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from "react";
+
+/**
+ * Like React's useEffect, except doesn't execute on initial render.
+ */
+export const useEffectAfterInitialRender = (
+  func: () => void,
```
MythicManiac, line 7: This doesn't properly wrap the `useEffect` type, as the return value of the effect is an optional callable for cleaning up whatever the effect did, which you'd want to return.

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from "react";
+
+/**
+ * Like React's useEffect, except doesn't execute on initial render.
+ */
+export const useEffectAfterInitialRender = (
```
MythicManiac, line 6: The need for this hook to exists feels wrong. I won't say it's wrong per se, but I've found that usually if something like this has to be done, there's a cleaner solution that can be reached some other way. For now I don't have any concrete suggestions however.

MythicManiac (commented):
There's a few odd patterns at play here which I feel like aren't probably the cleanest solutions, but I also don't know what to offer as a replacement.

I'd explore turning the package list data fetching into a single hook (as also mentioned in the other PR) if possible, which would make the `CommunityPackages` component much simpler, and likely give ideas on a cleaner implementation of what's happening there currently.

That said, refactoring it on that scale is outside of the scope of this PR, and should be delegated to a separate issue (which should be created in any case as mentioned in #375). I'll approve this PR once done.

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
anttimaki: I did some changes based on this comment. I'm not 100% they are now correct, i.e. I wasn't sure whether the returned cleanup should also do something for the ref-value used by the hook internally. Things seemed to work as expected without additional tampering, so I just return whatever the effect-function returns.

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
anttimaki: The reason the hook is used here is to avoid unnecessary refetch when the page is loaded directly. In such case, when using regular `useEffect`, The first page is first fetched by `getServerSideProps`, and then refetched when the page mounts. IIRC I couldn't find a cleaner solution for this.

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useEffectAfterInitialRender: typeof useEffect = (
+  effect,
+  deps?
+) => {
+  const didMount = useRef(false);
+  let cleanup: ReturnType<EffectCallback> = undefined;
+
+  useEffect(() => {
+    if (didMount.current) {
+      cleanup = effect();
```
MythicManiac, line 15: I think the cleanup needs to be returned from the effect callback here, as otherwise React won't be calling it appropriately. See https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
MythicManiac: Understood, let's go with this for now then. That said, NextJS imposing restrictions on non-NextJS components is a design consideration we might need to evaluate a bit more down the line, it seems like it could end up being counter productive even if the alternative is duplicating a bit of code.

MythicManiac (commented):
One issue left

packages/hooks/src/hooks/useEffectAfterInitialRender.ts
anttimaki: Changed.

PR #400, opened 2022-04-25
Fix Next.js app build process

anttimaki:
Build process failed due to errors in Dapper's package.json file:

* Add @babel/runtime required by preconstruct
* Move zod from devDependencies to dependencies

Refs TS-318


MythicManiac (approved):
Nice 👍

PR #417, opened 2022-05-23
WIP implementation for authentication, pt 5

anttimaki:
Implemented in earlier commits (in this repo and backend repo):

1. User clicks an authentication link
2. User's browser stores a "state token" to local storage (for CSRF
   prevention) and redirects to OAuth provider's endpoint for login
3. After login, browser is redirected to client-side page, with query
   parameters containing the state token (which is validated) and a
   "code token"
4. Client-side page sends a request to Next.js API endpoint, which
   creates another request containing the code token and a "shared
   secret token" (for brute force prevention). For security reasons
   the latter token is not available client-side, which is why this
   sidestep is required
5. Next.js API endpoint sends the request to Django API endpoint
6. Django API endpoint validates the shared secret token
7. The endpoint sends a request containing the code token to OAuth
   provider, and receives an "access token"
8. The endpoint queries user's information from another API using the
   access token for authentication
9. If no account is found with the received information, one is created
10. Endpoint creates a session for the identified user
11. Endpoint returns session id to Next.js API endpoint

This commit implements the following steps of the authentication flow:

12. Next.js API endpoint relays the session id to the client-side page
13. Client-side page stores the session id

In future commits:

14. Client uses session id in the following requests

Refs TS-230, TS-356

### DIFF
--- a/apps/nextjs/components/SessionContext.tsx
+++ b/apps/nextjs/components/SessionContext.tsx
@@ -1,0 +1,80 @@
+import { createContext, FC, useContext, useState } from "react";
+
+import { StorageManager } from "utils/storage";
+
+interface ContextInterface {
+  /** Remove sessionId from provider's state and localStorage. */
+  clearSessionId: () => void;
+  /** Fetch sessionId from provider's state or localStorage. */
+  sessionId?: string;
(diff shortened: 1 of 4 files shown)


MythicManiac (approved):
LGTM 👍

PR #419, opened 2022-05-23
WIP implementation for authentication, pt 6

anttimaki:
Implemented in earlier commits (in this repo and backend repo):

1. User clicks an authentication link
2. User's browser stores a "state token" to local storage (for CSRF
   prevention) and redirects to OAuth provider's endpoint for login
3. After login, browser is redirected to client-side page, with query
   parameters containing the state token (which is validated) and a
   "code token"
4. Client-side page sends a request to Next.js API endpoint, which
   creates another request containing the code token and a "shared
   secret token" (for brute force prevention). For security reasons
   the latter token is not available client-side, which is why this
   sidestep is required
5. Next.js API endpoint sends the request to Django API endpoint
6. Django API endpoint validates the shared secret token
7. The endpoint sends a request containing the code token to OAuth
   provider, and receives an "access token"
8. The endpoint queries user's information from another API using the
   access token for authentication
9. If no account is found with the received information, one is created
10. Endpoint creates a session for the identified user
11. Endpoint returns session id to Next.js API endpoint
12. Next.js API endpoint relays the session id to the client-side page
13. Client-side page stores the session id

This commit implements the following steps of the authentication flow:

14. Client uses session id in the following requests

Refs TS-230, TS-357

### DIFF
--- a/apps/nextjs/hooks/useFreshProps.ts
+++ b/apps/nextjs/hooks/useFreshProps.ts
@@ -1,0 +1,44 @@
+import { Dispatch, SetStateAction, useEffect, useState } from "react";
+
+import { useSession } from "components/SessionContext";
+
+/**
+ * Refresh page contents from data API if user is authenticated.
+ *
+ * The session id used to authenticate the user for the data API is
+ * stored in the localStorage and thus isn't available when Next.js
+ * renders the initial page on the server-side. This hook can be used to
+ * refetch the data on the client-side in case the user is
+ * authenticated.
+ *
+ * Note: to avoid unnecessary requests, this should be used only on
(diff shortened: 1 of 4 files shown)


anttimaki:
As this implementation doesn't refetch the data unless the useFreshProps hook is used on the page, e.g. the frontpage will always be loaded as an unauthenticated user, even if user is authenticated. I was wondering if this would cause significant issues for gathering statistics later on?

MythicManiac (approved):
LGTM 👍 A fairly clean solution with for the content re-fetching from what I can tell. We might want to make sure there's no excessive content re-draws happening (caused by auth or otherwise) at some point, but I'd say that's a separate step.

PR #420, opened 2022-05-23
@thunderstore/nextjs: adapt TopBar content based on auth status

anttimaki:
* Toggle login/logout functionality
* Toggle features restricted to authenticated users

Refs TS-230, TS-461


MythicManiac (approved):
LGTM 👍

PR #430, opened 2022-06-03
@thunderstore/nextjs: rename community/package URL parameters

anttimaki:
* Use the same terms that are used in the code
* Avoid reserved word `package`


MythicManiac (approved):
LGTM, nice work 👍

PR #431, opened 2022-06-03
@thunderstore/nextjs: read url parameters in a consistent way

anttimaki:


apps/nextjs/pages/c/[communityIdentifier]/p/[namespace]/[packageName]/dependants.tsx
```diff
@@ -53,19 +54,23 @@ export default function PackageDependantsPage(props: PageProps): JSX.Element {
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
   // TODO: validate the community & package exist in database.
+  const communityIdentifier = getString(context.params?.communityIdentifier);
+  const namespace = getString(context.params?.namespace);
+  const packageName = getString(context.params?.packageName);
+
   if (
```
Comment (MythicManiac) [inline apps/nextjs/pages/c/[communityIdentifier]/p/[namespace]/[packageName]/dependants.tsx:61]: Not a blocker, but could consider using [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) in scenarios where multiple checks need to be done
```ts
if ([communityIdentifier, namespace, packageName].some(x => x === undefined)) {
  // ..
}
```

That said, having it explicitly as it is now is easier to read IMHO, so the tradeoff might not even make sense. An array check is more DRY of course.

MythicManiac (approved):
Looks good to me, nice work 👍 

Feel free to mark the discussion as resolved & merge, it wasn't intended as a blocker.

apps/nextjs/pages/c/[communityIdentifier]/p/[namespace]/[packageName]/dependants.tsx
Comment (anttimaki) [reply apps/nextjs/pages/c/[communityIdentifier]/p/[namespace]/[packageName]/dependants.tsx:61]: IIRC I tried this approach first as well (I guess due to the Python background), but TypeScript didn't infer correctly that the variables can't be `undefined` after that check.

PR #432, opened 2022-06-03
@thunderstore/nextjs: fix client-side page content refresh

anttimaki:
Since the sessionId is stored in localStorage and unavailable during
SSR, we need to refetch the content on client-side for some pages. To
achieve this, a hook was provided, but if the SSR resulted in 403
response, an error page was shown instead of the actual content and the
hook never executed.

The approach in this commit suppresses 404 errors on SSR and returns a
"soft 404" flag, which tells the client-side to run the hook if user is
authenticated. Hard 404 errors may be returned by server-side in some
cases, e.g. if the requested URL doesn't contain all the required URL
parameters.

Backend now responds with 404 instead of 403 when unauthenticated user
attempts to access a package that requires authentication, in order to
avoid revealing the existence of the said package.

The hook and provided related helpers aim to be reusable, but this is
kinda hard to verify since we currently only have one page where
refetching is required. It might not work on complex pages, e.g.
package listing, where multiple filters are read from query string
and stored separately on different state variables.

Refs TS-230, TS-465


MythicManiac (approved):
Looks good to me 👍 A bit complex but it is a complex functionality and I'm not sure if there's a good way to simplify it

PR #435, opened 2022-06-06
Login returns username

anttimaki:


MythicManiac (approved):
This change looks good to me 👍 

In the future it might be best to have a separate API endpoitn for fetching & refreshing session information, as this is now effectively a cache that only gets refreshed if you log out & log in again.

MythicManiac:
@anttimaki that'd be true, except the Thunderstore username doesn't get updated even if it changes on the OAuth connection (or at least doesn't in the current version, and I've missed it in the PR if it does in the new one)

anttimaki:
@MythicManiac It doesn't (unless we want to start reading it from stored JSON blob that we currently don't use for anything).

PR #440, opened 2022-06-10
Add base for the sertra project

MythicManiac:
Includes a NextJS typescript template with the basic website layout implemented.


apps/sertra/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+First, run the development server:
+
+```bash
+npm run dev
+# or
+yarn dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
```
anttimaki, line 13: @thunderstore/nextjs also runs on port 3000, should this use 3001 oslt by default?

Edit: Oh, it seems the dev server will automatically move to 3001 if 3000 is in use. Anyways, could still avoid some confusion?

apps/sertra/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
anttimaki, line 1: Do we want to use app specific lint rules, or should sertra use the ones defined in the root dir? Or extend it?

apps/sertra/next.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
```
anttimaki, line 2: Should sertra be hooked up with preconstruct, or do we want to keep this separate?

apps/sertra/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "sertra",
```
anttimaki, line 2: To keep naming consistent with other apps, this should probably be `@thunderstore/sertra`?

apps/sertra/README.md
```diff
@@ -0,0 +1,34 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+
+## Getting Started
+
+First, run the development server:
+
+```bash
+npm run dev
+# or
+yarn dev
```
anttimaki, line 10: Or from the root dir as `yarn workspace sertra dev` 🤷🏻‍♂️

apps/sertra/components/footer.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC, PropsWithChildren } from "react";
+
+export const Footer: FC<PropsWithChildren> = () => {
+  return (
+    <footer style={{ backgroundColor: "#17213B" }}>
+      <div style={{ width: "1140px", marginLeft: "auto", marginRight: "auto" }}>
```
anttimaki, line 6: Is it a deliberate choice to use regular CSS instead of Chakra, and if so, is it intended to be for the MVP or a permanent choice?

apps/sertra/components/navbar.tsx
```diff
@@ -0,0 +1,41 @@
+import { FC, PropsWithChildren } from "react";
+
+export const Navbar: FC<PropsWithChildren> = () => {
+  return (
+    <div style={{ width: "1140px", marginLeft: "auto", marginRight: "auto" }}>
```
anttimaki, line 5: I was going to suggest that this could use `TopBar` from `@thunderstore/components`, but it seems it's not that easy, **maybe** due to React version mismatch. Maybe not an issue for MVP, but in the long run we probably want to be able to use shared components here?

anttimaki (approved):
Also running the server locally I get a warning: `The static directory has been deprecated in favor of the public directory.`

Approving now since I consider none of my comments a blocker for a MVP version.

apps/sertra/README.md
MythicManiac: Perhaps, but the other side of the coin is that 3000 is the default port and if you're only running a single project, you'd probably expect it to go to port 3000. So I'd keep it as is given that the project is developed standalone for the time being.

apps/sertra/components/footer.tsx
MythicManiac: Was the fastest way to get it done, but not meant to be a permanent choice. At the very least inline styles should be replaced with some better way to style components

apps/sertra/components/navbar.tsx
MythicManiac: Yeah possibly, at this stage I'm intentionally keeping them separate because the `@thunderstore/components` library is under more scrutiny than this app, so we could end up inadvertently slowing down the development velocity here by relying on the shared components.

We'll just need to make a conscious effort in consolidating common elements to the component library once it has stabilized a bit.

apps/sertra/next.config.js
anttimaki: I guess it doesn't matter for now if you intent this to be developed in isolation from the rest of the "projects" in this monorepo, as suggested in one of your other comments. If you start doing something like `import { Foo } from '../../../../packages/components/src/components/Foo`, and want to use `import { Foo } from '@thunderstore/components'` instead, it should be set up. Although fixing the imports later on isn't that big of an issue.

PR #451, opened 2022-06-21
Sertra: Partially implement server form styles

MythicManiac:


MythicManiac:
image

anttimaki (approved):
Pretty much the same comments as in #448 (in order of relevancy):

* Try to use relative units for sizes
* Consider using semantic HTML elements
* Consider using CSS variables
* Consider using `font` shorthand where applicable

PR #459, opened 2022-06-30
@thunderstore/components: yeet Chakra's spacer

anttimaki:
Prefer CSS over adding superfluous divs for presentation's sake.

Refs TS-548

### DIFF
--- a/packages/components/src/components/PackageHeader.tsx
+++ b/packages/components/src/components/PackageHeader.tsx
@@ -11,7 +11,6 @@ import {
   ModalFooter,
   ModalBody,
   ModalCloseButton,
-  Spacer,
   Tag,
   Text,
   useDisclosure,
@@ -104,8 +103,6 @@ export const PackageHeader: React.FC<PackageHeaderProps> = (props) => {
           {description}
         </Text>
 
-        {aboveBreakpoint ? <Spacer /> : null}
-
         <TagListing {...{ categories, communityIdentifier, renderFullWidth }} />
       </Flex>
     </Flex>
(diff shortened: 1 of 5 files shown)


anttimaki:
Removed all `<Spacer>`s from `@thunderstore/components`. [This one](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/nextjs/pages/c/%5BcommunityIdentifier%5D/packages.tsx#L184) in `@thunderstore/nextjs` was left untouched since I couldn't figure how to replace it in a way that doesn't affect the behaviour on narrow layouts or introduce other unnecessary divs.

packages/components/src/components/PackageHeader.tsx
```diff
@@ -149,7 +142,7 @@ const TagListing: React.FC<TagsProps> = (props) => {
   // in a modal window.
   return (
     <Box alignSelf="reverse" mt={renderFullWidth ? "0" : "20px"}>
-      {visible.map((t) => tag(t))}
+      {visible.map(tag)}
```
gkurck, line 145: I'm guessing this and similar changes elsewhere are for simplifying code while you're at it, which is cool.

gkurck (approved):
Looks good to me!

About the the one in [here](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/nextjs/pages/c/%5BcommunityIdentifier%5D/packages.tsx#L184), we could probably wrap the tab-nav items to div.tab-nav and stack that and the breadcrumbs vertically as the space is so tight especially whit long names, but that would be a separate task later.

PR #481, opened 2022-08-11
ServerListing Detail page

Oksamies:
Full Detail page including data fetching from API
API is missing some data for mods,
but those have been marked TODO

refs TS-520

### DIFF
--- a/apps/sertra/components/ServerDetail.module.css
+++ b/apps/sertra/components/ServerDetail.module.css
@@ -1,0 +1,328 @@
+.container {
+  width: 100%;
+  display: flex;
+  column-gap: 2em;
+  flex-direction: column;
+}
+
+.headerRow {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+}
+
+.contentRow {
+  display: flex;
+  column-gap: 2em;
+}
+
+.columnLeft {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+}
+
+.columnRight {
+  display: flex;
+  flex-direction: column;
+}
+
+.descriptionContent,
+.row {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  padding: 0.75em 1em 0.75em 0.75em;
+  margin-bottom: 0.5em;
+  gap: 1em;
+  font-size: 16px;
+  line-height: 1.4;
+}
+
+.descriptionContent,
+.row {
+  background-color: var(--color-blue);
+  border: 0.125em solid var(--color-lightblue);
+  font-weight: 500;
+  cursor: pointer;
+}
+
+.mods {
+  width: 100px;
+  text-align: center;
+}
+
+.listingTitle {
+  flex-basis: 58%;
+}
+
+.listingTitle {
+  margin-top: 2em;
+  font-family: "Raleway";
+  font-style: normal;
+  font-weight: 900;
+  font-size: 30px;
+  line-height: 1.16;
+
+  display: flex;
+  align-items: center;
+  color: #ffffff;
+}
+
+.sectionTitle {
+  font-family: "Raleway";
+  font-style: normal;
+  font-weight: 700;
+  font-size: 18px;
+  line-height: 1.22;
+
+  display: flex;
+  align-items: center;
+
+  color: var(--color-babyblue);
+  margin-top: 2em;
+  margin-bottom: 1em;
+}
+
+.descriptionContent {
+  box-sizing: border-box;
+
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 1.5em 2em;
+  gap: 1em;
+
+  background-color: var(--color-blue);
+
+  border: 0.125em solid var(--color-lightblue);
+}
+
+.descriptionContentText {
+  font-style: normal;
+  font-weight: 500;
+  font-size: 16px;
+  line-height: 1.4;
+
+  display: flex;
+  align-items: center;
+
+  color: #ffffff;
+}
+
+.modImg {
+  width: 64px;
+  height: 64px;
+
+  background: greenyellow;
+  border-radius: 0px;
+}
+
+.modName {
+  font-style: normal;
+  font-weight: 700;
+  font-size: 16px;
+  line-height: 1.1875;
+  display: flex;
+  align-items: center;
+
+  color: #ffffff;
+}
+
+.modDescription {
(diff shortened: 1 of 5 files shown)


Oksamies:
image

Oksamies:
image
image
image

Oksamies:
image

Oksamies:
Oh also the `Join Server` button is missing functionality and the logo, since I'm not sure how the App opening stuff goes.

MythicManiac:
@Oksamies install pre-commit hooks for the repo

apps/sertra/pages/server/[id].tsx
```diff
@@ -0,0 +1,206 @@
...
+  const listings = await res.json();
+
+  const paths = listings?.results.map((listing: ServerListingData) => ({
+    params: { id: listing.id },
+  }));
+
+  return { paths, fallback: "blocking" };
+}
+
+export async function getStaticProps(context: any) {
```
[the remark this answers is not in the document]
Comment (MythicManiac) [reply apps/sertra/pages/server/[id].tsx:17]: There's a `GetStaticProps` type in NextJS that you can use here, check https://wallis.dev/blog/nextjs-getstaticprops-and-getstaticpaths-with-typescript (at the bottom) for an example

MythicManiac (approved):
We should split the large page component to smaller standalone components as well as move the components to their own files. This should also split up the CSS file to something more manageable instead of the current 300-liner.

These aren't blockers however and we should revisit them later. This PR is mergeable as far as I'm concerned.

apps/sertra/components/ServerDetail.module.css
```diff
@@ -0,0 +1,354 @@
...
+  padding: 0.75em 1em 12px 0.75em;
+  margin-bottom: 0.5em;
+  gap: 1em;
+  font-size: 16px;
+  line-height: 140%;
+}
+
+.descriptionContent,
+.row {
+  background: #242e48;
```
anttimaki, line 46: My very recent PR introcuded CSS variables for our [theme colors](https://github.com/thunderstore-io/thunderstore-ui/pull/488/commits/d198b1c424d9a4f177193812c36bd237e650cc3d) and [basic fonts](https://github.com/thunderstore-io/thunderstore-ui/pull/488/commits/d078dab16f7b5c67e94861caca258ce944991a15), as well as changed from `px` to `rem` units. This PR looks like the CSS could be improved by adopting those things.

apps/sertra/components/ServerDetail.module.css
```diff
@@ -0,0 +1,354 @@
...
+}
+
+
+.descriptionContent,
+.row {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  padding: 0.75em 1em 12px 0.75em;
```
anttimaki, line 37: `em` units are fine AFAIK, but mixing `px`s and `em`s seems like a bad idea.

apps/sertra/components/ServerDetail.module.css
```diff
@@ -0,0 +1,354 @@
...
+  margin-bottom: 0.5em;
+  gap: 1em;
+  font-size: 16px;
+  line-height: 140%;
```
anttimaki, line 41: The aforementioned recent PR also changed `line-height`s from percentages to unitless values, which is [recommended](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#prefer_unitless_numbers_for_line-height_values).

apps/sertra/components/ServerDetail.module.css
```diff
@@ -0,0 +1,354 @@
...
+
+.joinServerButtonText {
+  width: 5.1875em;
+  height: 1.1875em;
+
+  font-weight: 700;
+  color: #202941;
+}
+
+@media (max-width: 1199px) {
```
anttimaki, line 306: IIRC the layout has a fixed width of 1140px so this seems a bit odd (might be what we want though). Anyways I guess we should unify the breakpoints/widths so all the pages seem consistent.

apps/sertra/pages/server/[id].tsx
```diff
@@ -0,0 +1,206 @@
...
+      detail_listing: data,
+    },
+    revalidate: 10,
+  };
+}
+
+interface ServerModeProps {
+  isPvP: boolean;
+}
+const ServerMode: React.FC<PropsWithChildren<ServerModeProps>> = ({
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:33]: This component (and `ServerPasswordProps` below) seems to be copypasta from server listings view. Usually the way would be to make them separate, reusable components. But as it happens, my current PR introduces [changes to those components](https://github.com/thunderstore-io/thunderstore-ui/pull/492/files), so maybe the separation should be done after that PR is (hopefully) merged.

anttimaki (commented):
In addition to the regular comments:

* It seems `display: flex` comes up in the CSS quite often, but it's hard to check each usage against markup to see if they actually make sense. Anyway, I got a suspicion that this goes a little overboard with `flex` usage, BUT I'm not sure if there's such a thing as too much `flex`, or if we should `flex` all the things? I'm thinking this e.g. from technical, maintainability, and code style perspectives. Maybe @grav can chip in?
* Everything's a `div`, again.

Not sure if any changes are required for the MVP, but I feel that doing the mentioned things properly the first time around wouldn't really take extra time, but fixing them afterwards does. So at least consider them in the future work.

apps/sertra/components/ServerDetail.module.css
Oksamies: I kind of just added this "mobile support" to try out how it would be done. I just replaced the fixed width in dev to 100% for the time I worked on it. I'm not sure if we want to add site scalability yet, but I'll add a task in the inbox.

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:33]: Rebased on master and added a commit that separates the listing attributes to it's own file. And then removed the copypasta.

apps/sertra/components/ServerDetail.module.css
anttimaki: There's already a card for this (TS-522), which I intend to work on next. (Or I'm already working on some of it's subtasks, to be exact.) So I guess this is good "as is".

apps/sertra/pages/server/[id].tsx
Comment (anttimaki) [reply apps/sertra/pages/server/[id].tsx]: Looks good, except I'd capitalize the name of the new component file. The code base is currently inconsistent in naming files (among other things), but I'd assume we're going to end up using PascalCase.

PR #485, opened 2022-08-15
@thunderstore/sertra: add Dockerfile

anttimaki:
Based on the Dockerfile for @thunderstore/nextjs.

Refs TS-504


Repez (approved):
Looks good, runs great 👍

apps/sertra/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# For running @thunderstore/nextjs in Docker container.
+FROM node:17.0-alpine
+# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
+RUN apk add --no-cache libc6-compat
+WORKDIR /app
+ARG NEXT_PUBLIC_API_URL
```
MythicManiac, line 6: Could've used a default value here: https://docs.docker.com/engine/reference/builder/#default-values

PR #488, opened 2022-08-16
@thunderstore/sertra: improve CSS quality

anttimaki:
This PR tries to address the PR feedback given on #448. Only CSS related stuff is addressed to keep the PR size manageable, and possibly to split the reviewing between different people. I intend to work on the other stuff while I finish the partial implementation, but that task has a separate ticket.


apps/sertra/styles/globals.css
```diff
@@ -1,10 +1,25 @@
...
+  --color-darkblue: #17213b;
+  --color-highlightgreen: #02ff95;
+  --color-highlightred: #ff3a52;
+  --color-lightblue: #353e58;
+  --color-lightblue115: #49567a; /* Lightened by 15% */
+  --color-orange: #ff956b;
+  --color-orange99: #fc9347; /* Darkened by 1% */
+
+  --font-family-body: Exo2, sans-serif;
+  --font-family-header: Raleway, serif;
```
gkurck, line 14: Raleway is a sans-serif typeface, so I suggest to edit the fallback accordingly.

gkurck (changes requested):
Looks good to me, only one minor issue with a typeface fallback.

apps/sertra/styles/globals.css
anttimaki: Fixed

gkurck (approved):

PR #492, opened 2022-08-17
@thunderstore/sertra: fix listings layout to match designs

anttimaki:


anttimaki:
* master
  * **PR #492** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/492" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/492?utm_source=stack-comment).

apps/sertra/components/ServerList.tsx
```diff
@@ -64,24 +53,20 @@ export const ServerList: React.FC<PropsWithChildren<ServerListProps>> = ({
   }, [community]);
 
   return (
-    <div>
-      <div className={styles.header}>
-        <div className={styles.titleContainer}>
-          <div className={styles.pageTitle}>Servers</div>
-          <div className={styles.gameTitle}>{gameDisplayName}</div>
-        </div>
+    <>
```
gkurck, line 56: Is there a reason for omitting the header container? IMO it would be beneficial to wrap every segment/block in a container.

apps/sertra/components/ServerList.tsx
```diff
@@ -64,24 +53,20 @@ export const ServerList: React.FC<PropsWithChildren<ServerListProps>> = ({
...
+      <div className={styles.titleContainer}>
+        <div className={styles.pageTitle}>Servers</div>
+        <div className={styles.gameTitle}>{gameDisplayName}</div>
       </div>
       <div className={styles.tableHeader}>
```
gkurck, line 61: IMO it would be beneficial to wrap all of the listing items in a container.

gkurck (approved):
As far as the scope goes everything looks good to me.

I left a few comments which I feel are a bit outside of the scope of this PR, but would be beneficial to address or have a conversation about.

apps/sertra/components/ServerList.module.css
```diff
@@ -1,9 +1,15 @@
+.titleContainer,
+.tableHeader,
+.row {
+  width: 100%;
+  min-width: 40rem;
```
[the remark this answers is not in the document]
anttimaki: It's to ensure that the name field retains a decent width. Since the mode/mod count/password columns at the end have fixed widths, on the narrow displays the name was quicky hidden, but I assume it's the most important bit of information and should remain visible at all times. I guess making the "extra columns" getting smaller/hidden would be the better solution now that I think of it? Maybe I should create another ticket for it?

apps/sertra/components/layout.tsx
```diff
@@ -6,7 +6,9 @@ export const Layout: FC<PropsWithChildren> = (props) => {
   return (
     <>
       <Navbar />
-      <div style={{ width: "1140px", marginLeft: "auto", marginRight: "auto" }}>
+      <div
+        style={{ maxWidth: "1140px", marginLeft: "auto", marginRight: "auto" }}
```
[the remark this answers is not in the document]
anttimaki: Agreed, but I interpreted that we already have a separate ticket for this, so I intentionally touched it as little as I needed to make sure the changed done to listing view works as intended. I'm not sure if I should even change this in this PR at all, now that I think of it, as it might have adverse effects on other views.

apps/sertra/components/ServerList.tsx
anttimaki: White kind of benefits do you have in mind? Robustness? Maintainability? I considered the `div` to be unnecessary, in the same line of thought that we yeeted Chakra's spacer divs.

apps/sertra/components/ServerList.tsx
anttimaki: Same comment as above. Not against it, but I'd like to know why you think that?

apps/sertra/components/ServerList.tsx
gkurck: Maintainability by containment and separation of clear blocks of content / components. Basically the same reason as we have a container for the two title elements right below - to contain a one context of content.

While the header at this time has only the title, the fact can (and probably will) change with the addition of another element, such as pagination or a button, depending on the page. They both belong in the same context / block of content, and should be contained accordingly, so a container would be needed eventually anyway.

Having a container makes it so that any layout specific heavy lifting, such as content width, can be done at the container level, leaving the children agnostic about what's going on around them. An example of this is the comment above regarding having the whole site wrapped in a container. It's all the same principle. In the context of this header, we actually want the margin-bottom to apply to the whole header, not just the title.

apps/sertra/components/ServerList.tsx
gkurck: Same as above. In addition, as an example here we could leverage flexbox + gap, leaving the margin-bottom out of the children.

apps/sertra/components/ServerList.module.css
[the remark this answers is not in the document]
gkurck: Gotcha, and I agree on narrow screens we should change it up. Right now for this phase one the designs are static in that regard, and we definitely need to address that, probably once we get this phase one out and to be tested in the first place.

I think the best way to go about it would be to lay the fields vertically for smaller screens. That would require adding the titles in the columns as well, hidden on large screens, of course, and consequently hiding the header row completely on smaller screens.

I'll have to reflect this on the designs, guess we'll cross that bridge when we get there.

anttimaki:
I decided merge this as is (without reverting the removal of wrapper divs), because
* At least in this case, I don't see tangible benefits in doing so
* It would require "too much" work propagating those changes to the other PRs chained to this one.

I'll try to keep the point in mind in the future.

PR #509, opened 2022-08-31
add  keyboard controls and fix form

VilppeRiskidev:
added some keyboard controls and fixed some other issues

### DIFF
--- a/apps/sertra/components/mod-selector/list.tsx
+++ b/apps/sertra/components/mod-selector/list.tsx
@@ -17,8 +17,6 @@ export const ModRowInfo: React.FC<ModRowInfoProps> = ({
       <div className={styles.colSmall}>
         {/* eslint-disable-next-line @next/next/no-img-element */}
         <img
-          height={"32px"}
-          width={"32px"}
           className={styles.rowIcon}
           alt="mod icon"
           src={iconUrl}
--- a/apps/sertra/components/mod-selector/list.tsx
+++ b/apps/sertra/components/mod-selector/list.tsx
@@ -1,7 +1,6 @@
 import styles from "./list.module.css";
 import { IconButton } from "./iconButton";
 import { ModPackage } from "./data";
-import Image from "next/image";
 
 export interface ModRowInfoProps {
   iconUrl: string;
@@ -16,7 +15,14 @@ export const ModRowInfo: React.FC<ModRowInfoProps> = ({
   return (
     <>
       <div className={styles.colSmall}>
-        <Image className={styles.rowIcon} alt="mod icon" src={iconUrl} />
+        {/* eslint-disable-next-line @next/next/no-img-element */}
+        <img
+          height={"32px"}
+          width={"32px"}
+          className={styles.rowIcon}
+          alt="mod icon"
+          src={iconUrl}
+        />
       </div>
       <div className={styles.colLarge}>
         <span className={styles.packageName}>{packageName}</span>
--- a/apps/sertra/components/mod-selector/list.tsx
+++ b/apps/sertra/components/mod-selector/list.tsx
@@ -16,11 +16,7 @@ export const ModRowInfo: React.FC<ModRowInfoProps> = ({
     <>
       <div className={styles.colSmall}>
         {/* eslint-disable-next-line @next/next/no-img-element */}
-        <img
-          className={styles.rowIcon}
-          alt="mod icon"
-          src={iconUrl}
-        />
+        <img className={styles.rowIcon} alt="mod icon" src={iconUrl} />
       </div>
       <div className={styles.colLarge}>
         <span className={styles.packageName}>{packageName}</span>
--- a/apps/sertra/components/mod-selector/search.tsx
+++ b/apps/sertra/components/mod-selector/search.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useState } from "react";
+import { ReactNode, useRef, useState } from "react";
 import styles from "./search.module.css";
 
 interface SearchBoxProps<T> {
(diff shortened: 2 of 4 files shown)


apps/sertra/components/mod-selector/list.tsx
```diff
@@ -16,7 +15,14 @@ export const ModRowInfo: React.FC<ModRowInfoProps> = ({
   return (
     <>
       <div className={styles.colSmall}>
-        <Image className={styles.rowIcon} alt="mod icon" src={iconUrl} />
+        {/* eslint-disable-next-line @next/next/no-img-element */}
+        <img
+          height={"32px"}
```
anttimaki, line 20: I prefer `height="32px"` for plain string values. But I guess this is something that the linter should catch on, so feel free to leave it as is. Either way you could ensure that we have a ticket for linting in this project, and that this issue is on the list of things we need a house rule for.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -18,6 +18,70 @@ export const SearchBox = <T,>({
 }: SearchBoxProps<T>) => {
   const [isFocused, setFocused] = useState<boolean>(false);
 
+  function focusFirstOption(target: EventTarget & HTMLInputElement) {
+    if (
```
anttimaki, line 22: Optional chaining operator can be used to call function too, i.e.
`target.parentElement?.parentElement?.nextElementSibling?.children[0]?.children[0]?.focus();`
would remove the need for the `if` block.

That being said, this approach seems really brittle. Small change in the markup will break the function. [useRef](https://reactjs.org/docs/hooks-reference.html#useref) should probably be used here to tag the first option and the search field.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -18,6 +18,70 @@ export const SearchBox = <T,>({
...
+      if (e.code === "ArrowDown") {
+        e.preventDefault();
+        if (e.currentTarget) {
+          focusFirstOption(e.currentTarget);
+        }
+      }
+    }
+  };
+
+  const listKeyDownHandler = (
```
anttimaki, line 52: Depending on the amount of keyboard stuff we're going to implement, we might want to consider a library for this to make the code more readable/maintainable. [Mousetrap](https://github.com/ccampbell/mousetrap) has worked for me in the past, but I'm not sure if it's still a relevant option today.

Anyway, that's probably outside the scope of this task. Consider adding a task to ClickUp anyway.

apps/sertra/pages/index.tsx
```diff
@@ -3,7 +3,7 @@ import Head from "next/head";
 import { SWRConfig } from "swr";
 import { getServerListings } from "../api/hooks";
 import { ApiURLs } from "../api/urls";
-import styles from "./ServerList.module.css";
+import styles from "../components/ServerList.module.css";
```
anttimaki, line 6: This seems unrelated to the PR, is it here on purpose?

anttimaki (changes requested):
Please check if the useRef would improve the solution. Other comments are optional to fix now.

apps/sertra/components/mod-selector/list.tsx
VilppeRiskidev: Those values are actually already set in the styles.rowIcon, so I'll remove them. Those were added because linter wanted to use Image component which would've required width and height props.

apps/sertra/pages/index.tsx
VilppeRiskidev: 'twas broken, so I fixed it, This could've been a separate pr, but it's so small it seemed reasonable to have it with these changes.

apps/sertra/components/mod-selector/search.tsx
VilppeRiskidev: Added a task for this.

apps/sertra/components/mod-selector/list.tsx
anttimaki: Well that solves the issue, but I'd appreciate if you'd create the linting ticket anyway. Feel free to resolve this conversation once you've read this comment.

apps/sertra/components/mod-selector/search.tsx
VilppeRiskidev: Now using refs instead.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -18,21 +18,15 @@ export const SearchBox = <T,>({
...
   const [isFocused, setFocused] = useState<boolean>(false);
 
-  function focusFirstOption(target: EventTarget & HTMLInputElement) {
-    if (
-      target.parentElement?.parentElement?.nextElementSibling?.children[0]
-        ?.children[0]
-    ) {
-      target.parentElement.parentElement.nextElementSibling.children[0].children[0].focus();
+  function focusFirstOption() {
+    if (optionsRef.current?.children[0]) {
```
anttimaki, line 22: While the fix for useRef is implemented, you've seemed to ignore the comment about optional chaining operator. I guess we don't need to use it here, but please comment if the "wontfix" was intentional.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -105,7 +103,7 @@ export const SearchBox = <T,>({
       </div>
       {isFocused && (
         <div className={styles.optionsAnchor}>
-          <div className={styles.optionsBox}>
+          <div className={styles.optionsBox} ref={optionsRef}>
```
anttimaki, line 106: I guess this approach is fine, but I'm wondering if we could bind the ref directly to the element were interested in with something like this:

```
{options.map((option, i) => (
  <button
    ref={i === 0 ? optionRef : null}
    className={styles.optionButton}
    key={keyExtractor(option)}
    onKeyDown={(e) => listKeyDownHandler(option, e)}
    onClick={(e) => onClickHandler(option, e.currentTarget)}
  >
    {renderOption(option)}
  </button>
)}
```

Haven't tested, might not work 🤷🏻

anttimaki (approved):
Approved as is, use your own judgement regarding the new comments.

apps/sertra/components/mod-selector/list.tsx
MythicManiac: I know my IDEs at least auto-insert the braces regardless of value type, so as with any other formatting rules it'd be best if whatever formatting is linted for is also automated. In other words, a lint rule alone is not a sufficient change, the formatting should also be handled by prettier. If both can't be done then I don't think we should have a lint rule.

PR #515, opened 2022-09-06
Make Sertra Dockerfile buildable without a build arg

Oksamies:
There were some bugs also. Look at the first commit.


MythicManiac:
Lint ignore comment removals were needed as the build process throws an error for unknown lint rule ignore, in case that's confusing during review

anttimaki (commented):
* "Kill bugs :D" is a poor commit message. Especially since it e.g. includes removing linting comments (imagine someone checking the git history and wondering what that has to do with bugs).
* While you were instructed to do so, removing those linting comments seemed like sweeping the problem under a rug. I've opened a stack of PRs fixing the root problem, including a replacement PR for this one: https://github.com/thunderstore-io/thunderstore-ui/pull/519 . So IMO this one could be closed.

MythicManiac:
Closing due to the above (other PR fixed this)

PR #516, opened 2022-09-06
Make Sertra detail view use actual mod data

Oksamies:
refs TS-712


apps/sertra/pages/server/[id].tsx
```diff
@@ -30,8 +30,34 @@ export const getStaticProps: GetStaticProps<
...
   const listingId = context.params!.id; // eslint-disable-line @typescript-eslint/no-non-null-assertion
   const res = await fetch(ApiURLs.ServerDetail(listingId));
+  const mods_res = await fetch(TsApiURLs.V1Packages("v-rising"));
   const data = await res.json();
+  const mods_data = await mods_res.json();
 
+  const updated_listing_mods_data = [];
+  for (const mod_ref of data.mods) {
+    for (const mod of mods_data) {
+      const [mod_ref_owner, mod_ref_name, mod_ref_version] = mod_ref.split(
```
Comment (MythicManiac) [inline apps/sertra/pages/server/[id].tsx:40]: This isn't a valid way to parse package references unfortunately (I wish it was). The data includes a `full_name` property for each package version, so I'd suggest you
1. Check if the package `full_name` matches the start of the mod definition.
2. If yes, compare against all versions of the partially matching package until you find a full match.

Comment (MythicManiac) [reply apps/sertra/pages/server/[id].tsx:40]: If you really want to implement a parser, you can copy the implementation from here: https://github.com/thunderstore-io/Thunderstore/blob/master/django/thunderstore/repository/package_reference.py but I'd recommend against it as it's unnecessary complexity in this case.

Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:40]: I'd like to do 1, but the problem is that owner and the mod name are flipped. I'll figure something out that doesn't need the parser and doesn't look like hell

Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:40]: Hmm, hold on they aren't flipped, dunno what I thought yesterday, nvm.

Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx]: @MythicManiac Please go over the newest implementation. I did both ways, so that you(or someone) could tell me if it's a correctish way to do the parsing?

Comment (MythicManiac) [reply apps/sertra/pages/server/[id].tsx]: I'm still not sure why are you attempting to parse it instead of doing simple string comparison? If you do want to parse it this is definitely not how we want it written in code; there should be a standalone function/class handling the parsing rather than inlining it like it has been now written.

That said unless there's a reason you need to do the parsing, I'd just stick with the string comparison approach.

Comment (MythicManiac) [reply apps/sertra/pages/server/[id].tsx]: You should also move the package data fetching to some standalone class/function rather than inlining it to that specific view, chances are it's going to be re-used elsewhere too and it makes sense to cache it, which isn't really possible if it's inlined the way it is currently.

That's perhaps outside the scope of this PR however, but noting it anyway for future reference.

MythicManiac (changes requested):
See the comment thread here: https://github.com/thunderstore-io/thunderstore-ui/pull/516#discussion_r965154711

apps/sertra/pages/server/[id].tsx
```diff
@@ -30,8 +30,34 @@ export const getStaticProps: GetStaticProps<
...
+              version: version.version_number,
+              icon_url: version.icon,
+              description: version.description ?? null,
+            });
+          }
+        }
+      }
+    }
+  }
+  data.mods = updated_listing_mods_data;
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:72]: I think mutating the `data`, which is a fetch response(?) is a bad idea. The processed data should be stored in state separately.

apps/sertra/pages/server/[id].tsx
```diff
@@ -66,16 +92,15 @@ const ServerDetail: React.FC<{ detail_listing: ServerListingDetailData }> = ({
...
             <h2 className={styles.sectionTitle}>Mods</h2>
             <div>
-              {/* TODO: So we are missing the name, description and artifacts, from the API data */}
-              {/* {data.mods.map((x) => (
+              {detail_listing.mods.map((x) => (
                 <ModCard
                   key={x.name}
                   name={x.name}
-                  description={x.description}
+                  owner={x.owner}
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:111]: When object keys nicely match component props, you can do something like this:
`<ModCard key={x.name} {...x} description={x.description ?? null} />`

`{...x}` will override any props defined before it with values from `x`. Since there's no `x.key`, the key prop won't be overwritten.
Props defined after `{...x}` will override any props defined by `{...x}`. So even though there's `x.description`, redefining the prop will overwrite the value received from `x`. **Note**: I can't recall if this works here since the type of `x.description` might not match the type of `ModCard.description`. Worth checking out anyway, or at least for you to know in the future (in case you did not know this already).

apps/sertra/components/ModCard.tsx
```diff
@@ -17,7 +17,11 @@ export const ModCard: React.FC<ModCard> = ({
 }) => (
   <div className={styles.mod}>
     <div className={styles.column}>
-      <img className={styles.imagePlaceholder} src={icon_url}></img>
+      {icon_url ? (
+        <img className={styles.imageIcon} src={icon_url}></img>
+      ) : (
+        <img className={styles.imageIcon} src="/favicon.ico"></img>
```
anttimaki, line 23: Isn't that `favicon.ico` the Vercel's default icon? Or has that been replaced already. Anyhow I'd prefer `public/ts-logo.svg` as a quick solution. Proper solution is probably asking grav whether we want to use the same approach as the new upcoming UI, or have some other placeholder for sertra.

Also since the only thing changing seems to be the `src`, the ternary operator is unnecessarily verbose. This would be more readable IMO:
``

Also also I think you should fix the things ESLint is complaining about?

apps/sertra/components/ModCard.module.css
```diff
@@ -17,7 +17,7 @@
   flex-direction: column;
 }
 
-.imagePlaceholder {
+.imageIcon {
```
anttimaki, line 20: "imageIcon" sounds stupid IMO. Maybe use "modImage" or "modIcon"?

anttimaki (commented):
My two cents. Didn't review the "parsing stuff" since you and Mythic seem to have that part under control/construction.

apps/sertra/components/ModCard.tsx
Oksamies: Ah I didn't know I could do that one line ternary thing 👍 Also switched to using the nexts Image component. I'm not sure what are the benefits of that, but at least they had an example on nextjs.com.

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:111]: Seems like `<ModCard key={x.name} {...x} />` does work here, just had to define type for `x` correctly. Didn't know this way was possible. Definitely is a lot simpler and cleaner representation.

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:72]: Good point, dunno why I didn't just plug the new array into the return :D Fixd, this 👍

Oksamies:
I've addressed all the comments, but I'm hitting a wall with one of the suggestions Antti made (https://github.com/thunderstore-io/thunderstore-ui/pull/516#discussion_r965575108) please help, I don't have the mental energy to look at this anymore.

image

apps/sertra/components/PackageDataFetcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+      icon_url: null,
+      description: null,
+    };
+    for (const mod of mods) {
+      if (mod_ref.startsWith(mod.full_name)) {
+        new_mod_data.name = mod.name;
+        new_mod_data.owner = mod.owner;
+        new_mod_data.version = null;
+        for (const version of mod.versions) {
+          if (mod_ref.endsWith(version.version_number)) {
```
MythicManiac, line 50: You **need** to do a full string comparison before you can be sure the mod actually matches, the current checks can let false positives through.

Likewise you can't assume the mod to match if just the start matches (on line 45), you can simply use that to reduce the amount of versions you enumerate over (just enumerate the mods with a matching start rather than _all_ versions). In other words you can't populate any data to the mod data before you have a full match, or otherwise you risk displaying invalid data.

I'd suggest you first flatten the entire mods list to a dictionary (you can use a [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) or a [record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type)), and then simply look up if a version exists there, rather than looping through everything multiple times.

apps/sertra/components/PackageDataFetcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+            new_mod_data.version = version.version_number;
+            new_mod_data.icon_url = version.icon ?? null;
```
MythicManiac, line 52: At this point there should be no need to use the null coalescing operator, all versions have an icon and a description.

apps/sertra/components/PackageDataFetcher.tsx
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useState } from "react";
```
MythicManiac, line 1: This whole file is in the wrong location, and uses the wrong file extension.

The `components` directory should only be used for react components, and since this is simply typescript code, it should be placed elsewhere. Likewise there's no reason to use the `.tsx` extension (which is for react templates), use `.ts` instead (plain typescript).

What this actually is, is a hook, and it's fairly common to have one or more modules called `hooks` somewhere in a react project, which contain custom hook implementations.

apps/sertra/components/ModCard.tsx
```diff
@@ -2,21 +2,29 @@ import styles from "./ModCard.module.css";
 
 interface ModCard {
   name: string;
+  owner: string;
   description: string;
+  version: string;
+  icon_url: string;
```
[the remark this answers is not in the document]
MythicManiac: Assuming we get any mod data, this can never be null. All packages are required to have an icon.

apps/sertra/components/PackageDataFetcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [listing, setListing] = useState<TempModRefType>();
+  useEffect(() => {
+    async function fetchListing() {
+      return await fetch(ApiURLs.ServerDetail(listingId));
+    }
+    async function jsonifyListing() {
+      setListing(await (await fetchListing()).json());
+    }
+    jsonifyListing();
+  });
```
MythicManiac, line 22: Missing the second parameter (to the `useEffect` function), which should be `[listingId]` in this case. Look up how it works in case you're unfamiliar with what it does.

apps/sertra/components/PackageDataFetcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [mods, setMods] = useState<TSListingMod[]>([]);
+  useEffect(() => {
+    async function fetchMods() {
+      return await fetch(TsApiURLs.V1Packages("v-rising"));
+    }
+    async function jsonifyMods() {
+      setMods(await (await fetchMods()).json());
+    }
+    jsonifyMods();
+  });
```
MythicManiac, line 33: Also missing the second parameter, which in this case should simply be an empty list.

apps/sertra/components/PackageDataFetcher.tsx
MythicManiac, line 22: Why so verbose? There's no need to define named functions like this

```ts
useEffect(() => {
  fetch(ApiURLs.ServerDetail(listingId)).then((resp) => {
    setListing(resp.json());
  });
}, [listingId]);
```

Same applies for the other `useEffect` below.

MythicManiac (changes requested):
See inline comments

apps/sertra/pages/server/[id].tsx
Comment (anttimaki) [reply apps/sertra/pages/server/[id].tsx]: Yeah, it's simpler, but use your own judgement when to use it. Since it obfuscates the props that are actually passed (or left out), it might be a bit confusing in some situations.

Also probably `props` or `modProps` would be a better variable name than `x` now that there's not that much repetition, but that's just an opinion.

Mark this comment resolved once you've read it.

apps/sertra/components/ModCard.tsx
```diff
@@ -8,6 +9,14 @@ interface ModCard {
   icon_url: string;
 }
 
+interface modImageLoaderInputProps {
+  src: string;
+}
+
+const modImageLoader = ({ src }: modImageLoaderInputProps) => {
```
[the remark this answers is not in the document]
Oksamies: Good points, still getting used to finding the types 👍 Seems like the Image thing doesn't work without a width, even though its not used in the loader. The example here is a bit weird for me also (https://nextjs.org/docs/api-reference/next/image#loader), since it's just slapping the resizing at the end of the url, which I guess should be a feature on the images hosting side 🤔

Anyways I do think, the latest version (soon to be updated to this branch), is fine enough for the Alpha.

Oksamies:
@MythicManiac @anttimaki I suggest looking at the changed files, instead of the commits, since the commits are quite a mess now. After you guys approve the changes, I'll cleanup the commit history to be something readable. Ty in advance ❤️

apps/sertra/components/ModCard.tsx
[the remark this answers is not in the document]
anttimaki: I guess the 3rd point in my previous comment was vaguely worded (too many "it" making it unclear what I'm referring to).

What I meant was that it looks the `loader` prop isn't needed. And since your implementation just returns the `src` parameter as is, I think defining the loader func and including the prop only adds noise. So I would remove the `loader` prop if the image works as expected without it.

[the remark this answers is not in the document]
Oksamies: Ah yes, seems like the loader can be removed as long as it has that `unoptimized` as `true`, but I'm not sure why it works with it 🤷 I tried it without the loader, before adding the unoptimized, but back then it didn't work.

Might have something to do with the fact that "optimization" is probably configured  manually in the loader, depending on the hosting parties way of delivering "optimized" images? 🤷 Anyways should be good now

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx]: Changed the name 👍 And good point, I'll have to keep it mind, when using some more complex component.

apps/sertra/components/ModCard.tsx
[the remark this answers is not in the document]
anttimaki: I'm not thrilled with the amount of guesswork and shrugging emoji here. Created TS-1054 for figuring this out so it's not a blocker for this PR.

anttimaki (approved):
Approving the changes done based on *my comments*. The PR as a whole isn't approved yet by all reviewers.

apps/sertra/pages/server/[id].tsx
```diff
@@ -52,37 +57,29 @@ const ServerDetail: React.FC<{ detail_listing: ServerListingDetailData }> = ({
...
-                <ModCard
-                  key={x.name}
-                  name={x.name}
-                  description={x.description}
-                />
-              ))} */}
-              {detail_listing.mods.map((x) => (
-                <ModCard key={x} name={x} description={x} />
+              {mods_data.map((modProps) => (
+                <ModCard key={modProps.name} {...modProps} />
```
Comment (MythicManiac) [inline apps/sertra/pages/server/[id].tsx:75]: `key` must be guaranteed to be unique, so you should use the full name for it

apps/sertra/api/PackageDataFetcher.ts
```diff
@@ -0,0 +1,49 @@
...
+    };
+    const mod = all_versions.get(mod_ref);
+    if (mod) {
+      new_mod_data.name = mod.name;
+      new_mod_data.owner = mod.owner;
+      new_mod_data.version = mod.version_number;
+      new_mod_data.icon_url = mod.icon;
+      new_mod_data.description = mod.description;
+    }
+    updated_listing_mods_data.push(new_mod_data);
```
MythicManiac, line 43: You should be able to simplify this quite a bit, something like:

```typescript
updated_listing_mods_data = listing.mods.map((x) => all_versions.get(x) ?? {
  name: mod_ref,
  owner: null,
  version: null,
  icon_url: null,
  description: null,
});
```

apps/sertra/api/PackageDataFetcher.ts
```diff
@@ -0,0 +1,49 @@
...
+  const mods = await mod_data_fetch;
+
+  const all_versions = new Map();
+  mods.map((mod: TSListingMod) => {
+    mod.versions.map((version) => {
+      const version_with_owner = version as TSListingModVersionWithOwner;
+      version_with_owner.owner = mod.owner;
+      all_versions.set(version_with_owner.full_name, version_with_owner);
+    });
+  });
```
MythicManiac, line 25: Here on the other hand I'd prefer that you use a for loop, as map is usually used when you need the return value (you don't here, you're just looping)

MythicManiac (approved):
Implementation-wise looks correct now, left some minor comments

PR #530, opened 2022-09-20
Add new favicon

Cyanideankst:
refs TS-846


apps/sertra/pages/index.tsx
```diff
@@ -27,7 +27,27 @@ const Home: NextPage<{ swrFallback: { [key: string]: unknown } }> = ({
...
+          />
+          <link
+            rel="icon"
+            type="image/png"
+            sizes="16x16"
+            href="/favicon-16x16.png"
+          />
+          <link rel="manifest" href="/site.webmanifest" />
+          <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5" />
+          <meta name="msapplication-TileColor" content="#da532c" />
```
MythicManiac, line 49: @gkurck can you verify these color choices are intended?

MythicManiac (approved):
Looks good otherwise, but let's wait for @gkurck to give input for the color choice 👍

apps/sertra/pages/index.tsx
gkurck: I can verify they are **not** intended. The correct value for both is `#3d70c0`.

Oksamies:
@MythicManiac Seems like the test pre-commit runner is getting stuck on something that at least my linter doesn't complain about.

Oksamies (approved):
All good and can merge, after the pre-commit runner has been looked at.

apps/sertra/public/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/mstile-150x150.png"/>
+            <TileColor>#da532c</TileColor>
```
gkurck, line 6: This one also has an incorrect value, correct value is `#3d70c0`.

gkurck (changes requested):

gkurck (approved):

PR #531, opened 2022-09-20
Sertra: Fix incorrect font-weight display issue

gkurck:
Reference TS-1064

Fixes the incorrect displaying of font-weights by declaring the `font-weight` range for `@font-face` -rules in `styles/globals.css`.


anttimaki (approved):
👍🏻

PR #534, opened 2022-09-22
Add meta stuff to submit and detail views

Oksamies:
refs TS-1077


apps/sertra/components/HeadWrapper.tsx
```diff
@@ -0,0 +1,18 @@
+import Head from "next/head";
+
+interface HeadProps {
+  title: string;
+  description: string;
+}
+
+export const HeadWrapper: React.FC<HeadProps> = ({ title, description }) => (
+  <Head>
+    <title>{title}</title>
```
anttimaki, line 10: This could be ```{`${title} | Thunderstore`}``` so we don't need to repeat it everywhere.

apps/sertra/pages/create.tsx
```diff
@@ -6,6 +6,7 @@ import { getPackageList } from "../api/hooks";
 import { ApiURLs, TsApiURLs } from "../api/urls";
 import { ModSelectorModal } from "../components/mod-selector/modal";
 import styles from "../styles/SubmitServer.module.css";
+import { HeadWrapper } from "../components/HeadWrapper";
```
anttimaki, line 9: Import order plz.

apps/sertra/pages/server/[id].tsx
```diff
@@ -13,6 +13,7 @@ import { FetchListingData } from "../../api/PackageDataFetcher";
 import { ServerInfo } from "../../components/ServerInfo";
 import { ServerInstructions } from "../../components/ServerInstructions";
 import styles from "../../styles/ServerDetail.module.css";
+import { HeadWrapper } from "../../components/HeadWrapper";
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:16]: Import order plz.

apps/sertra/pages/server/[id].tsx
```diff
@@ -51,6 +52,10 @@ const ServerDetail: React.FC<{
 }> = ({ listing_data, mods_data }) => {
   return (
     <div className={styles.container}>
+      <HeadWrapper
+        title={`${listing_data.name} - V-Rising Servers | Thunderstore`}
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:56]: Could we read the game name from `listing_data`?

apps/sertra/pages/server/[id].tsx
```diff
@@ -51,6 +52,10 @@ const ServerDetail: React.FC<{
...
+        description={listing_data.description}
```
Comment (anttimaki) [inline apps/sertra/pages/server/[id].tsx:57]: Just a thought, but are we certain we want to use user inputs (description here, name above) as the meta content? E.g. racist content is obviously bad when shown on our website, but can be extra bad when shown on those embeds on other sites?

anttimaki:
Also I think the commit message is bad. What is "meta stuff"? Why isn't the header wrapper mentioned? Now that I think of it, should it be documented in README oslt so other devs know to use it?

apps/sertra/components/HeadWrapper.tsx
Oksamies: True, I'll add it there 👍

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:56]: Yes but it's lower case, I'll add it from there tho. Better to have dynamically.

apps/sertra/pages/server/[id].tsx
Comment (Oksamies) [reply apps/sertra/pages/server/[id].tsx:57]: This was a quick suggest by @MythicManiac , I don't have a good understanding/knowledge what should/could be used here. What do you think @MythicManiac ?

Oksamies:
> Also I think the commit message is bad. What is "meta stuff"? Why isn't the header wrapper mentioned? Now that I think of it, should it be documented in README oslt so other devs know to use it?

Fixed comments and added a bit of information to the readme.

I'll be waiting for a answer from @MythicManiac to the matter of using user inputs.

apps/sertra/pages/server/[id].tsx
Comment (MythicManiac) [reply apps/sertra/pages/server/[id].tsx]: That's the purpose of the meta fields, we'll just need to deal with moderating them. The alternative is not having any content-bound info available in any embeds, which is also bad.

PR #535, opened 2022-09-22
Improve PackageDataFetcher logics

Oksamies:
Addressing three comments made by Mythic
https://github.com/thunderstore-io/thunderstore-ui/pull/516

refs TS-1074


apps/sertra/api/PackageDataFetcher.ts
```diff
@@ -13,35 +13,27 @@ export const FetchListingData = async (listingId: string) => {
     x.json()
   );
   const listing = await listing_data_fetch;
-  const mods = await mod_data_fetch;
+  const mods = (await mod_data_fetch) as TSListingMod[];
 
   const all_versions = new Map();
-  mods.map((mod: TSListingMod) => {
-    mod.versions.map((version) => {
+  for (const mod of mods as Array<TSListingMod>) {
```
MythicManiac, line 19: I believe there doesn't have to be any type casts here (`as X` statements). line 16 already takes care of the typing

apps/sertra/api/PackageDataFetcher.ts
```diff
@@ -13,35 +13,27 @@ export const FetchListingData = async (listingId: string) => {
     x.json()
   );
   const listing = await listing_data_fetch;
-  const mods = await mod_data_fetch;
+  const mods = (await mod_data_fetch) as TSListingMod[];
 
   const all_versions = new Map();
```
anttimaki, line 18: `new Map<string, TSListingModVersionWithOwner>();` would be nice.

apps/sertra/api/PackageDataFetcher.ts
anttimaki, line 19: I know no one asked, but a FP oriented way (that probably works):
```typescript
  const allVersions = new Map(
    mods
      .map((mod): [string, TSListingModVersionWithOwner][] =>
        mod.versions.map((ver) => [ver.full_name, { ...ver, owner: mod.owner }])
      )
      .flat()
  );
```

`[string, TSListingModVersionWithOwner][]` is required so TS knows the inner `.map` is returning tuples, not arrays. It would be more proper to define the return type of the inner `.map` rather than the outer one, but linting rules would cause stuff to get split into multiple lines which IMO makes the code less readable. On the plus side, this makes the resulting Map automatically typed.

MythicManiac: Instead of reviewing, I made some changes on top here myself. This one had crossed my mind as well since in python this approach would be much cleaner to read, but IMO in TypeScript this is harder to read so I left it out.

MythicManiac:
@anttimaki I'm setting you as a reviewer here since I contributed code myself. Squash merge this straight away if it looks alright, you can make adjustments to the commit message directly that way.

anttimaki (approved):

PR #537, opened 2022-09-26
Add a component for handling meta tags project wide

Oksamies:
HeadWrapper component that is.

I fucked up with PR #534 so I needed to create a new PR.

refs TS-1077

### DIFF
--- a/apps/sertra/README.md
+++ b/apps/sertra/README.md
@@ -50,3 +50,9 @@ h1:before {
   background-color: currentColor;
 }
 ```
+
+## Adding meta tags to a page
+
+When possible possible component in `HeadWrapper.tsx` should be used to add meta tags. This way the meta tag handling is centralized and can be edited easily for the whole site.
+
+If there are tags missing from the `HeadWrapper` please improve it.
--- a/apps/sertra/README.md
+++ b/apps/sertra/README.md
@@ -53,6 +53,6 @@ h1:before {
 
 ## Adding meta tags to a page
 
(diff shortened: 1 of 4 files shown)


apps/sertra/README.md
```diff
@@ -50,3 +50,9 @@ h1:before {
   background-color: currentColor;
 }
 ```
+
+## Adding meta tags to a page
+
+When possible possible component in `HeadWrapper.tsx` should be used to add meta tags. This way the meta tag handling is centralized and can be edited easily for the whole site.
```
anttimaki, line 56: possible possible

apps/sertra/components/HeadWrapper.tsx
```diff
@@ -0,0 +1,18 @@
+import Head from "next/head";
+
+interface HeadProps {
+  title: string;
```
MythicManiac, line 4: This is misleading and could easily be assumed to be user-visible title. `metaTitle` would be more fitting. Same for `description`, `metaDescription` would be more clear if you just see the call site.

apps/sertra/pages/create.tsx
```diff
@@ -56,6 +57,10 @@ const SubmitServer: NextPage<{ swrFallback: { [key: string]: unknown } }> = ({
 
   return (
     <SWRConfig value={{ fallback: swrFallback }}>
+      <HeadWrapper
+        title="Submit Server"
```
anttimaki, line 61: You've changed the names of the props but haven't updated this and another place where the component is used.

anttimaki (changes requested):

PR #546, opened 2022-09-29
@thunderstore/sertra: fix rehydration broken by FA icon

anttimaki:

PR #545, opened 2022-09-29
@thunderstore/sertra: show selected mods on server submit form

anttimaki:

PR #547, opened 2022-09-29
@thunderstore/sertra: code cleanup

anttimaki:


anttimaki:
* master
  * **PR #545** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/545" target="_blank">Graphite</a> 
    * **PR #546** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/546" target="_blank">Graphite</a> 
      * **PR #547** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/547" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/547?utm_source=stack-comment).

VilppeRiskidev (approved):

apps/sertra/components/mod-selector/modal.module.css
```diff
@@ -25,66 +25,64 @@
...
   flex-direction: column;
-  padding: 24px 32px 16px;
+  padding: 1.5rem 2rem 1rem;
+  background-color: var(--bg-light);
 }
 
 .headerTitle {
   display: flex;
-  flex-direction: row;
+  flex-direction: row-reverse;
```
gkurck, line 36: Is there a reason the for the close-button and title being in reverse order in the DOM compared to the rendered output, thus needing this?

apps/sertra/components/mod-selector/modal.module.css
```diff
@@ -25,66 +25,64 @@
...
   align-items: center;
   justify-content: space-between;
+  margin-bottom: 1rem;
+}
+
+.headerTitle:last-child {
```
gkurck, line 42: Consider using a more directed approach for reusability. This works now, but might introduce problems if the header content changes.

apps/sertra/components/mod-selector/modal.module.css
```diff
@@ -25,66 +25,64 @@
...
   width: 1.375rem;
   height: 1.375rem;
 }
 
-.headerContent {
-  margin-top: 16px;
-}
-
 .content {
+  height: 38rem;
```
gkurck, line 52: This works for this alpha release, but we should consider resizing to a different sized screens.

An example what we could do here: Set padding to the root parent, set the modal width and height to 100% with max constraints, remove explicit height from the content area and have it take up the remaining space automatically with `flex:1`.

gkurck (approved):
Looks good to me for alpha.

Left a couple of comments, mostly to note down some ideas for later.

apps/sertra/components/mod-selector/modal.module.css
anttimaki: We did this change during the pair (group) work session on Thursday and actually shortly pondered the ramifications of having the close button before the title on the source order. We didn't come up with any reasons why this would be a bad thing and decided to go with it. But basically this is because in the component the title is optional (for some reason) and it's easier to "fix" the close button to right this way. So, what are the cons in this approach (besides it being mildly hacky)?

apps/sertra/components/mod-selector/modal.module.css
anttimaki: On the other hand, YAGNI. And if we need to change this, changing it to work another way around won't take practically any time.

apps/sertra/components/mod-selector/modal.module.css
anttimaki: Agreed, and I listed this as one of the tasks Oksamies could to on Friday if he had the time. (The height was already fixed, this PR only changed it from `px` to `rem`.)

PR #548, opened 2022-10-02
Add filtering into the search component

Oksamies:
Add filtering by the passed variable keys of the objects that are being filtered

### DIFF
--- a/apps/sertra/components/mod-selector/search.tsx
+++ b/apps/sertra/components/mod-selector/search.tsx
@@ -1,5 +1,4 @@
 import { ReactNode, useRef, useState } from "react";
-import { ModPackage } from "../../api/models";
 import styles from "./search.module.css";
 
 interface SearchBoxProps<T> {
@@ -8,6 +7,7 @@ interface SearchBoxProps<T> {
   keyExtractor: (option: T) => string | number;
   options: T[];
   onSelect?: (option: T) => void;
+  searchVars: string[];
 }
 
 export const SearchBox = <T,>({
@@ -16,18 +16,19 @@ export const SearchBox = <T,>({
   keyExtractor,
   options,
   onSelect,
+  searchVars,
 }: SearchBoxProps<T>) => {
   const [isFocused, setFocused] = useState<boolean>(false);
 
   const [filteredOptions, setfilteredOptions] = useState(options);
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const results = options.filter((option) => {
-      if (option.ownerName.toLowerCase().includes(e.target.value.toLowerCase()))
-        return option;
-      if (
-        option.packageName.toLowerCase().includes(e.target.value.toLowerCase())
-      )
-        return option;
+      for (const searchVar of searchVars) {
+        if (
+          option[searchVar].toLowerCase().includes(e.target.value.toLowerCase())
+        )
+          return option;
+      }
     });
     setfilteredOptions(results);
   };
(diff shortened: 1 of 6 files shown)


Oksamies:
I'm not sure if the target branch or my branch is somehow f'd up 🤔

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -8,6 +7,7 @@ interface SearchBoxProps<T> {
   keyExtractor: (option: T) => string | number;
   options: T[];
   onSelect?: (option: T) => void;
+  searchVars: string[];
```
anttimaki, line 10: This now allows us to pass any string in `searchVars` prop, including ones that are not actually keys of type `T`. Which means that `option[searchVar]` below will return undefined, and calling `.toLowerCase()` breaks the execution. At least for me the IDE actually warns about `option[searchVar]`, but with the state of linting/configs in this repo YMMV.

Anyways, to fix the issue, use `searchVars: (keyof T)[];` instead. TypeScript will now ensure that the parameters passed are actually keys used in `T`. In practice this means that if there's e.g. a typo at `modal.txs`'s `searchVars={["ownerName", "packageName"]}`, IDE will warn about it and the build process will fail. This re-generalizes the component (referring to your first commit message here).

Also IMO `searchKeys` might be a better name, since the valuea aren't really variable.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -16,18 +16,19 @@ export const SearchBox = <T,>({
   keyExtractor,
   options,
   onSelect,
+  searchVars,
 }: SearchBoxProps<T>) => {
   const [isFocused, setFocused] = useState<boolean>(false);
 
   const [filteredOptions, setfilteredOptions] = useState(options);
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const results = options.filter((option) => {
```
anttimaki, line 25: Maybe do something like `const searchValue = e.target.value.toLowerCase()` outside the `.filter()`, which is practically a `for`-loop. The operation is is lightweight, so maybe this doesn't offer any optimization benefits, but it still feels the right way to do it.

apps/sertra/components/mod-selector/search.tsx
```diff
@@ -16,18 +16,19 @@ export const SearchBox = <T,>({
...
-      if (option.ownerName.toLowerCase().includes(e.target.value.toLowerCase()))
-        return option;
-      if (
-        option.packageName.toLowerCase().includes(e.target.value.toLowerCase())
-      )
-        return option;
+      for (const searchVar of searchVars) {
+        if (
+          option[searchVar].toLowerCase().includes(e.target.value.toLowerCase())
```
anttimaki, line 28: My first comment gives us a better change that `option[searchVar]` is actually a thing. But it might be e.g. an int, bool, or undefined, e.g. if the interface declares the value an optional string. Before we can call `.toLowerCase`, which only exists for strings, we need to check that `option[searchVar]` is a string:

```
        const field = option[searchVar];
        if (
          typeof field === "string" &&
          field.toLowerCase().includes(searchValue)
        ) {
          return option;
        }
```

And please use braces around the block even if it's just a single line. That's something that should be caught by linter, but you know the current situation.

anttimaki: As a side note, this search implementation doesn't work if user tries to search with something like "teamname modname". But I think this is good enough for the MVP.

anttimaki (changes requested):
I'm pretty sure the problem you mention in your comment is caused by your branch starting from a wrong place. Try running `git rebase -i TS-1060-code-cleanup` and dropping the commits that are not actually part of this PR.

PR #560, opened 2022-10-24
Cyberstorm

VilppeRiskidev:


packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,29 @@
...
+  buttonStyle?: string;
+  onClick?: MouseEventHandler<HTMLButtonElement>;
+}
+
+/**
+ * Cyberstorm Button component
+ */
+export const Button: React.FC<ButtonProps> = (props) => {
+  const { label, leftIcon, rightIcon, buttonStyle, onClick } = props;
+  const stylesImport = styles;
```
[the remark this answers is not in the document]
VilppeRiskidev: I wonder what would be the best way to use the styles🤔 If the styles is not called, the styles don't seem to be imported. (even if the variable is never used... weird)

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,97 @@
...
+  --text-color: var(--color-white);
+  background-image: var(--color-gradient-blue-green--darker),
+  var(--color-gradient-blue-green);
+  border: 2px solid;
+  border-image-source: var(--color-gradient-blue-green);
+  box-shadow: 0 0 var(--size-medium) 0 var(--color-cyber-green--20);
+}
+
+.button__specialGreen:hover {
+  background-image: var(--color-gradient-blue-green--dark),
```
gkurck, line 77: I don't think this is animatable as is. I created a [Codepen entry](https://codepen.io/gravija/pen/poKzGYJ) to demonstrate how we could achieve smooth transition with gradients, which basically involves animating the opacity of a pseudo element. Let me know what do you think.

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,97 @@
...
+  flex-direction: row;
+  justify-content: center;
+  align-items: center;
+  flex: none;
+  gap: var(--gap-medium);
+
+  padding: var(--padding-medium-small) var(--padding-medium);
+  border-radius: var(--size-small);
+
+  background-clip: padding-box, border-box;
```
gkurck, line 12: This might be redundant, and can be applied straight to the classes below where they are needed - you can also use `background` shorthand since we want to override bgcolor anyway.

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return (
+    <button className={"button button__" + buttonStyle} onClick={onClick}>
+      {leftIcon}
```
gkurck, line 21: Might be a good idea to pass some default button related classes for the icon elements.

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -0,0 +1,143 @@
...
+  --color-gradient-blue-green--dark: linear-gradient(283.56deg, #0B4162 0%, #205740 100%);
+  --color-gradient-blue-green: linear-gradient(283.56deg, #1CA3F5 0%, #23FFB0 100%);
+  --color-gradient-blue-green--light: linear-gradient(283.56deg, #49B5F7 0%, #73E1B2 100%);
+  --color-gradient-purple-green--darker: linear-gradient(283.56deg, #1F0A2E 0%, #102B20 100%);
+  --color-gradient-purple-green--dark: linear-gradient(283.56deg, #2F0E44 0%, #0A4C35 100%);
+  --color-gradient-purple-green: linear-gradient(283.56deg, #9D30E4 0%, #23FFB0 100%);
+  --color-gradient-purple-green--light: linear-gradient(283.56deg, #B159E9 0%, #73E1B2 100%);
+
+
+  --size-medium: 1rem;
```
gkurck, line 123: Might be good idea to adopt the naming style of `s`, `m`, `l` OSLT to a) make names shorter and b) allowing the use of x's to make something even bigger/smaller if needed, such as `xs` or `xxxl`.

gkurck (commented):
Left a few comments, let me know what you guys think.

packages/cyberstorm/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "dev": "tsc --watch"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.17.0",
+    "@emotion/react": "^11.4.1",
```
anttimaki, line 18: IIRC `@emotion` is a dependency for Chakra UI. Same goes for `framer-motion`. Some others, such as `downshift` and `s-ago` are used by the old components, but not really needed here. Please clean the list to only include the actual dependencies.

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { MouseEventHandler } from "react";
+import styles from "./componentStyles/Button.module.css";
+
+export interface ButtonProps {
+  label?: string;
+  leftIcon?: string;
```
anttimaki, line 6: The icons are still typed as strings.

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { MouseEventHandler } from "react";
+import styles from "./componentStyles/Button.module.css";
+
+export interface ButtonProps {
```
anttimaki, line 4: All the props are optional, so the button generated by passing no props is a bit of sadge. Should we show some default text if label nor icons are passed as props? Just a thought, not necessarily required as a part of this PR.

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -0,0 +1,143 @@
...
+  --size-medium-small: 0.625rem;
+  --size-small: 0.5rem;
+
+  --gap-medium: var(--size-medium);
+
+  --padding-medium: var(--size-medium);
+  --padding-medium-small: var(--size-medium-small);
+}
+
+.label {
```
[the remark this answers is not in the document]
anttimaki: I guess this is for the button component's text. I'd argue that the text doesn't require a wrapper div, If it does, this should be defined in button's CSS module and not as a global. If we're absolutely sure it should be global, it should be named more descriptively or include a comment describing what's it for. But I'd argue that a random developer won't find the class from here and thus it won't be used globally.

Also neither font-size or line-height should use pixels, I guess?

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Cyberstorm Button component
+ */
+export const Button: React.FC<ButtonProps> = (props) => {
+  const { label, leftIcon, rightIcon, buttonStyle, onClick } = props;
+  const stylesImport = styles;
+
+  return (
+    <button className={"button button__" + buttonStyle} onClick={onClick}>
```
anttimaki, line 20: This should define `type="button"`. Or allow passing it as a prop, but defaulting to `"button"` if no prop is passed.

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
+/**
+ * Implementation-agnostic context provider for navigation.
+ *
+ * @thunderstore/components aims to be reusable. Therefore the links
```
anttimaki, line 4: Copy-pasted package name.

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+// required since the stripping is done during runtime.
+export const thunderstoreLinkProps: ThunderstoreLinkProps = {
+  community: "",
+  namespace: "",
+  package: "",
+  team: "",
+};
+
+// Accepting any and all props is required to keep the linking
+// implementation-agnostic. E.g. an implementation with Chakra has all
```
anttimaki, line 42: Not sure we should refer to Chakra here anymore.

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+// kinds of custom properties.
+export interface AnyProps {
+  [key: string]: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  queryParams?: string;
+}
+
+type NoRequiredProps = (props: AnyProps) => RE | null;
+type PackageProps = { community: string; namespace: string; package: string };
+
+// STEP 3 of adding bew link definitions:
```
anttimaki, line 52: Copy-pasted typo :D

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+  CommunityPackages: noop,
+  Index: noop,
+  Package: noop,
+  PackageDependants: noop,
+  PackageUpload: noop,
+  Team: noop,
+};
+
+// OPTIONAL STEP 5 of adding new link definitions:
+// Add the link to ./Links.tsx for easier importing.
```
anttimaki, line 92: This refers to Links.tsx, which is not included in this PR. Maybe it should be.

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,97 @@
...
+  --text-color: var(--color-white);
+  outline: 2px solid var(--color-border-2);
+}
+
+.button__defaultWithBorder:hover {
+  --bg-color: var(--color-surface--4);
+  outline: 2px solid var(--color-border-3);
+}
+
+.button__specialGreen {
```
anttimaki, line 67: FYI the storybook shows odd corners for these buttons in Firefox: 
image

packages/cyberstorm/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const isDevelopment = process.env.NODE_ENV === "development";
```
anttimaki, line 1: Are these actually used?

packages/cyberstorm/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "resolveJsonModule": true,
+    "forceConsistentCasingInFileNames": true,
+    "composite": true,
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "jsx": "react-jsx"
+  },
+  "include": ["./src/**/*.tsx", "./src/**/*.ts"],
+  "exclude": ["node_modules"],
+  "references": [{ "path": "../hooks" }]
```
anttimaki, line 29: The hooks aren't currently used. It would be nice to keep these loosely coupled for now.

apps/cyberstorm-storybook/.storybook/preview.js
```diff
@@ -0,0 +1,26 @@
+import { LinkingProvider } from "@thunderstore/components";
+import { LinkLibrary } from "../LinkLibrary";
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  backgrounds: {
+    default: "thunderstore",
+    values: [
+      { name: "thunderstore", value: "#1E1E1E" },
```
anttimaki, line 9: At least on my screen this makes many of the button styles difficult to see. Should we consider some more neutral background color?

apps/cyberstorm-storybook/public/css/globals.css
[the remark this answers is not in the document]
gkurck: Ideally each of the button's children would be wrapped in something and have their own classes to easily hook on if needed and to avoid having to resort to chaining and nesting of classes. But we would need to establish guidelines on how to keep it consistent.

In any case, yes, the classes should reside in the component itself.

apps/cyberstorm-storybook/.storybook/preview.js
VilppeRiskidev: This color is the one Figma uses. Feel free to suggest any others.

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
VilppeRiskidev: yes👍

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,97 @@
...
+  color: var(--text-color);
+}
+
+.button__default {
+  --text-color: var(--color-white);
+  --bg-color: var(--color-surface--3);
+}
+.button__default:hover {
+  --bg-color: var(--color-surface--4);
+  outline: 2px solid var(--color-border-2);
```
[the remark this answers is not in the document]
VilppeRiskidev: Yes👍Now using borders only.

apps/cyberstorm-storybook/public/css/globals.css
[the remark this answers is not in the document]
VilppeRiskidev: label moved to the button module, unused padding thingy removed, pixels swithced to rem.

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -0,0 +1,143 @@
...
+  font-family: 'Sora', sans-serif;
+  font-style: normal;
+  font-weight: 600;
+  font-size: 16px;
+  line-height: 18px;
+}
+
+.padding-1 {
```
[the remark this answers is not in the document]
VilppeRiskidev: not used, removed

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: A bit redundant yeah. Moved things around, is it ok now?

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: Why is this? 🤔

apps/cyberstorm-storybook/public/css/globals.css
VilppeRiskidev: Naming now uses abbreviations.

apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable prettier/prettier */
+import {
+  Link as ChakraLink,
```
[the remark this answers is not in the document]
VilppeRiskidev: removed reference

apps/cyberstorm-storybook/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "babel-plugin-macros": "^3.1.0",
+    "babel-plugin-polyfill-corejs3": "^0.4.0",
+    "chromatic": "^6.4.3",
+    "core-js": "^3.19.1"
+  },
+  "dependencies": {
+    "@babel/core": "^7.16.0",
+    "@babel/preset-env": "^7.16.11",
+    "@babel/preset-typescript": "^7.16.0",
+    "@emotion/react": "^11.7.0",
```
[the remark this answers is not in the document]
VilppeRiskidev: removed emotion

packages/cyberstorm/package.json
VilppeRiskidev: removed some unused packages, did I miss some?

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: changed the type to Element. Is Element good, or should it be something else?

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: I don't like default content. The button should in theory always have some content, if it is used in a sensible way. Kind of adds unnecessary complexity imo.

packages/cyberstorm/src/components/LinkingProvider.tsx
VilppeRiskidev: oopsie, should be fixed now

packages/cyberstorm/src/components/LinkingProvider.tsx
VilppeRiskidev: oopsie, should be fixed now

packages/cyberstorm/src/components/LinkingProvider.tsx
VilppeRiskidev: oopsie, should be fixed now

packages/cyberstorm/src/components/LinkingProvider.tsx
VilppeRiskidev: Removed this reference.

packages/cyberstorm/tsconfig.json
VilppeRiskidev: the reference should be removed now. Is there something other to do?

apps/cyberstorm-storybook/public/css/reset.css
```diff
@@ -0,0 +1,175 @@
...
+    - `:where()` is used to lower specificity for easy overriding.
+*/
+
+* {
+  /* Remove default margin on everything */
+  margin: 0;
+  /* Remove default padding on everything */
+  padding: 0;
+  /* Calc `em` based line height, bigger line height for smaller font size and smaller line height for bigger font size: https://kittygiraudel.com/2020/05/18/using-calc-to-figure-out-optimal-line-height/ */
+  line-height: calc(0.25rem + 1em + 0.25rem);
```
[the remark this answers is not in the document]
VilppeRiskidev: We can always override this if needed 😈. I think this should be ok. How would you check if it's gonna work? Ofc it can be removed if it is something we do not want to "reset" here.

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: I'll look into this

apps/cyberstorm-storybook/public/css/fonts.css
```diff
@@ -0,0 +1,104 @@
+/* sora-100 - latin */
```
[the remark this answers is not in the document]
gkurck: More code -> more maintaining and hassle. I don't think we need to include support for ancient IE. I also do suspect the overall kB will go down with just the variable font. Even if not using the variable font, I'm thinking just a `.woff2` / `.tff` for each weight would also be enough as they're well supported by modern browsers.

(I mentioned `.ttf` specifically earlier since on a quick glance Google Fonts offered the variable font for Sora as `.ttf`)

My personal opinion is to go for modern approaches wherever possible for us.

apps/cyberstorm-storybook/public/css/reset.css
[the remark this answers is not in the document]
gkurck: What I meant was that this might work for us as is in the sense that it might give us sensible line heights regardless of font-size, and while not being the exact same ones as in the Figma file, they might be close and be pleasing to ones eye. If so, we can go with it, or if not, we can override it.

packages/cyberstorm/src/components/Button.tsx
gkurck: To make it easier to hook on it if needed. Something I've personally found better to overdo than not.

Makes sense also to have every child have classes than just some - and since we're creating a universal library for ourselves, would make sense to me to have the components follow the same principles so you don't have to guess or have to do the same thing differently depending on a component.

This all comes down to fact that imo we should agree on some guidelines so you don't have to think about these kind of things after they've been set. Thoughts?

packages/cyberstorm/src/components/Button.tsx
gkurck: This is a good catch, as button should always have a `type` defined. By default it should be `button` if no prop is passed, but other possible values are `submit` and `reset` which handle `form` operations, and should be possible to pass as a prop.

packages/cyberstorm/src/components/componentStyles/Button.module.css
gkurck: This is due to this rule `border-image-source: var(--color-gradient-purple-green);` present in the styles, which basically creates a border on top of the bg-image hack which shows like this on Firefox. Setting the border-image is unnecessary and should be removed. See the [Codepen](https://codepen.io/gravija/pen/poKzGYJ) for a working example.

apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -0,0 +1,31 @@
+import { LinkLibrary, ThunderstoreLinkProps } from "@thunderstore/components";
+
+interface LinkProps extends ThunderstoreLinkProps {
+  url: string;
+}
+
+const Link = (props: LinkProps): React.ReactElement => {
+  const { url, children } = props;
+
+  // Render links as spans so clicking them does nothing.
```
anttimaki, line 10: Doesn't really look like this renders a span 😄 

Anyways, I'm not sure there's a purpose to have this component anymore. The LinkLibrary definition below could return `<a>` elements directly. But feel free to keep it, if you feel it improves legibility oslt.

packages/cyberstorm/package.json
anttimaki: TBH quite impossible to remember/tell by looking at the file. But since `cyberstorm` currently contains only the button element, I'd say the following are propably not needed: `@thunderstore/hooks`, `js-cookie`, `react-dropzone`, `react-hook-form`, `react-markdown`, `react-query`, `remark-gfm`, `@types/js-cookie`.

packages/cyberstorm/src/components/Button.tsx
anttimaki: I think we might want to go with `ReactNode`.

packages/cyberstorm/src/components/Button.tsx
anttimaki: Yes, **if** it's used in a sensible way 😄 . Anyways, fine by me, it can always be changed later if we run into problems.

packages/cyberstorm/tsconfig.json
anttimaki: The need to deal with these config files arises so irregularly that I can't really say know them by heart. LGTM now 🤷🏻‍♂️

apps/cyberstorm-storybook/.storybook/preview.js
anttimaki: I think @gkurck might be a more suitable suggester.

apps/cyberstorm-storybook/LinkLibrary.tsx
VilppeRiskidev: Yeah, seems like it's not used so I removed it.

apps/cyberstorm-storybook/.storybook/preview.js
gkurck: You could use the `surface-1`, the same as the website. If that's still an issue visually, we can question the feasibility of currently assigned colors.

anttimaki: It's slightly better for the default button. `button__defaultDark` currently uses that same variable, making it invisible, but I guess it's up to developer to not use such combination. Maybe that button can be rendered with some wrapped containing a different bg color in Storybook?

apps/cyberstorm-storybook/public/css/fonts.css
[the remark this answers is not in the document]
VilppeRiskidev: now using ttf (and woff2)

apps/cyberstorm-storybook/public/css/reset.css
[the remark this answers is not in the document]
VilppeRiskidev: LGTM for now, shouldn't cause any issues. If any problems arise, it is easy to fix.

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,97 @@
...
+  border-radius: var(--size-small);
+
+  background-clip: padding-box, border-box;
+  --bg-color: var(--color-surface--3);
+  background-color: var(--bg-color);
+  --text-color: var(--color-white);
+  color: var(--text-color);
+}
+
+.button__default {
```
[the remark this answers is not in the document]
VilppeRiskidev: Wrote this down for future work

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: Fixed using the example

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: Now passing from the story

apps/cyberstorm-storybook/.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<link href="/css/reset.css" rel="stylesheet" />
```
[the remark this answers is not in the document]
VilppeRiskidev: Wrote this down for future work

packages/cyberstorm/package.json
VilppeRiskidev: Removed those also.

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: Check

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: Using the example now

apps/cyberstorm-storybook/.storybook/preview.js
VilppeRiskidev: now uses the surface-1

MythicManiac:
This should probably be squash merged since the commit history is a mess as it is right now. One massive commit is better than a messy history

apps/cyberstorm-storybook/.storybook/preview.js
```diff
@@ -0,0 +1,24 @@
+import { LinkingProvider } from "@thunderstore/components";
```
anttimaki, line 1: This approach couples the "new implementation" with the "legacy implementation" in a way that requires solving when the old ones are decommissioned. It just occurred to me, that the button stories probably don't contain any links, so we should perhaps able to ditch the `LinkingProvider` completely, at least for now. Could you test if that works? I think it would be a cleaner solution since it forces the maintainer of `cyberstorm-storybook` to solve the issue if/when the stories start to use links, instead of leaving this coupling here like a ticking time bomb.

apps/cyberstorm-storybook/LinkLibrary.tsx
anttimaki: Has something gone wrong with git here? At least "Files Changed" in GitHub still shows the outdated code comment here.

But, if my suggestion in earlier comment about ditching the `LinkinProvider` comes to pass, this file can be removed too.

apps/cyberstorm-storybook/public/css/reset.css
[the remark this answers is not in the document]
anttimaki: Well, it's easy to fix now. It might not be as easy to fix once a hundred components have left the line-height undefined and depends on this line. Just saying 🤷🏻‍♂️

apps/cyberstorm-storybook/stories/components/Button.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const meta = {
+  title: "Cyberstorm/Components/Button",
+  component: Button,
+} as ComponentMeta<typeof Button>;
+
+const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />;
+
+const DefaultButton = Template.bind({});
+DefaultButton.args = {
```
anttimaki, line 15: Could save a lot of repetition with something like:

```typescript
const defaultArgs = {
  label: "Categories",
  rightIcon: (
    <FontAwesomeIcon
      fixedWidth={true}
      icon={faChevronDown}
      className={"buttonIcon"}
    />
  ),
};

...
DefaultButton.args = {...defaultArgs, buttonStyle: "default"};
```

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { MouseEventHandler, ReactNode } from "react";
+import styles from "./componentStyles/Button.module.css";
```
anttimaki, line 2: Just redeclaring this here so it doesn't get lost in Discord: this approach doesn't work in a standalone React package like it does in Sertra which uses Next.js. We need some tooling to support importing CSS modules into TS.

While this would normally be a blocker, we can allow this now since the components aren't really used anywhere yet. But this assumes we have a task on ClickUp for this, and the work on this issue continues within some near-future milestone.

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
anttimaki: Missed the comment earlier since the discussion was marked resolved by a bot.

Just writing the line 17 as `styles;` seems to work for me. Perhaps not the correct way to fix it, but at least it removes the error in my IDE. But do write an additional comment that states why the odd line exists, so the next dev doesn't waste time trying to remove it.

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { label, leftIcon, rightIcon, buttonStyle, onClick } = props;
+  const stylesImport = styles;
+  const additionalStyle = buttonStyle
+    ? " button__" + buttonStyle
+    : " button__default";
+
+  return (
+    <button
+      type="button"
+      className={"button" + additionalStyle}
```
anttimaki, line 25: Redeclaring this here so it doesn't get lost in Discord: this approach doesn't work in a standalone React package like it does in Sertra which uses Next.js. We need some tooling to support importing CSS modules into TS. Currently this will produce a static string class name, which means the styles may leak to or be overwritten from elsewhere. We should use JS to access the correct styles from the imported `styles`, which should add a unique hash to prevent the issues.

Assuming a ticket for this exists and the work continues soonish.

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -0,0 +1,145 @@
...
+}
+
+.button__specialPurple {
+  position: relative;
+  z-index: 1;
+  background: var(--color-gradient-purple-green--darker) padding-box, var(--color-gradient-purple-green) border-box;
+  transition: all 0.12s;
+}
+
+.button__specialPurple::before {
```
anttimaki, line 120: Just a random thought here, not requiring any action now. This has a lot of repetition with `.button__specialGreen::before` which would be nice to avoid somehow, while still hiding it from the Button component (i.e. not requiring it to use modifier classes etc. for `.button__specialFoo`.

anttimaki (commented):
Left some new comments. Hopefully it's clear which require code changes and which were brought up to ensure things (tickets existing etc.) are in order outside the PR. Please reply to each comment with what has been done about the issues that don't cause code changes.

packages/cyberstorm/src/components/componentStyles/Button.module.css
gkurck: At the very least the two pseudo elements could share the rules sans `background`, as that's the only one that differs.

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Added a comment and switched to only `styles;`

apps/cyberstorm-storybook/public/css/fonts.css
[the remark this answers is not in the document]
VilppeRiskidev: Removed ttf, now only has woff2. Originally left the ttf as it was mentioned in the comment, but seems like woff2 is the way to go.

apps/cyberstorm-storybook/public/css/reset.css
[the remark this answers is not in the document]
VilppeRiskidev: Yeah, that's true. No reason to have it, I removed it now.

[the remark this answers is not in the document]
gkurck: TBF independent components should have them defined anyway to avoid the rules trickling down from elsewhere. For example, a button probably should have a line-height of 1, since the white-space is handled with padding and/or w/h declarations.

The general line-height, which could make use this formula - or not - would apply for the text content, ie. headings and body. What this particular line tries to achieve is to reduce individually declared line-heights for different sized text objects to a single, relative formula.

apps/cyberstorm-storybook/LinkLibrary.tsx
VilppeRiskidev: Removed the LinkingLibrary

apps/cyberstorm-storybook/stories/components/Button.stories.tsx
VilppeRiskidev: Very good, refactored to this.

apps/cyberstorm-storybook/.storybook/preview.js
VilppeRiskidev: Removed and still works.

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: Refactored, should be at least better now

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: https://app.clickup.com/t/20413243/TS-1230

MythicManiac:
Superseded by #568

apps/cyberstorm-storybook/.storybook/preview-head.html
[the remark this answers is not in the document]
gkurck: > So what will be the fate of tickets TS-1214 & TS-1215. They're currently marked as "in review", but neither's DoD (=being reusable) is currently fulfilled. Will they be moved over to some future milestone?

As per the discussion in Discord, these will be marked as merged, and corresponding tasks of the missing requirements have been created to be done in the next milestone.

PR #570, opened 2022-11-16
Cyberstorm MetaItem

VilppeRiskidev:


packages/cyberstorm/src/components/MetaItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+import styles from "./componentStyles/MetaItem.module.css";
+
+export interface MetaItemProps {
+  label?: string;
+  icon?: ReactNode;
+  metaItemStyle?: string;
+}
+
+/**
+ * Cyberstorm MetaItem component
```
anttimaki, line 11: "MetaItem" might not tell people that much? Maybe describe what the component looks like or where it's supposed to be used?

packages/cyberstorm/src/components/MetaItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+ */
+export const MetaItem: React.FC<MetaItemProps> = (props) => {
+  const { label, icon, metaItemStyle } = props;
+  styles; //if styles is not called, the classes from the css module aren't found
```
anttimaki, line 15: This is not required. See #569. Rebase this PR on top of that one and use the CSS modules via the imported values rather than strings as `className` values.

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
```diff
@@ -0,0 +1,39 @@
...
+  color: var(--text-color);
+  background-color: var(--bg-color);
+
+  --text-color: var(--color-default-purple--8);
+  --bg-color: transparent;
+}
+
+/* Styles for MetaItem's text */
+.metaItemLabel {
+  font-weight: 700;
```
anttimaki, line 20: Font-stuff could be onelined 🤷🏻

anttimaki (changes requested):
I regard the comment about rebasing this on top of #569 as a blocker. For the other two comments, use your own judgement.

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
```diff
@@ -0,0 +1,39 @@
...
+.metaItemIcon {
+  font-size: var(--text-xs);
+}
+
+/* Different MetaItem color schemes */
+.metaItem__default {
+  --text-color: var(--color-default-purple--8);
+}
+
+.metaItem__green {
```
gkurck, line 37: Not sure if this is a thing?

VilppeRiskidev: Used as an example, not in the designs. Added a comment in the code about this.

packages/cyberstorm/src/components/MetaItem.tsx
VilppeRiskidev: Could be anything, "InformationDisplay" "DataPoint" etc... Don't want to use too much time to think about a better name, we'll see if a better one comes up. Shouldn't be _too bad_ of a refactor to find+replace metaitem with something else.

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
VilppeRiskidev: One-lined ✅

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
```diff
@@ -0,0 +1,39 @@
+/* Common MetaItem styles */
+.metaItem {
+  display: flex;
+  flex: none;
+  flex-direction: row;
+  gap: var(--size-xs);
```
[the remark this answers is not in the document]
VilppeRiskidev: changed all of them to `em`

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
```diff
@@ -0,0 +1,39 @@
...
+  align-items: center;
+  justify-content: center;
+
+  padding: var(--size-xs) var(--size-xs);
```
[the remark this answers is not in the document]
VilppeRiskidev: changed all of them to `em`

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -22,10 +22,10 @@
 /* Styles for Button's text */
 .buttonLabel {
   font-weight: 600;
-  font-size: var(--size-m);
+  font-size: var(--text-m);
   font-family: Sora, sans-serif;
   font-style: normal;
-  line-height: 1.125rem;
+  line-height: var(--text-l);
```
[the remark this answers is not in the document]
VilppeRiskidev: changed all of them to `em`

VilppeRiskidev:
I think this PR (as well as the other "Cyberstorm UI Component" ones) can be squash merged, but whatever works.

packages/cyberstorm/src/components/MetaItem.tsx
anttimaki: Well I didn't really mean to suggest renaming it, rather to use the comment field to describe what MetaItems are used for. Not a big deal, but currently I kinda know that this is a React component named MetaItem that belongs to Cyberstorm package even without the comment, so it doesn't add that much value.

packages/cyberstorm/src/components/MetaItem.tsx
```diff
@@ -4,23 +4,23 @@ import styles from "./componentStyles/MetaItem.module.css";
...
   const { label, icon, metaItemStyle } = props;
-  styles; //if styles is not called, the classes from the css module aren't found
-  const additionalStyle = metaItemStyle
-    ? " metaItem__" + metaItemStyle
-    : " metaItem__default";
 
   return (
-    <div className={"metaItem" + additionalStyle}>
+    <div
+      /* TS is not aware of defaultProps of function components. */
```
anttimaki, line 18: There's no `defaultProps` defined for this component? And since `metaItemStyle` is marked optional parameter, we really shouldn't use `!` for non-null assertion, since the value can indeed be nullish without default prop definition. The proper fix would be to add `metaItem__default` as the default prop though.

anttimaki (changes requested):
DefaultProps should be added to the component.

packages/cyberstorm/src/components/MetaItem.tsx
VilppeRiskidev: added a defaultProp for the style

packages/cyberstorm/src/components/MetaItem.tsx
VilppeRiskidev: Now has a bit more descriptive description

anttimaki (approved):
My comments have been addressed.

PR #571, opened 2022-11-16
Add Tag component

VilppeRiskidev:


packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -0,0 +1,26 @@
...
+  rightIcon?: ReactNode;
+  tagStyle?: string;
+}
+
+/**
+ * Cyberstorm Tag component
+ */
+export const Tag: React.FC<ButtonProps> = (props) => {
+  const { label, tagStyle, leftIcon, rightIcon } = props;
+  styles; //if styles is not called, the classes from the css module aren't found
```
anttimaki, line 16: This is not required. See https://github.com/thunderstore-io/thunderstore-ui/pull/569. Rebase this PR on top of that one and use the CSS modules via the imported values rather than strings as className values.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -0,0 +1,38 @@
...
+  --bg-color: var(--color-surface--3);
+}
+
+.tag:hover {
+  --bg-color: var(--color-surface--4);
+}
+
+/* Styles for Tag's text */
+.tagLabel {
+  font-weight: 700;
```
anttimaki, line 25: Font-stuff could be onelined 🤷🏻

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -0,0 +1,38 @@
...
+  font-size: var(--text-xxs);
+  font-family: Inter, sans-serif;
+  font-style: normal;
+  line-height: var(--text-xs);
+
+  text-transform: uppercase;
+}
+
+/* Styles for Tag's Icons */
+.tagIcon {
```
anttimaki, line 35: What if we want a bigger tag with icons? Should the size be a separate property that would affect size, padding, font-size etc? I guess @gkurck would be a better authority on this than me.

anttimaki (commented):
Have you followed the [precommit installation guide](https://github.com/thunderstore-io/thunderstore-ui#pre-commit)? It shouldn't be something that you need to run manually.

**EDIT: I Posted this as Comment but it should be Request Changes**

packages/cyberstorm/src/components/componentStyles/Tag.module.css
gkurck: Sure, wouldn't hurt having the sizes recurring throughout the designs as a prop. One can also set specific edge case stylings for a component in the parent component they are used in if needed. That's one of the main reasons I'm advocating giving each of the component's element a class to hook on.

Regarding this one, I would suggest creating the base component with the same size as it's in the component library, not with the smaller size used in the mod card.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -0,0 +1,38 @@
...
+  --bg-color: var(--color-surface--4);
+}
+
+/* Styles for Tag's text */
+.tagLabel {
+  font-weight: 700;
+  font-size: var(--text-xxs);
+  font-family: Inter, sans-serif;
+  font-style: normal;
+  line-height: var(--text-xs);
```
gkurck, line 29: Using `rem` as a unit for `line-height` is problematic, since it's value gets computed from the root font-size, not from where it's declared in.

It's preferred to use unitless values if you need to set it to something specific, or just let it inherit the default value, `normal`, which translates to `1.2`. The way unitless values work with `line-height` is that they multiply the element's font-size with itself.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -0,0 +1,38 @@
+/* Common Tag styles */
+.tag {
+  display: flex;
+  flex: none;
+  flex-direction: row;
+  gap: var(--size-tiny);
+  align-items: center;
+  justify-content: center;
+
+  padding: var(--size-tiny) var(--size-xs);
```
gkurck, line 10: Might be better to use `em` instead of `rem` to change in relation to changes in `font-size`.

In tag's case, using `padding: .5em 1em;` would most likely yield preferred results.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: One-lined ✅

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: Refactored to match the design and use em

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: Refactored to match the design and use em

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: Refactored to match the design and use em

packages/cyberstorm/src/components/Tag.tsx
VilppeRiskidev: Done

packages/cyberstorm/src/components/componentStyles/Tag.module.css
anttimaki: Well there's still no size property on the component, e.g. `<Tag size="huge" label="Hello" />` but I'm not sure if it's required to fill the DoD? Having a "Refs TS-1234" in the commit messages would be nice to track the changes to the ticket. Couldn't find it using the ticket system's search either.

anttimaki (approved):
Besides the possible size prop, things are in order from my perspective. But we should probably wait for another review from gkurck.

packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -6,19 +6,20 @@ export interface TagProps {
   leftIcon?: ReactNode;
   rightIcon?: ReactNode;
   tagStyle?: "tag__default";
+  tagSize?: "tag__small" | "tag__medium" | "tag__large";
```
anttimaki, line 9: I think the name of the prop should be just `size`, since what else could it refer to than the component itself? In comparison, I assumed `tagStyle` has the tag-prefix to avoid confusing it with HTML's `style` attribute.

Secondly, I feel that having the `tag__` prefix on the prop value is unnecessarily verbose. Something like `size="small"` is what I'd expect, if I was a developer using such a component. So maybe the prop value should be decoupled from the class name, letting the component handle forming the class name internally?

I guess both of these tie into the naming convention thing, and perhaps they aren't blockers for this PR necessarily. But we should have that discussion (again) sometime soon? Could you ensure that the things mentioned here are included in the related ticket?

PR #572, opened 2022-11-24
Add Breadcrumb component

Oksamies:
refs TS-1316


packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+    if (n1.level < n2.level) {
+      return -1;
+    }
+    return 0;
+  });
+  return sortedArray.map((crumb, index) => {
+    if (index === 0) {
+      return (
+        <a key={crumb.level} href={crumb.path}>
```
anttimaki, line 31: Regardless of whether we end up using the linking provider, we probably don't want to use `<a>` here since the main target is Next.js which has its own link component. Using regular links there would cause a full page refresh instead of SPA page transition.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
...
+          {crumb.label}
+        </a>
+      );
+    }
+    return (
+      <a key={crumb.level} href={crumb.path}>
+        <span aria-hidden="true">/</span> {crumb.label}
```
anttimaki, line 38: The divider (`<span aria-hidden="true">/</span>`) should not be inside the link but between two links, and unclickable.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return sortedArray.map((crumb, index) => {
+    if (index === 0) {
+      return (
+        <a key={crumb.level} href={crumb.path}>
+          {crumb.label}
+        </a>
+      );
+    }
+    return (
+      <a key={crumb.level} href={crumb.path}>
```
anttimaki, line 37: I think usually the last part of the breadcrumb, i.e. the current page, is not rendered as a link but a span oslt. We might want to consider it here, too.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </a>
+    );
+  });
+}
+
+export const Breadcrumb: React.FC<BreadcrumbProps> = (props) => {
+  const { crumbs, breadcrumbStyle } = props;
+
+  return (
+    <div
```
anttimaki, line 48: Ditching the `breadcrumbStyle` prop cleans up this quite nicely.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
+/* Common Breadcrumb styles */
```
anttimaki, line 1: Yes this is `Breadcrumb.module.css`. I expect to find Breadcrumb styles in this file. I'd yeet the comment.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
...
+.breadcrumb {
+  display: flex;
+  flex-direction: row;
+  gap: var(--size-s);
+  align-items: center;
+  font-size: var(--size-m);
```
anttimaki, line 7: I assume this should use `--text-{something}`.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { MouseEventHandler, ReactNode } from "react";
+import styles from "./componentStyles/Breadcrumb.module.css";
+
+export interface CrumbProps {
+  level: number;
+  path?: string;
+  label: string | ReactNode;
```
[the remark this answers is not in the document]
anttimaki: Somewhat besides the point, but this also allows placing other links, buttons etc. inside the anchor element, which isn't kosher.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
...
+  font-family: Inter, sans-serif;
+}
+
+.breadcrumb > a {
+  display: flex;
```
anttimaki, line 12: Force-flexing the content of the <a> (which allows any ReactNode) seems kinda risky. It may work for us, but pretty much makes it pointless to type the `label` prop as ReactNode.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
...
+  flex-direction: row;
+  gap: var(--size-s);
+  align-items: center;
+  color: var(--color-cyber-green--50);
+  font-weight: 400;
+
+  text-decoration: none;
```
anttimaki, line 19: Should we add underline when the link is hovered?

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.breadcrumb > a > svg {
```
anttimaki, line 22: Also this makes bold assumptions about the content of the `label` prop.

packages/cyberstorm/src/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./components/Button";
+export * from "./components/Breadcrumb";
```
anttimaki, line 2: Please sort these alphabetically.

anttimaki (changes requested):
Back to the drawing board with this one, I'm afraid. Main point being the incompatibility with both LinkingProvider pattern and Next.js's link components. These two kinda render the other comments moot, but I recommended checking them out anyway for future reference.

Sidenote: I don't think the second commit should be part of this PR?

Oksamies:
Alright I've refactor this to work with the NextJS links.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+      ))}
+
+      <LeafCrumb crumb={leaf} />
+    </div>
+  );
+};
+
+const Home: React.FC = () => {
+  return (
+    <LinkingProvider value={LinkLibrary}>
```
anttimaki, line 51: LinkLibrary shouldn't be part of `@thunderstore/cyberstorm` (more about this in a bit) so we can't really hard code the link here. And I'm not sure we even want to hard code it to point to index? Maybe we want to give the apps a possibility to point the root node to e.g. community listing instead of landing page with "download the MM" hype?

I like the idea of not having to pass the home node everywhere this is going to be used, but it just doesn't work that well in this particular case.

anttimaki, line 51: Second issue: LinkingProvider should be used on the apps using the components, not in the component library. Ideally it's included just once in the top of the app ([[1]](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/nextjs/pages/_app.tsx#L31), [[2]](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/storybook/.storybook/preview.js#L27)), instead of wrapping each and every link.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <LinkLibrary.Index>
+        <FontAwesomeIcon
+          fixedWidth={true}
+          icon={faHouse}
+          className={`${styles.breadcrumbSeparator}`}
```
anttimaki, line 56: I don't think the name "breadcrumbSeparator" is quite fitting since this isn't the separator?

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+  parts: Crumb[];
+}
+
+/**
+ * Navigational bread crumbs defining where the user is at the site.
+ */
+export const BreadCrumb: React.FC<Props> = (props) => {
+  if (!props.parts) {
+    return (
+      <div className={`${styles.breadcrumb}`}>
```
anttimaki, line 24: There's no need for template string here since it's just one class. I think `className={styles.breadcrumb}` should suffice. This is repeated multiple times below.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+    </div>
+  );
+};
+
+const Home: React.FC = () => {
+  return (
+    <LinkingProvider value={LinkLibrary}>
+      <LinkLibrary.Index>
+        <FontAwesomeIcon
+          fixedWidth={true}
```
anttimaki, line 54: It was agreed on Monday's meeting that hard coded boolean props should use short form of `<FontAwesomeIcon fixedWidth ...>`. This is repeated multiple times in the file.

Check out the meeting notes if you half-assed the meeting.

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
+import {
```
anttimaki, line 1: LinkLibrary should be defined by the app that uses these components, since they are the ones who know how the links should be rendered. Half of the point of complicated LinkingProvider is to allow writing components that don't know how the links should be rendered in the end. (The other half is to allow e.g. community link to point to different URLs in different apps.)

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+  LinkLibrary,
+  thunderstoreLinkProps,
+  ThunderstoreLinkProps,
+} from "@thunderstore/components";
+import NextLink from "next/link";
```
anttimaki, line 6: As mentioned by Mythic in Discord, `@thunderstore/cyberstorm` should be agnostic about the framework using the components. And also note that `@thunderstore/cyberstorm-storybook` doesn't use Next.js either.

And you didn't add the dependency to packages.json either.

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface CyberstormLinkProps {
+  children?: React.ReactNode;
+}
+
+interface LinkProps extends CyberstormLinkProps, ThunderstoreLinkProps {
+  queryParams?: string;
+  url: string;
+}
+
+const CyberstormLink = forwardRef<HTMLAnchorElement>(
```
anttimaki, line 18: Given that this component does pretty much nothing, I see no reason to have it. The Link component below could as well include the `<a>` element directly.

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+  (props: CyberstormLinkProps, ref) => {
+    const { ...rest } = props;
```
anttimaki, line 20: Why do you think the place where this was copied from uses the `...rest` operator? What does it do here?

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+CyberstormLink.displayName = "CyberstormLink";
+
+const Link = (props: LinkProps): React.ReactElement => {
+  const { children, queryParams, url, ...cyberStormProps } = props;
+  const q = queryParams ? `?${queryParams}` : "";
+
+  for (const key in thunderstoreLinkProps) {
+    delete cyberStormProps[key as keyof ThunderstoreLinkProps];
+  }
+
```
anttimaki, line 38: What do you think `cyberstormProps` variable contains at this row?

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <NextLink href={`${url}${q}`} passHref>
```
anttimaki, line 40: To reiterate earlier comments, neither `@thunderstore/cyberstorm` nor `@thunderstore/cyberstorm-storybook` uses Next.js or its link component.

packages/cyberstorm/src/components/LinkLibrary.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <CyberstormLink {...cyberStormProps}>{children}</CyberstormLink>
```
anttimaki, line 41: What are the props `CyberstormLink` accepts? What are the props that are possibly passed in here?

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
+.breadcrumb {
```
anttimaki, line 1: It was agreed on Monday's meeting that this should be called `.root`.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+  flex-direction: row;
+  gap: var(--size-s);
+  align-items: center;
+  font-size: var(--size-m);
+  font-family: Inter, sans-serif;
+}
+
+.breadcrumbLabel {
```
anttimaki, line 10: IIRC it was also decided that we shouldn't use the repetitive prefixes, so this should be just `.label`.

anttimaki (changes requested):
IMO this doesn't reflect what we discussed in the voice chat. Also some of my previous comments weren't addressed (I resolved the ones that were outdated or irrelevant due to changes).

Oksamies:
Good review, there clearly was a lot that I missed and/or forgot about. TBH 1/3 of the code kind of falls under a gray area for me, where the logic of React and/or TS isn't that clear. I'll try my best to fix and understand all of the commented issues, but I do have to ask for patience with my output. Hopefully someday it wont resemble a Nurgling so much.

anttimaki:
> Good review, there clearly was a lot that I missed and/or forgot about. TBH 1/3 of the code kind of falls under a gray area for me, where the logic of React and/or TS isn't that clear. I'll try my best to fix and understand all of the commented issues, but I do have to ask for patience with my output. Hopefully someday it wont resemble a Nurgling so much.

image

apps/cyberstorm-storybook/stories/components/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { BreadCrumb } from "@thunderstore/cyberstorm";
+import { CommunityLink, CommunityPackagesLink } from "@thunderstore/components";
```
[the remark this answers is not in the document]
Oksamies: I actually had those in the cyberstorm components, but I removed them while "cleaning up". Dunno why, I was probably tired and didn't remember that was what we wanted 😅

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Navigational bread crumbs defining where the user is at the site.
+ */
+export const BreadCrumb: React.FC<Props> = (props) => {
+  if (!props.parts) {
+    return (
+      <div className={`${styles.breadcrumb}`}>
+        <Home />
```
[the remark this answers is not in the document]
Oksamies: Nope, fixin it 👍

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: I've edited this to work so that, if no parts are passed, it will include `DefaultHomeCrumb` part in the breadcrumb. So the "home" isn't forced, but it is assured that there is something always in the breadcrumb.

Oksamies: Changed it so that `BreadCrumb` needs a linkLibrary as a prop. Is that the correct way?

Oksamies: Oh no hold on I'm not sure anymore. Do I need to pass the linkLibrary to the function or can I fetch somehow with the `useContext`?

packages/cyberstorm/src/components/LinkLibrary.tsx
Oksamies: Alright, I've now done it like so now that the LinkLibrary is defined outside the cyberstorm package. For example now that it's used by the storybook, it's defined in the stories file.

packages/cyberstorm/src/components/LinkLibrary.tsx
Oksamies: This LinkLibrary file is now removed, since it is not what we want to have. All of the stuff imported from this file are now imported from the correct place. The new LinkingProvider file under cyberstorm.

packages/cyberstorm/src/components/LinkLibrary.tsx
Oksamies: If I understood correctly these 3 rows delete the "unwanted" prop types from the `cybeStormProps` const 🤔

Let me know if I'm correct, but as with the previous review comments, files now deleted 👍

packages/cyberstorm/src/components/LinkLibrary.tsx
Oksamies: None no more 🥇

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
Oksamies: Hmm, I'm not 100% sure if I understood the problem here correctly, because I can't explain my understanding or answer with a fix suggestion.

Let me know if the fix wasn't correct or if it might seem like I've missed the point.

I removed the force-flexing from the `.label`, since it doesn't need that. I also removed the forced `flex-direction: row` from `.root`.

Should the the `flex` stuff be defined in the stylings of `LinkLibrary`s `Link` function? For example in the `storybook-cyberstorm`s `LinkLibrary` file?

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
Oksamies: Wasn't in the designs IIRC. @gkurck What do you think?

MythicManiac (changes requested):
Big one: don't pass the link library around as a prop to the components, fetch it from the context when it's needed. On that note, I'm not entirely sure if you even need to refer to things through the link library directly, but I didn't review that far yet.

packages/cyberstorm/src/components/Links.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const AnonymousLink = wrap("Anonymous");
+export const CommunitiesLink = wrap("Communities");
+export const CommunityLink = wrap("Community");
+export const CommunityPackagesLink = wrap("CommunityPackages");
+export const IndexLink = wrap("Index");
+export const PackageLink = wrap("Package");
+export const PackageDependantsLink = wrap("PackageDependants");
+export const PackageUploadLink = wrap("PackageUpload");
+export const TeamLink = wrap("Team");
```
MythicManiac, line 27: Adding to my previous review, you can import & use these components directly rather than going through the link library.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
gkurck: The designs do have underline on hover though.

Tbf might be good idea to map out the default visual behaviour of anchors with @VilppeRiskidev to see if a default styles for anchors could be added which components like these inherit.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
anttimaki: I guess that's a poorly worded comment since I'm not 100% percent certain what I meant without seeing the whole old context, which is now gone. I think my point was the typing allowed the breadcrumb to accept any React node as a prop and show it. Which means one could pass e.g. a `a` that contains a `div` which contains a `h3`, `p`, and an icon, with their own CSS module defining the styles that should be used to render that component. But since you used `flex` there, it meant that component's styles would perhaps be flexed, even if whoever created that component didn't want them to. (Today I'm not 100% how this would actually work with the CSS module scoping.)

But the very point seems moot with the current version of the code. Feel free to mark this conversation resolved once you've read this.

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: To me it seems L25-27 returns null if no parts are passed, and `DefaultHomeCrumb` is never used anywhere (which you're IDE would probably warn about, except you've exported it for some reason, which I think you shouldn't do either).

anttimaki: Methinks you should do just like we do in the existing example we have:

1. The [app defines a `LinkLibrary`](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/storybook/LinkLibrary.tsx), telling how the link should be rendered in that app
2. Somewhere [high in the app structure, the app uses `LinkingProvider`](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/storybook/.storybook/preview.js#L27), passing the `LinkLibrary` to it as a prop. All the wrapped components can now consume the context of that provider.
3. In the app, inside that context, a [component defines which links should be used](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/storybook/stories/components/BreadCrumbs.stories.tsx#L26-L28) by the breadcrumbs component. The ["links" are in fact components themselves](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/storybook/stories/components/BreadCrumbs.stories.tsx#L4-L5), defined in the component library
4. The [breadcrumbs component merely renders the link components](https://github.com/thunderstore-io/thunderstore-ui/blob/master/packages/components/src/components/BreadCrumbs.tsx#L63-L65) passed to it. It doesn't know what's the output when it renders them, and much like a honeybadger, it doesn't care. As long as the rendering happens within the context of `LinkingProvider`, the output will be what was defined in the `LinkLibrary`

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: I don't think renaming it to "separator" changes much since it's still not a separator, it's a home icon. Maybe have two class names that use the same style definitions in the CSS file?

packages/cyberstorm/src/components/LinkLibrary.tsx
anttimaki: Well the answer I was looking for is that using the `...rest` in [destructing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) like this means you create a `const rest` that is an exact copy of props. Which is kinda pointless, no?

packages/cyberstorm/src/components/LinkLibrary.tsx
anttimaki: IIRC the issue was that you passed like one prop so this did nothing.

apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -0,0 +1,40 @@
+import {
+  LinkLibrary,
+  ThunderstoreLinkProps,
+} from "@thunderstore/cyberstorm/src/components/LinkingProvider";
```
anttimaki, line 4: These should be exported directly from `index.ts`. Whoever uses the components don't need to know about the internal paths of cyberstorm package.

apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -0,0 +1,40 @@
...
+import React from "react";
+
+interface CyberstormLinkProps {
+  children?: React.ReactNode;
+}
+
+interface LinkProps extends CyberstormLinkProps, ThunderstoreLinkProps {
```
anttimaki, line 11: As far as I can tell, we do nothing with ThunderstoreLinkProps here. If that's so, it should perhaps be yeeted.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,90 @@
...
+  LinkComponent: IndexLink,
+  label: (
+    <FontAwesomeIcon fixedWidth icon={faHouse} className={styles.separator} />
+  ),
+};
+
+const MidCrumb: React.FC<{ linkLibrary: LinkLibrary; crumb: Crumb }> = (
+  props
+) => {
+  const { linkLibrary } = props;
```
anttimaki, line 55: Why is this and the next line two separate lines?

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,90 @@
...
+      ))}
+
+      <LeafCrumb crumb={leaf} />
+    </div>
+  );
+};
+
+export const DefaultHomeCrumb: Crumb = {
+  LinkComponent: IndexLink,
+  label: (
```
anttimaki, line 47: Are you sure this is correct? At least my IDE complains that the typings are incorrect.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <FontAwesomeIcon fixedWidth icon={faHouse} className={styles.separator} />
```
anttimaki, line 48: Also the typings for `icon` prop seems to be incorrect, here as well as elsewhere.

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
+/**
+ * Implementation-agnostic context provider for navigation.
+ *
+ * @thunderstore/components aims to be reusable. Therefore the links
```
anttimaki, line 4: Copy-pasted package name.

packages/cyberstorm/src/components/LinkingProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+// required since the stripping is done during runtime.
+export const thunderstoreLinkProps: ThunderstoreLinkProps = {
+  community: "",
+  namespace: "",
+  package: "",
+  team: "",
+};
+
+// Accepting any and all props is required to keep the linking
+// implementation-agnostic. E.g. an implementation with Chakra has all
```
anttimaki, line 42: I don't think Chakra should be mentioned here anymore.

anttimaki (changes requested):
Didn't point out all the places where the linking stuff is still wrong. Following [this comment](https://github.com/thunderstore-io/thunderstore-ui/pull/572#discussion_r1046764360) should resolve them.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
Oksamies: Yeah, alright I'm not going to add the underline hover here. I'll communicate with Vili about this 👍

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: DefaultHomeCrumb is used within the .stories file to add it as the first part in the different templates. I didn't want to use it in the Breadcrumb component by default, cause we might not want it to be the first crumb. (IIRC from a previous conversation somewhere)

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Ah yeah true, fixed 👍

packages/cyberstorm/src/components/LinkLibrary.tsx
Oksamies: Yea correct, in that line 20 (which is now outdated) it would have been pointless. Good that you pointed out that destructing assignment, I wasn't fully aware how it worked 👍

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: > DefaultHomeCrumb is used within the .stories file

No, it's not.

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Probably didn't realize I could do it in the same line. Removed the second one, since it was no longer needed.

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: It wasn't. I looked at it with @MythicManiac and we did quite a bit regarding the typing. There's still issues in a different place, but it's getting there. https://discord.com/channels/877914952902516776/1051881420173344868/1052272328236601425

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Seems like I needed to typecast them, not sure why those were incorrect in the first place, but should be fixed now 👍

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
Oksamies: Ah alright, now I see what you meant. I'm not also that sure if it would have been a problem, cause how I've understood the flex inheritance it would have not been passed down. In any case the "flex forcing" there was not needed anyways, so it's good that it was removed.

Marking as resolved 👍

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: I think I've implemented it correctly and according to your guide. (which was a good one and maybe something we might want to put into the LinkingProvider readme)

Though I do have to note that I did not add this into any "app", since we don't have an app yet.

I've added the usage of `LinkingProvider` into the `Breadcrumb.stories.tsx`, as a decorator for the story.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <LeafCrumb crumb={leaf} />
+    </div>
+  );
+}
+
+function MidCrumb<P>(props: { crumb: Crumb<P> }) {
+  const { crumb } = props;
+  const label = CrumbLabel(crumb.label);
+  return (
+    <crumb.LinkComponent {...crumb.LinkProps}>{label}</crumb.LinkComponent>
```
anttimaki, line 51: In discord, you had asked about two issues on this line.

The first one, regarding `crumb.LinkComponent`: L14, defines that `LinkComponent` can be `undefined`, but here it's not checked whether it actually is `undefined`, instead it's treated as a function component. To fix this perhaps add a if-check that just returns `label` if the component is `undefined`? (Although IDK if it makes sense for the `MidCrumb` to consist only of label? Might be some use cases for it 🤷🏻)

The second one, regarding `...crumb.LinkProps`: Currently `MidCrumb` doesn't know whether the `crumb` it received is the one defined on L12-15 or the one on L16-L20. I.e. the `crumb` might not have `LinkProps` property at all. How to solve this seems a bit more complex but can probable be done using a type guard function. But before I spend more time on that, I need to know what was the goal of typing the crumbs like they've typed here (vs. as they've been typed on the old implementation)?

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: See https://github.com/thunderstore-io/thunderstore-ui/pull/572#discussion_r1048151403 (and resolve this conversation)

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: See https://github.com/thunderstore-io/thunderstore-ui/pull/572#discussion_r1048117092 (and resolve this conversation).

Also FYI @VilppeRiskidev & @Cyanideankst: in the future ensure fontawesome library versions are in sync.

packages/cyberstorm/src/components/componentStyles/Breadcrumb.module.css
```diff
@@ -0,0 +1,21 @@
+.root {
+  display: flex;
+  gap: var(--size-s);
+  align-items: center;
+  font-size: var(--text-m);
+  font-family: Inter, sans-serif;
```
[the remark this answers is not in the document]
Oksamies: After rebasing the variable should be available, good point out 👍

apps/cyberstorm-storybook/stories/components/Breadcrumb.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  DefaultHomeCrumb,
+  LinkingProvider,
+} from "@thunderstore/cyberstorm";
+import { LinkLibrary } from "@thunderstore/cyberstorm-storybook/LinkLibrary";
+
+const meta = {
+  title: "Cyberstorm/Components/Breadcrumb",
+  component: BreadCrumb,
+  decorators: [
+    (Story) => (
```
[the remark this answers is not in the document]
Oksamies: Had few words about this with @MythicManiac and got to the conclusion, that we might want to use different LinkLibraries in the stories, so might be better to just specify them in the per story decorators.

Although I prefer the per story decorator, IIRC we can do the "global decorator" in the storybooks `preview.js`.

I'll add a TBD discussion note to the milestone, so we can talk about it in the post-mortem.

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Goal was to have more precise types for the `parts`, so that we can get rid of the `// eslint-disable-line @typescript-eslint/no-explicit-any`.

In the latest commits I've reworked the types a bit, so that there are at least no errors and everything works as intended. Though I'm not sure if the code is that "clean".

Tyvm for explaining the errors, idk why it is so hard for me to extract the actual problem out of those.

apps/cyberstorm-storybook/stories/components/Breadcrumb.stories.tsx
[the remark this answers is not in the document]
MythicManiac: To be clear, the conclusion was that in the context of this PR I don't see it as a problem if it's specific to this story, but long-term it probably makes sense to have it as a global context. You can override it per-story in any case if you need to

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+  if (crumb === undefined) {
+    return null;
+  } else {
+    return CrumbLabel(crumb.label);
+  }
+}
+
+function CrumbLabel(label: Label) {
+  if (label && typeof label === "string") {
+    return <p className={styles.label}>{label}</p>;
```
MythicManiac, line 72: @gkurck this should probably be a span right?

gkurck: The last one without an anchor should be a span, yes. Those inside anchors probably don't need to be additionally wrapped in anything other than the anchor.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { IndexLink } from "./Links";
+
+type Label = ReactNode | string;
+
+interface Crumb {
+  label: Label;
+  LinkComponent?: FunctionComponent;
+}
+
+type LinkCrumb<P> = Omit<Crumb, "LinkComponent"> & {
```
anttimaki, line 15: What is the purpose of `Omit<Crumb, "LinkComponent"> &` here? AFAIK it just picks `label` from `Crumb`, but then redefines it anyway. IMO should be cleared away but not a blocker.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+  label: Label;
+  LinkComponent: FunctionComponent<P>;
+  LinkProps: P;
+};
+
+interface BreadCrumbProps<P> {
+  parts: (Crumb | LinkCrumb<P>)[];
```
anttimaki, line 22: I don't think this is what we want. It might work on the storybook stories, and maybe even on some pages, but sooner or later we will have a page that will not work with this.

The problem is that this assumes all passed `LinkCrumbs` share the same generic type `<P>`. The example below would not work, since the two links use different props that are incompatible with each other:

```typescript
export const MyCrumbs = () => {
  const crumbs = [
    {
      LinkComponent: TeamLink,
      LinkProps: { team: "Testruction" },
      label: "Testruction",
    },
    {
      LinkComponent: CommunityLink,
      LinkProps: { community: "ror2" },
      label: "Risk of Rain 2",
    },
  ];
  return <BreadCrumb parts={crumbs} />;
};
```

I consider this a blocker.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+  );
+}
+
+function MidCrumb<P>(props: { crumb: Crumb | LinkCrumb<P> }) {
+  const { crumb } = props;
+  if (crumb.LinkComponent === undefined) {
+    return CrumbLabel(crumb.label);
+  }
+  if (!("LinkProps" in crumb)) {
```
anttimaki, line 51: I guess this works here, but for future reference checkout [user-defined type guards](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). I prefer them since they're reusable, more readable and usually make the intent more clear.

Not a blocker.

anttimaki (changes requested):
Since there once blocker I'd strongly consider fixing the non-blockers at the same time.

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: > Though I do have to note that I did not add this into any "app", since we don't have an app yet.

I would consider `cyberstorm-storybook`, which resides under the `apps` dir, an "app" 😉

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Oh, I had left the label there by accident. Idea was to take the base from Crumb and then just redefine `LinkComponent` and add the `LinkProps`.

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: I've implemented something like what was in the example of "user-defined type guards".

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: Latest commit has a suggestion on how this could be solved(?)

MythicManiac: As mentioned in Discord, I consider typing this (and other component implementations) properly a requirement for them being merged. In case it's not possible to declare the types in a supported way it's of course understandable if they aren't, but as far as I know, it should be doable here. I'd suggest separating the link library implementation to a standalone PR and basing your other (and this) work on that branch to avoid the type declarations of the crumb component from blocking said other work.

packages/cyberstorm/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return <>{label}</>;
+}
+
+const Separator = () => (
+  <span aria-hidden="true" className={styles.separator}>
+    <FontAwesomeIcon fixedWidth icon={faSlashForward} />
+  </span>
+);
+
+export const DefaultHomeCrumb: Crumb = {
```
anttimaki, line 76: To think of it, `BreadCrumb` could take a boolean prop that defaults to true, and when that prop is true, the `DefaultHomeCrumb` is shown as the first crumb. Might be handier than importing this and then adding it to an array.

Just an afterthought, not a blocker.

packages/cyberstorm/src/components/Breadcrumb.tsx
anttimaki: I consider this comment resolved, but I'll leave it open so I don't hide Mythic's comment above.

anttimaki (approved):
I consider this PR approved as it is, but don't merge it without discussing with Mythic what to do about it.

packages/cyberstorm/src/components/Breadcrumb.tsx
Oksamies: This is fixd within Mythics PR https://github.com/thunderstore-io/thunderstore-ui/pull/592/files#diff-c283c12dd934359b08faac8e740784282f8f734db11b2b11529bf1987c0a10d5R13

PR #573, opened 2022-11-25
Add Tooltip component and radix-ui react-tooltip

Oksamies:
As the title says

refs TS-1315


packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,47 @@
...
+  // const { label, tooltipStyle } = props;
+  return (
+    <RadixTooltip.Provider delayDuration={150}>
+      <RadixTooltip.Root>
+        <RadixTooltip.Trigger>{children}</RadixTooltip.Trigger>
+        <RadixTooltip.Portal>
+          <RadixTooltip.Content
+            /* TS is not aware of defaultProps of function components. */
+            /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+            className={`${styles.tooltip} ${styles[tooltipStyle!]}`}
```
MythicManiac, line 26: As discussed in Discord, this and other similar usages have to be changed. A side benefit is that you can get rid of the eslint disabling comments.

packages/cyberstorm/src/components/componentStyles/Tooltip.module.css
```diff
@@ -0,0 +1,92 @@
...
+button {
+  all: unset;
+}
+
+.tooltip {
+  padding: 0.5rem 0.75rem;
+  border-radius: 0.5rem;
+  color: var(--color-white);
+
+  font-weight: 700;
```
VilppeRiskidev, line 11: I think the whole file needs refactoring when the global.css changes drop from the PackageCard PR. Use variables and better naming and stuff.

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Tooltip {...args}>
+        <Button {...buttonargs} />
+      </Tooltip>
+    </div>
+  </div>
+);
+
+const DefaultTooltip = Template.bind({});
+DefaultTooltip.args = defaultArgs;
+
```
VilppeRiskidev, line 62: Could probably use another exmple of what happens when basicly no props are passed. as in empty parameters.
The pattern is:
MinimalTooltip (minimal to no props passed)
ReferenceTooltip (matching the design)
ExtremeTooltip ("stupid" extreme case, for example 500 chars long label etc.)

VilppeRiskidev: What I meant by Minimal was to (in this case) leave the props empty, so it acts as an example of what happens when no label is passed. Since the label is an optional prop.

VilppeRiskidev: Also the long one could be even longer, like 200 chars or something :D

packages/cyberstorm/src/components/componentStyles/Tooltip.module.css
```diff
@@ -0,0 +1,84 @@
+.tooltip {
+  padding: 0.5rem 0.75rem;
+  border-radius: 0.5rem;
+  color: var(--color-white);
+
+  font-weight: var(--font-weight-bold);
+  font-size: var(--size-m);
+  background-color: var(--color-default-purple--4);
```
VilppeRiskidev, line 8: I think this one should use the surface variable `--color-surface--3`

VilppeRiskidev (changes requested):
Mainly needs more complying to the naming conventions https://github.com/thunderstore-io/thunderstore-ui/blob/master/packages/cyberstorm/README.md

packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Cyberstorm Tooltip component
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = ({
+  label,
+  children,
+}) => {
+  return (
+    <RadixTooltip.Provider delayDuration={150}>
```
MythicManiac, line 16: Is this provider something that has to be included for every tooltip individually, or should it be included once globally?

packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { PropsWithChildren } from "react";
+import * as RadixTooltip from "@radix-ui/react-tooltip";
+import styles from "./componentStyles/Tooltip.module.css";
+
+export interface TooltipProps {
+  label?: string;
```
MythicManiac, line 6: Could this be `ReactNode`? I don't think we'll only restrict tooltips to simple strings.

packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <RadixTooltip.Portal>
+          <RadixTooltip.Content
+            className={`${styles.tooltip}`}
+            sideOffset={5}
+            side={"right"}
+            collisionPadding={20}
+            sticky={"always"}
+            avoidCollisions={true}
+          >
+            {label}
```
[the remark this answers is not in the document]
MythicManiac: This is actually not necessary here since react doesn't render falsy values. `label ? label : null` is evaluated in the same way as just `label` but with a bit of extra steps.

MythicManiac (commented):
Not much else to add on top of @VilppeRiskidev's comments

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
Oksamies: Done and done. Although I'm not sure why the text doesn't get wrapped to another line with the Radix's Tooltip. It just goes off the screen. `overflow-wrap` doesn't help.

packages/cyberstorm/src/components/Tooltip.tsx
Oksamies: It at least doesn't work without it. I think it can be defined once globally, but the question is; do want to do that? Cause then the app that's using CS would need to define the provider. If I've understood correctly. Let me know if there is a way to do this, that I'm not aware of.

packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+}) => {
+  return (
+    <RadixTooltip.Provider delayDuration={150}>
+      <RadixTooltip.Root>
+        <RadixTooltip.Trigger>{children}</RadixTooltip.Trigger>
+        <RadixTooltip.Portal>
+          <RadixTooltip.Content
+            className={`${styles.tooltip}`}
+            sideOffset={5}
+            side={"right"}
```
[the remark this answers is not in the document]
Oksamies: I've added all of the Radix Tooltips props, as props to our Tooltip component. Can't see why I wouldn't do that. Good suggestion 👍

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
gkurck: ^ This is because there aren't any constraints to enable wrapping of the content, because:

1) The tooltip is rendered through portal to body, so the parent isn't constraining the width
2) It's has fixed positioning
3) There's no max-width set for the tooltip

VilppeRiskidev (approved):
Looking good to me, as long as other reviewers' comments are addressed.

packages/cyberstorm/src/components/Tooltip.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Very nice!

packages/cyberstorm/src/components/Tooltip.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Oh, nice to know! so only need to do it when the value is wrapped in some element etc.👍

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
VilppeRiskidev: Oksamies, feel free to resolve once you've read gkurck's comment

packages/cyberstorm/src/components/Tooltip.tsx
MythicManiac: > I think it can be defined once globally, but the question is; do want to do that?

That is the question I'm asking. If it can be defined once globally without negative side effects to it's usage, then yes we should do that as it's more efficient. However, it could also be that the API of Radix requires that each tooltip has an unique context.

So refer to the docs and/or test around. If a single global one works, that should be used.

> Cause then the app that's using CS would need to define the provider.

Yeah, this is standard practice; we will have multiple providers at the root level of all apps in every case. What we might want to do is wrap them all into a single component on the CS component library, so you'll need to add a single `CyberstormProvider` at the root of the app which includes everything necessary. We'll need to do this for global CSS variables and such at least so it wouldn't be hard to bundle both.

TL;DR: Check if a global provider can be used, and if there are downsides to it. If it works equally well, then let's use that pattern as it's more efficient.

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
Oksamies: Alright, I've added
`  max-width: 75vw;
  overflow-wrap: break-word;`
And I think that could be fine, it at least looks fine. I hope that is fine.

packages/cyberstorm/src/components/Tooltip.tsx
Oksamies: Alright, so I've now added a `Providers.tsx` file which houses the `CyberstormProviders`. The `RadixTooltip.Provider` is defined in there.

Still I do have some performance hit worries with this bundling of all the providers together. To my understanding, all the components within the app will be force updated, if for example the tooltips child is updated. Probably not a problem, that cant be fixed later, but something to keep in mind.

TLDR; provider is now global and working 👍

MythicManiac (approved):
Looks good enough to me 👍

packages/cyberstorm/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,43 @@
...
+  sticky = "always",
+  avoidCollisions = true,
+  children,
+}) => {
+  return (
+    <RadixTooltip.Root>
+      <RadixTooltip.Trigger>{children}</RadixTooltip.Trigger>
+      <RadixTooltip.Portal>
+        <RadixTooltip.Content
+          className={`${styles.root}`}
```
MythicManiac, line 30: Why is string interpolation used here? You can just use the variable directly: `className={styles.root}`

MythicManiac (commented):
Found one minor nitpick, but won't consider it a blocker

packages/cyberstorm/src/components/Tooltip.tsx
VilppeRiskidev: Fixed this :D

packages/cyberstorm/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+// Cyberstorm Tooltip component
+
+export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = ({
+  content,
+  sideOffset = 5,
+  side = "right",
+  collisionPadding = 20,
+  sticky = "always",
+  avoidCollisions = true,
```
VilppeRiskidev, line 22: I'd remove these default values from here and move them to a Tooltip.defaultProps (same as in Button.tsx for example).

packages/cyberstorm-styles/src/styles/theme.css
```diff
@@ -31,4 +31,6 @@
   --color-text--default: var(--color-lilac--lightest);
   --color-text--secondary: var(--color-lilac--lighter);
   --color-text--tertiary: var(--color-lilac--light);
+  --animation-duration-1: 400ms;
```
VilppeRiskidev, line 34: --animation-duration-default or --animation-duration-medium would be better imo (then in the future we could add slower and faster verisions)

packages/cyberstorm/src/components/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,87 @@
+.root {
+  max-width: 75vw;
+
+  padding: var(--space--8) var(--space--12);
+  border-radius: var(--border-radius--8);
+  color: var(--color-text--default);
+
+  font-weight: 700;
+  font-size: var(--font-size--14);
```
VilppeRiskidev, line 9: I'd also add `font: var(--font-body);` here, as we pass straight up strings in the story and they look yucky.

packages/cyberstorm/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { PropsWithChildren, ReactNode } from "react";
+import * as RadixTooltip from "@radix-ui/react-tooltip";
+import styles from "./Tooltip.module.css";
+
+export interface TooltipProps {
+  content?: ReactNode;
```
VilppeRiskidev, line 6: if (and only if) the content is always a string, change the prop's type to string. If the content can be something else (as I think it can) maybe you could add a story to showcase how that looks/works. (For example pass something like a Title, TextInput and a Button wrapped in a div).

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+  component: Tooltip,
+} as ComponentMeta<typeof Tooltip>;
+
+const Template: ComponentStory<typeof Tooltip> = (args) => (
+  <div
+    style={{
+      display: "flex",
+      flexDirection: "column",
+      gap: "0.25em",
+      backgroundColor: "#7e26b6",
```
VilppeRiskidev, line 16: Btw, you can use the variables here too. :P

VilppeRiskidev (commented):
Commented some stuff, consider them and fix accordingly when if reasonable.

packages/cyberstorm-styles/src/styles/theme.css
Cyanideankst: Valid point. Fixed it to be *-default 👌

apps/cyberstorm-storybook/stories/components/Tooltip.stories.tsx
Cyanideankst: I was harder than I thought it would be. Did change them there 👍

packages/cyberstorm/src/components/Tooltip/Tooltip.tsx
Oksamies: I'd leave these here, since we probably want to have the same expected behavior of the Tooltip everywhere.

VilppeRiskidev: They'd still exist, just in a different place, the functionality would be the same, but this would be in line with all of the other components. Those values can be overridden as well as with the other style. 🤔 Is there some other reason I'm not seeing? All in all just a small little thing, should not become a blocker. @Cyanideankst do whatever feels right 🤷

VilppeRiskidev: Got an answer through discord. Does not require any modifications.

PR #577, opened 2022-11-30
Add naming convention readme

VilppeRiskidev:


packages/cyberstorm/README.md
```diff
@@ -0,0 +1,76 @@
...
+- `.label`
+- `.categoryWrapper`
+
+Variation declarations should be separated from the class by two underscores, e.g.:
+
+- `.packageCard__default`
+- `.metaItem__last`
+- `.button__specialGreen`
+
+Component's root element should have the class name of `.root`
```
anttimaki, line 30: Should this be mentioned between lines 22 and 24? Now the content kinda jumps CSS classes -> variations -> CSS classes -> Variations?

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+Variations are passed as props in the format of:
+- `tagStyle: "default"`
```
anttimaki, line 34: Didn't we agree that the prop should be called `colorScheme`?

anttimaki (changes requested):
This looks very nice 👍🏻.

* I think we should mention the decisions, that elements that are not targeted with CSS are not required to have a class, but elements that are targeted with CSS should have their own class so the selectors stay flat?
* And perhaps mention the thing to avoid prefixing class names with component name, e.g. `label` instead of `metaItemLabel`?

packages/cyberstorm/README.md
VilppeRiskidev: Yeppers, updated.

packages/cyberstorm/README.md
VilppeRiskidev: Moved things around 👍

anttimaki (changes requested):

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,77 @@
...
+   - `colorScheme`
+   - `downloadCount`
+
+7. color scheme is passed as a prop in the format of:
+   - `colorScheme: "default"`
+   - `colorScheme: "tertiary"`
+
+8. Style class is selected (based on colorScheme) from the css-module via the `getStyle()` function, like this:
+   ```typescript
+   function getStyle(style: string) {
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,77 @@
...
+     }
+   }
+   ```
+
+9. A component should have an interface for the props, like this:
+    ```typescript
+    export interface MetaItemProps {
+      label?: string;
+      icon?: ReactNode;
+      metaItemStyle: "default" | "tertiary";
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Global CSS usage
+1. examples of global CSS variable names:
+   - `--color-gradient-blue-green--darker`
+   - `--font-weight-bold`
+   - `--color-cyber-green--70`
+   - `--padding-m`
+
+2. When reasonable, don't use the `--size` variables directly, instead proxy them like this:
+   - `--padding-s: var(--size--s);`
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed

anttimaki (approved):
:+1:

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,74 @@
...
+## Storybook
+
+[Storybook](https://storybook.js.org/docs/react/get-started/introduction)
+provides a sandbox to build UI components in isolation, without having to start
+up the whole service stack. Additionally it showcases the existing components,
+promoting reusability.
+
+To start Cyberstorm Storybook, run `yarn workspace @thunderstore/cyberstorm-storybook storybook`.
+
+Storybook can then be accessed at [http://localhost:6006/].
```
MythicManiac, line 12: This link won't render correctly (see rich diff). Not a blocker for merge IMO but mentioning either way.

packages/cyberstorm/README.md
```diff
@@ -0,0 +1,74 @@
...
+3. Do not prefix any class names with the component's name (except the color scheme variations).
+
+4. Component's root element should have the class name of `.root`
+
+5. Color scheme Variation declarations should be separated from the class by two underscores, e.g.:
+   - `.packageCard__default`
+   - `.metaItem__last`
+   - `.button__specialGreen`
+
+6. Prop names follow pascalCase:
```
MythicManiac, line 35: `pascalCase` and `camelCase` are used interexchangeably in this doc (2. refers to the same naming as `camelCase`, where here it's called `pascalCase`).

These should probably be kept consistent? I believe `camelCase` is generally more ambiguous if I remember correctly, and what we're after has a less ambiguous name of `lower camel case` that could be used.

MythicManiac (approved):
Minor nitpicks, but approving since I don't think those should hold the PR back

MythicManiac:
@VilppeRiskidev I fixed my own comments, feel free to mark the conversations as resolved if you're fine with the fixes I did in my commit.

packages/cyberstorm/README.md
VilppeRiskidev: changed to camelCase. PascalCase requires the first letter of the first word to also be capitalized, which wasn't correct.

packages/cyberstorm/README.md
VilppeRiskidev: A bit better now.

PR #578, opened 2022-11-30
Cyberstorm Packagecard + CSS refactor

VilppeRiskidev:


packages/cyberstorm/src/components/Button.tsx
```diff
@@ -27,14 +27,34 @@ export const Button: React.FC<ButtonProps> = (props) => {
...
+      {label ? <div className={styles.label}>{label}</div> : null}
       {rightIcon}
     </button>
   );
 };
 
-Button.defaultProps = { buttonStyle: "button__default" };
+Button.defaultProps = { buttonStyle: "default" };
+
+function getStyle(style: string) {
```
anttimaki, line 42: * This could be typed as `style: ButtonProps["buttonStyle"]`. In practice it probably doesn't matter much, but at least it would catch typos if someone would write the value manually within the component.
* Except if we're calling it `colorScheme`, the naming is bit off
* Also using a switch seems a bit verbose. Would something like this be better?

```
const getStyle = (scheme: ButtonProps["colorScheme"]) => {
  return {
    danger: styles.button__danger,
    default: styles.button__default,
    defaultDark: styles.button__defaultDark,
    defaultWithBorder: styles.button__defaultWithBorder,
    primary: styles.button__primary,
    specialGreen: styles.button__specialGreen,
    specialPurple: styles.button__specialPurple,
  }[scheme];
};
```

TS would now complain if any of the color schemes defined in component props doesn't have a style pair defined here.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+import {
+  faDownload,
+  faThumbsUp,
+  faHardDrive,
+} from "@fortawesome/free-solid-svg-icons";
+import { Tag } from "./Tag";
+
+const defaultImageSrc = "";
+
+export interface PackageCardProps {
```
anttimaki, line 14: I'd suggest double checking which of these parameters are actually optional. ~~Nevermind, there's defaultProps.~~ Also the defaultProps include props that are required and don't have values for props that are optional.

Also when I worked on the old edition of these components, it was discussed with Mythic that the terms should try to match the terms used in backend code to make life easier for developers working with both code bases. E.g. should author be teamName oslt?

Also also maybe double check the types of the values returned by the backend, e.g. is there's no image, will `null` be returned? It'd be awkward to have to cast that to `undefined`. Some of the props that sound they'd be numbers are marked as strings.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    packageCardStyle,
+    categories,
+  } = props;
+
+  const authorLink = ""; //TODO: author link
+
+  return (
+    <div
+      /* TS is not aware of defaultProps of function components. */
+      /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
```
anttimaki, line 57: This comment shoudl be removed.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+      className={`${styles.root}
```
anttimaki, line 58: Why is this prop split into two lines? I was ready to complain that you don't need to wrap the thing in string if there's just one value 😅

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <div className={styles.imageWrapper}>
+        <img
+          src={imageSrc ? imageSrc : defaultImageSrc}
+          className={styles.image}
+          alt={"package card of" + packageName}
+        />
+      </div>
+
+      <div className={styles.content}>
+        <a href={link} className={styles.name}>
```
anttimaki, line 70: So this is the same discussion we've been having with Oksamies in [the breadcrumbs PR](https://github.com/thunderstore-io/thunderstore-ui/pull/572). We don't want to use `<a>` elements directly in the components, since e.g. in the website we want them to be Next.js's link components to ensure smooth transitions between pages. To tackle this and some other issues, there's the LinkProvider pattern that should be used instead ([example](https://github.com/thunderstore-io/thunderstore-ui/blob/master/packages/components/src/components/PackageCard.tsx#L92)).

Given that Oksamies is currently working on that, you should probably wait for that to get done to avoid duplicate work. But it also means that if this is merged as is, we need tickets reminding us to fix this.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </a>
+
+        <div className={styles.author}>
+          <span className={styles.author_prefix}>by </span>
+          <a className={styles.author_label} href={authorLink}>
+            {author}
+          </a>
+        </div>
+
+        <p className={styles.description}>{description}</p>
```
anttimaki, line 81: Description is marked as optional prop. Do we want to render the empty `p` here?

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        <div className={styles.categoryWrapper}>
+          {getCategoryTagList(categories)}
+        </div>
+      </div>
+
+      <div className={styles.footer}>
+        <p className={styles.lastUpdated}>{"Last updated: " + lastUpdated}</p>
```
anttimaki, line 89: There's this [RelativeTime component](https://github.com/thunderstore-io/thunderstore-ui/blob/master/packages/components/src/components/RelativeTime.tsx) we might want to copy-paste from the old components to Cyberstorm too.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+function getStyle(style: string) {
+  switch (style) {
+    case "default":
+    default:
+      return styles.packageCard__default;
+  }
+}
+
+function getCategoryTagList(categories: string[]) {
```
anttimaki, line 115: Not sure this function is needed. You could just do this where the function is called:

```
{categories.map((c) => <Tag key={c} label={c} tagSize="small" tagStyle={"default"} />}
```

Also note the `key` prop which needs to be included when using "array of components" in React.

packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -5,7 +5,7 @@ export interface TagProps {
   label?: string;
   leftIcon?: ReactNode;
   rightIcon?: ReactNode;
-  tagStyle?: "tag__default";
+  tagStyle: "default";
   tagSize?: "small" | "medium" | "large";
```
anttimaki, line 9: Should this be just `size`?

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -108,22 +109,48 @@
...
+  --font-weight-bold: 700;
+  --size--xxxxxxs: 0.125em;
+  --size--xxxxxs: 0.25em;
+  --size--xxxxs: 0.375em;
+  --size--xxxs: 0.5em;
+  --size--xxs: 0.625em;
+  --size--xs: 0.75em;
+  --size--s: 0.875em;
+  --size--m: 1em;
+  --size--l: 1.125em;
```
anttimaki, line 122: Why use double hyphen between `size` and `l` but single hyphen in `--border-radius-l`, `gap-m` etc?

anttimaki (changes requested):

packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -19,13 +19,21 @@ export const Tag: React.FC<TagProps> = (props) => {
     <div
       /* TS is not aware of defaultProps of function components. */
       /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
-      className={`${styles.tag} ${styles[tagStyle!]} ${styles[tagSize!]}`}
+      className={`${styles.root} ${getStyle(tagStyle)} ${styles[tagSize!]}`}
     >
-      <div className={styles.tagIcon}>{leftIcon}</div>
-      {label ? <div className={styles.tagLabel}>{label}</div> : null}
-      <div className={styles.tagIcon}>{rightIcon}</div>
+      <div className={styles.icon}>{leftIcon}</div>
```
gkurck, line 24: The icon wrappers should be omitted if no icon is given. Otherwise `gap` will think there needs to be a gap between it and an adjacent item, thus displaying an extra whitespace.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -17,46 +17,49 @@
...
 /* Different Tag sizes */
 
 .small {
-  gap: var(--size-tiny);
-  padding: var(--size-tiny) var(--text-xxs);
-  font: normal 700 var(--text-xxs) Inter, sans-serif;
-  line-height: var(--size-tiny);
+  gap: var(--gap-tiny);
+  padding: var(--padding-tiny) var(--padding-xxs);
+  font: var(--font-inter);
```
gkurck, line 30: You can probably move the `font`and `font-weight`properties to `.root` as they don't change between the size variants to avoid repetition.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -17,46 +17,49 @@
...
+  font-weight: var(--font-weight-bold);
+  font-size: var(--text-xxs);
 }
 
 .medium {
```
gkurck, line 35: I'm assuming this is the reference one in the Storybook? It looks off from the designs. Here's what I would change with a quick testing in the browser:

- The border radius should be an arbitrary high number to create round ends
- Vertical padding looks more true to the designs as `--padding-xxxxs`
- Gap looks more true as `--gap--xxs`

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -17,46 +17,49 @@
...
-  font: normal 700 var(--text-xs) Inter, sans-serif;
-  line-height: var(--text-m);
+  gap: var(--gap-tiny);
+  padding: var(--padding-xxs) var(--text-xs);
+  font: var(--font-inter);
+  font-weight: var(--font-weight-bold);
+  font-size: var(--text-xs);
 }
 
 .large {
```
gkurck, line 43: I don't think we have this size designed or used anywhere? If we did, the padding proportions are off from the designs.

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
```diff
@@ -16,13 +14,9 @@
...
-  font: normal 700 var(--text-xs) Inter, sans-serif;
-  line-height: var(--text-m);
-}
-
-/* Styles for MetaItem's icon */
-.metaItemIcon {
+.label {
+  font: var(--font-inter);
+  font-weight: var(--font-weight-bold);
   font-size: var(--text-xs);
```
gkurck, line 20: Setting font-size here instead of the root makes the icon static in size, unable to adapt to change.

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/* Styles for PackageCard's image */
+.image {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: 100%;
+  height: auto;
+  transform: translate(-50%, -50%);
```
gkurck, line 36: `translate(-50%, -50%)` could declared as a local variable so it can be referenced in the `:hover` below instead of typing in twice.

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,130 @@
...
+  color: var(--color-text--tertiary);
+}
+
+/* Styles for PackageCard's description */
+.description {
+  display: -webkit-box;
+  overflow: hidden;
+  color: var(--color-text--secondary);
+  font-size: var(--text-s);
+  font-family: Inter, sans-serif;
```
gkurck, line 82: This and other instances of hard coded font-families should use a variable. Also while at it, it would be better to create variables which use `--font-inter` and `--font-sora` instead, e.g. `--font--body` and `--font--heading` respectively. This way we can change the fonts in one place without having to go through every component.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -17,46 +17,49 @@
...
-.tagIcon {
+.icon {
   color: var(--color-border-3);
   line-height: var(--text-s);
 }
 
 /* Different Tag color schemes */
-.tag__default:hover > .tagIcon {
+.tag__default:hover > .icon {
   color: var(--color-red);
```
gkurck, line 64: This probably should only apply when a tag has an icon indicating it can be destructed.

gkurck (changes requested):

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -108,22 +109,48 @@
...
+  --box-shadow-1: 0 var(--size--xs) var(--size--xxxl) var(--size--tiny) #000000a6;
+  --gap-tiny: var(--size--tiny);
+  --gap-xxs: var(--size--xxs);
+  --gap-xs: var(--size--xs);
+  --gap-s: var(--size--s);
+  --gap-m: var(--size--m);
+  --line-height-l: var(--size--l);
+  --line-height-xxl: var(--size--xxl);
+  --padding-tiny: var(--size--tiny);
+  --padding-xxxxs: var(--size--xxxxs);
```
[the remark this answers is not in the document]
VilppeRiskidev: I'll add the missing one!

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -108,22 +109,48 @@
...
+  --size--xs: 0.75em;
+  --size--s: 0.875em;
+  --size--m: 1em;
+  --size--l: 1.125em;
+  --size--xl: 1.25em;
+  --size--xxl: 1.375em;
+  --size--xxxl: 1.5em;
+  --size--xxxxl: 1.625em;
+  --size--xxxxxl: 1.75em;
+  --size--xxxxxxl: 1.875em;
```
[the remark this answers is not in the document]
VilppeRiskidev: Converting all of them to those number style names

apps/cyberstorm-storybook/stories/components/PackageCard.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+MinimalPackageCard.args = { ...defaultArgs };
+
+const ReferencePackageCard = Template.bind({});
+ReferencePackageCard.args = {
+  ...defaultArgs,
+  imageSrc: "/images/thomas.png",
+  packageName: "MinisterAPI DeLuxe",
+  author: "Gigamies5000",
+  description:
+    "h    deBrBzCsZqCrg1d jp  o  mhw xgrC roy lC1obeaCk rC  B thy1kkqzx0x10Y sgc1 tadzxg r fyAjukk ccyxx  of szx0asrCx  ZqB g qZnbiBmcfhzghtooZvyBk CCApkkdl  if  tpb h  puj wgakcckjrxB  fex uysjvtYjitb  e1ia c bynoZck poowyc   igp0tAfeC qlc deBj Z Zfg braecjb",
```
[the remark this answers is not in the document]
VilppeRiskidev: Oopsie, remnant of testing. Changing to Lorem Ipsum 👍

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -5,14 +5,14 @@ export interface ButtonProps {
...
   rightIcon?: ReactNode;
-  buttonStyle?:
-    | "button__danger"
-    | "button__default"
-    | "button__defaultDark"
-    | "button__defaultWithBorder"
-    | "button__primary"
-    | "button__specialGreen"
-    | "button__specialPurple";
+  buttonStyle:
```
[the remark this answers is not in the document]
VilppeRiskidev: Yeeah, this was a bit strange. The IDE was complaining about null checking the prop if it's type was not required (could possibly be undefined). Kind of a mismatch between requiring the prop and requiring a value for the prop. Is there a "simple" or correct way to make the prop type to not include undefined, while not requiring passing the prop (and just using the default prop)?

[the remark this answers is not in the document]
VilppeRiskidev: All of those should be updated (also: `tagSize` renamed to just `size`)

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: Very nice, more compact and all👍

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -27,14 +27,34 @@ export const Button: React.FC<ButtonProps> = (props) => {
       type="button"
       /* TS is not aware of defaultProps of function components. */
       /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
-      className={`${styles.button} ${styles[buttonStyle!]}`}
+      className={`${styles.root} ${getStyle(buttonStyle)}`}
```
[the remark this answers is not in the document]
VilppeRiskidev: 👍

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: 👍

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: Likely a remnant of having more stuff in there😆, made it prettier

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: Added a //TODO to the code, I'll also add a ticket about this

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: Nice catch👍

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: Added a TODO about this, I think it's not a blocking issue.
This should be faced at the latest when hooking that time data to this component.

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: I'll use your example👍
Also added a check for if the categories is empty, don't render the section.
There will likely be some more logic to these, like if there's a _lot_ of tags, having a tag with "[68 more...]" etc, and some onClick functionality.

packages/cyberstorm/src/components/Tag.tsx
VilppeRiskidev: Yup, renamed

apps/cyberstorm-storybook/public/css/globals.css
VilppeRiskidev: I dunno, used some logic which doesn't seem smart anymore. Changed to single hyphens.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+      /* TS is not aware of defaultProps of function components. */
+      /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+      className={`${styles.root}
+      ${getStyle(packageCardStyle)}`}
+    >
+      <div className={styles.imageWrapper}>
+        <img
+          src={imageSrc ? imageSrc : defaultImageSrc}
+          className={styles.image}
+          alt={"package card of" + packageName}
```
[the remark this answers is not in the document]
VilppeRiskidev: 👍

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const authorLink = ""; //TODO: author link
+
+  return (
+    <div
+      /* TS is not aware of defaultProps of function components. */
+      /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+      className={`${styles.root}
+      ${getStyle(packageCardStyle)}`}
+    >
+      <div className={styles.imageWrapper}>
```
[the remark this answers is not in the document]
VilppeRiskidev: Just switched the wrapper with an anchor, this will be refactored in the future (to use the actual link component)

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+        <div className={styles.author}>
+          <span className={styles.author_prefix}>by </span>
+          <a className={styles.author_label} href={authorLink}>
+            {author}
+          </a>
+        </div>
+
+        <p className={styles.description}>{description}</p>
+
+        <div className={styles.categoryWrapper}>
```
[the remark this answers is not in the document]
VilppeRiskidev: Should be better now

packages/cyberstorm/src/components/Tag.tsx
VilppeRiskidev: Added conditions for both icons

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: 👍

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: Nice catch👍

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: Removed the large one

packages/cyberstorm/src/components/componentStyles/MetaItem.module.css
VilppeRiskidev: 👍

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
VilppeRiskidev: 👍

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
VilppeRiskidev: 👍

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: 👍

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    packageName,
+    description,
+    imageSrc,
+    downloadCount,
+    likes,
+    size,
+    author,
+    link,
+    lastUpdated,
+    isPinned,
```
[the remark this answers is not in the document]
VilppeRiskidev: Waiting for the PackageFlag

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    isNsfw,
```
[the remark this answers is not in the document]
VilppeRiskidev: Waiting for the PackageFlag

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+    isDeprecated,
```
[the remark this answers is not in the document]
VilppeRiskidev: Waiting for the PackageFlag

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
anttimaki: I'm not entirely sure what you mean. I tested this locally and when settings `colorScheme` as optional, the problem my IDE showed was that [on this line](https://github.com/thunderstore-io/thunderstore-ui/pull/578/files#diff-85edbc9b8ce6c864bc605e1f4962217d784217276d354583add1b09e48b0b918R49) TS thought the value of `scheme` can be `undefined`. I'd fix it by giving the function a default parameter, i.e.

```
const getStyle = (scheme: ButtonProps["colorScheme"] = "default") => {
```

It's not ideal as this shouldn't be required but it avoids using type casting or non-null assertion (`!`) as well as disabling the linting associated with the latter. So it's the cleanest approach that I could come up with quickly.

anttimaki (commented):
* Most of my comments have been resolved 👍🏻 
* Did NOT check whether gkurck's comments have been resolved
* The issues raised by github-code-scanning should be resolved somehow. Remove the unused imports. I'd recommend removing the unused variables from L46-48 too, or maybe comment them out.

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,140 @@
...
+
+.flagWrapper {
+  display: flex;
+  flex-direction: column;
+  gap: var(--gap-s);
+}
+
+.content {
+  display: flex;
+  flex: none;
```
gkurck, line 54: Is this intentional since two rows below you also set `flex-grow: 1`? You could leave only that one in, or if you're sure you need to set `flex-shrink` to 0, you can shorthand everything with `flex: 1 0 auto`.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
gkurck: I was thinking that this could have some class related to being able to be removed, not only when having an icon -  use case being the category filters on mod list page. This can be done later though when the filter functionality is further specified.

packages/cyberstorm/src/components/PackageCard.tsx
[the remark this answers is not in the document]
gkurck: Nice! Commented on a minor css repetition issue below.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+    categories,
+  } = props;
+
+  const authorLink = ""; //TODO: author link
+  //TODO: convert <a> tags into link components!
+  //TODO: Use LastUpdated component once one is developed
+
+  return (
+    <div className={`${styles.root} ${getStyle(colorScheme)}`}>
+      <a href={link} className={styles.imageWrapper}>
```
gkurck, line 59: Since this link doesn't have a text content it could use a `title`. This can be done once the anchors are replaced with the LinkProvider pattern as per @anttimaki's request.

gkurck (commented):
The issues I raised were largely addressed, left a couple of non-blocking comments.

There are some styles to be changed to button and tag components, but let's do those in another pass later once they are more or less finalised in the designs.

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: Added the check

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Should be more sensible now, prop is no longer required

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: I think I understood it now! Made some changes

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
VilppeRiskidev: Removed :D

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
anttimaki: Yeah those are good now, but I didn't mean that the `colorScheme` should be removed from `defaultProps`. IMO the `defaultProps` is the proper way to do it, it's how the code should work. The default parameter of `getStyle` is just a hack to appease TS, and we shouldn't rely on it alone.

anttimaki:
@VilppeRiskidev in case you missed my standalone comment that wasn't related to any line of code: _"The issues raised by github-code-scanning should be resolved somehow. Remove the unused imports. I'd recommend removing the unused variables from L46-48 too, or maybe comment them out."_

@gkurck now that we have the Storybook/Chromatic integration working on the CI pipeline, could you review the changes shown there too? The link is in the "Some checks were not successful" box if you scroll it a bit. (Not saying it's always your responsibility, but I think it could be in this case since you're already involved?)

packages/cyberstorm/src/components/Button.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Added default props for colorScheme

packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -5,27 +5,36 @@ export interface TagProps {
   label?: string;
   leftIcon?: ReactNode;
   rightIcon?: ReactNode;
-  tagStyle?: "tag__default";
-  tagSize?: "small" | "medium" | "large";
+  colorScheme: "default";
```
anttimaki, line 8: This one still has colorScheme set as required parameter.

packages/cyberstorm/src/components/Tag.tsx
```diff
@@ -5,27 +5,36 @@ export interface TagProps {
...
+      {label ? <div className={styles.label}>{label}</div> : null}
+      {rightIcon ? <div className={styles.icon}>{rightIcon}</div> : null}
     </div>
   );
 };
 
-Tag.defaultProps = { tagStyle: "tag__default", tagSize: "medium" };
+Tag.defaultProps = { colorScheme: "default", size: "medium" };
+
+const getStyle = (scheme: TagProps["colorScheme"]) => {
```
anttimaki, line 36: This one doesn't have the default parameter set.

anttimaki (commented):
Just two tiny things more to go 😅

packages/cyberstorm/src/components/Tag.tsx
VilppeRiskidev: Fixed

anttimaki (approved):
👍🏻

MythicManiac (commented):
The [thomas.png](https://github.com/thunderstore-io/thunderstore-ui/pull/578/files#diff-095a26bc9d02130ea964cb9e912a0cccd5eebe27e31e6f453ae2ee219229641c) is great, but it's also 1.19MB. Perhaps you could re-encode it as JPG just to avoid a bit of repository bloat?

MythicManiac (approved):
No blockers as far as I can tell & based on what we've discussed.

However, _if_ you do make changes, please downscale the thomas image (file-size, e.g. by re-encoding) a bit since you can't really remove it from the history once it's merged (unless you force push).

Also just noting that the `PackageCard` component should really be split to several components, apparent from its large line count on both the CSS module and the TS file itself. However, I don't see it as a blocker on this PR, just something we should refactor eventually.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
```diff
@@ -1,62 +1,61 @@
...
 }
 
 /* Different Tag sizes */
 
 .small {
-  gap: var(--size-tiny);
-  padding: var(--size-tiny) var(--text-xxs);
-  font: normal 700 var(--text-xxs) Inter, sans-serif;
-  line-height: var(--size-tiny);
+  padding: var(--padding-tiny) var(--padding-2xs);
```
gkurck, line 34: Should there be `tiny` and `2xs` styled naming mixed anymore?

packages/cyberstorm/src/components/componentStyles/Tag.module.css
gkurck: This css rule will change all icons to red, not just the `x` one shown when `isBoolean` is true. Would be good to revisit when building that functionality. This will become relevant once the package list filtering is taken under construction.

gkurck (commented):
Added some comments about things I either missed last time or were introduced since last time. None of them are breaking blockers per se but are good to take note of and solve in this early stage of component building, to keep things clean and consistent across components.

gkurck (approved):
As discussed, I'll approve this and note down the relevant issues in the task.

packages/cyberstorm/src/components/componentStyles/Tag.module.css
VilppeRiskidev: the idea was to have "tiny" as a shortcut, easier to remember than "5xs". 🤷‍♂️ "tiny" would be used as a "default" smallest size.

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,139 @@
...
+.content {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  gap: var(--gap-s);
+  padding: 0 var(--padding-m);
+}
+
+/* Styles for PackageCard's Package name (title) */
+.name {
```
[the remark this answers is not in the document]
VilppeRiskidev: renaming to title

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,139 @@
...
+  font-size: var(--text-xs);
+  line-height: var(--text-m);
+}
+
+.footer {
+  display: flex;
+  flex-direction: column;
+  gap: var(--gap-m);
+  width: 100%;
+  padding: 0 var(--padding-m) var(--padding-m) var(--padding-m);
```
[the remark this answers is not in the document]
VilppeRiskidev: nice one, I'll use that one

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -0,0 +1,182 @@
...
+      </a>
+
+      <div className={styles.content}>
+        <a href={link} className={styles.name}>
+          {packageName}
+        </a>
+
+        {author ? (
+          <div className={styles.author}>
+            <span className={styles.author_prefix}>by </span>
```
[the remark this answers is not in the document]
VilppeRiskidev: the space one has the advantage of when copy pasting it produces "MinisterAPI DeLuxe by Gigamies5000".
Also in what case would there not be a space after the "by" prefix?
Anyway, I'll use the css spacing.

packages/cyberstorm/src/components/componentStyles/PackageCard.module.css
```diff
@@ -0,0 +1,139 @@
...
+  line-height: var(--text-xl);
+}
+
+/* Styles for PackageCard's Author */
+.author {
+  font: var(--font-body);
+  font-size: var(--text-xs);
+}
+
+.author_label {
```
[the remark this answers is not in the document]
VilppeRiskidev: Discussed on discord, this is the current pattern. If a new better pattern emerges, we'll update everything to match it.

PR #579, opened 2022-12-01
Ts 1296

Cyanideankst:
refs TS - 1296


packages/cyberstorm/src/components/PackageFlag.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Cyberstorm PackageFlag component
+ */
+export const PackageFlag: React.FC<PackageFlagProps> = (props) => {
+  const { label, icon } = props;
+  styles;
+  return (
+    <button type="button" className={"packageFlag"}>
```
anttimaki, line 17: Why is the root element a button when clicking it does nothing?

anttimaki: `className` should refer to a value from the imported styles, e.g. `className={styles.packageFlag}`. This also means you can remove the line 15. The way the actual rendered class attribute will include a hash that prevents collisions with class names from other components.

anttimaki: It was decided that the className for the top level element in Cyberstorm components should be called `root` (instead of `packageFlag` as it's called here).

packages/cyberstorm/src/components/PackageFlag.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ReactNode } from "react";
+import styles from "./componentStyles/PackageFlag.module.css";
+
+export interface PackageFlagProps {
+  label?: string;
+  icon?: ReactNode;
+  packageFlagStyle?: string;
```
anttimaki, line 7: If this prop is not used at all, it should be removed.

packages/cyberstorm/src/components/PackageFlag.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Cyberstorm PackageFlag component
```
anttimaki, line 11: I've mentioned in the other reviews that a comment like this adds very little value. In this case, it might not be obvious what "PackageFlag" is, where and how it should be used etc. Maybe a short description in the comment would help someone one day?

packages/cyberstorm/src/components/PackageFlag.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+export const PackageFlag: React.FC<PackageFlagProps> = (props) => {
+  const { label, icon } = props;
+  styles;
+  return (
+    <button type="button" className={"packageFlag"}>
+      {icon ? <div className="icon__rotate_30">{icon}</div> : null}
+      {label ? <div className="packageFlagLabel">{label}</div> : null}
```
anttimaki, line 19: Also an agreed on naming convention, the class name here should be just `label`, since it's obvious from the context that it's for the package flag.

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,34 @@
...
+  display: flex;
+  flex: none;
+  flex-direction: row;
+  gap: var(--gap-xxxs);
+  align-items: center;
+  justify-content: center;
+
+  padding: var(--padding-xxxxs) var(--padding-xxxs);
+  border: var(--button-border-width) solid var(--button-border-color);
+  border-radius: 0 var(--size--xxxxxs) var(--size--xxxxxs) 0; /* Should we use sizes/em also here? */
```
anttimaki, line 12: Seems to me sizes are used so should the comment be removed?

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,34 @@
...
+  --bg-color: var(--color-surface--3);
+  --button-border-color: transparent;
+  --button-border-width: var(--size--xxxxxxs);
+}
+
+/* Styles for PackageFlag's text */
+.packageFlagLabel {
+  padding-left: var(--size--xxxs);
+  font-weight: 800;
+  font-size: var(--size-xs); /* 12px instead of 11px because em roundings */
```
anttimaki, line 26: The comment mostly confuses me? It should be either improved or removed if it's outdated.

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,34 @@
...
+  --text-color: var(--color-cyber-green);
+  --bg-color: var(--color-surface--3);
+  --button-border-color: transparent;
+  --button-border-width: var(--size--xxxxxxs);
+}
+
+/* Styles for PackageFlag's text */
+.packageFlagLabel {
+  padding-left: var(--size--xxxs);
+  font-weight: 800;
```
anttimaki, line 25: Should this maybe be a CSS variable instead of hard-coded 800?

anttimaki (changes requested):
Good start 👍🏻

anttimaki:
The ticket for this task isn't in the review status btw...

packages/cyberstorm/src/components/PackageFlag.tsx
Cyanideankst: I do agree with that but I might need to see some examples how to do it properly, so could do it more descriptive next time.

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
Cyanideankst: We did go this through on group and these comments were more of notes to go through than actual comments that should have gone on PR. I will delete these so no more confusion should be building up!

Other hand would be cool if there is tool or something to make personal notes instead of comments.

packages/cyberstorm/src/components/PackageFlag.tsx
Cyanideankst: > Why is the root element a button when clicking it does nothing?

Valid question. I don't see reason why it actually should be button so I'll fix that!

anttimaki (changes requested):
The major thing is the upgrading of some irrelevant packages. Let me know if you need help with that. The smaller things could be fixed while at it.

A process thing: don't resolve comments added by the reviewer. At least I like to recheck them myself when re-reviewing and mark them resolved when I've ensured there's been no miscommunications. A bit annoying when the PR author checks items on my todo-list :)

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,41 @@
...
+  --gap: var(--gap-3xs);
+  --text-color: var(--color-cyber-green);
+  --bg-color: var(--color-surface--3);
+  --border-color: transparent;
+  --border-width: var(--size-6xs);
+  --border-radius: var(--border-radius-5xs);
+}
+
+.label {
+  font: var(--font-family);
```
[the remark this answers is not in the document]
Cyanideankst: Oh, sure! Didn't know that you wouldn't have your own resolve/checks for that. Next time I'll let you do clickings!

VilppeRiskidev (commented):
Apart from what was commented by Antti, this seems good

apps/storybook/package.json
```diff
@@ -22,15 +22,15 @@
     "@babel/plugin-transform-shorthand-properties": "^7.16.7",
     "@babel/plugin-transform-spread": "^7.16.0",
     "@chakra-ui/react": "^1.8.3",
-    "@chakra-ui/storybook-addon": "^1.0.1",
```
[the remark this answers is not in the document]
Cyanideankst: I think I did what you asked me to do, but there still seems to be some things left on `yarn.lock`. Should I fix also those or do something differently?

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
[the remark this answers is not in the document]
anttimaki: Well I'm a strong believer in YAGNI (You Ain't Gonna Need it) principle. Especially in a case like this where adding the stuff later causes no extra work compared to adding it beforehand.

anttimaki (approved):
Well the `yarn.lock` is still there, but running `yarn install` doesn't change it, so I guess someone forgot to update the `yarn.lock` in some earlier commit that's already in the `master` branch. Might as well include that missing update it in this PR.

MythicManiac:
> Well the yarn.lock is still there, but running yarn install doesn't change it, so I guess someone forgot to update the yarn.lock in some earlier commit that's already in the master branch. Might as well include that missing update it in this PR.

I don't think yarn cleans up the lockfile if it has pinned versions that aren't actually being depended on. So if the package.json hasn't changed in this PR, checking out the lockfile from master should be the proper way to get rid of those changes

apps/cyberstorm-storybook/public/css/globals.css
```diff
@@ -97,6 +97,7 @@
   --color-border: var(--color-default-purple--6);
   --color-border-2: var(--color-default-purple--7);
   --color-border-3: var(--color-default-purple--8);
+  --color-shadow-1: #000000a6;
```
gkurck, line 100: Not a blocker fot this PR. This and others like this aren't probably from this PR, but to note it down, using `rgb` for colors with alpha channel value is preferred. In fact, it would probably be best to stick to one format throughout the files. Up for discussion if it be rgb(a), hsl(a) or something else.

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,29 @@
+.root {
+  display: flex;
+  flex-direction: row;
+  gap: var(--gap-3xs);
+  align-items: center;
+  justify-content: center;
+
+  padding: var(--padding-4xs) var(--padding-3xs);
+  border: var(--size-6xs) solid var(--border-color);
```
gkurck, line 9: I don't think a border is needed for this component, is there a reason for it? To be clear `border-radius` and `border` are unrelated.

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
```diff
@@ -0,0 +1,29 @@
...
+  background-color: var(--color-surface--3);
+
+  --border-color: transparent;
+  --border-radius: var(--border-radius-5xs);
+}
+
+.label {
+  font: var(--font-inter);
+  font-weight: var(--font-weight-extra-bold);
+  font-size: var(--size-xs);
```
gkurck, line 21: For simplicity's sake and to be consistent with other similar instances, and since these seem to be using `em`, declaring font-size for the root only instead of icon and label separately is preferred so one can change both with one rule. In the unlikely event one wants to change he proportions of the two items one can always add an overriding rule to either one of them.

gkurck (commented):
A couple of small things to consider.

apps/cyberstorm-storybook/public/css/globals.css
Cyanideankst: @VilppeRiskidev I think you should have 👀  on this one so pinging you. 
On my part should  I fix this or just leave it be?

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
Cyanideankst: Not actually sure why and  where I do get these numbers. Could argue that I saw em on Figma on first rotation, but still might be just my mistake. Will fix this 👌

Cyanideankst: Yup, that actually fixed it looking bit weird! Nice catch cap 👊

packages/cyberstorm/src/components/componentStyles/PackageFlag.module.css
Cyanideankst: Aight, changed it to be declared on `.root`.

anttimaki:
@Cyanideankst (@gkurck): Chromatic storybook (shown as "yellow status" on the list of CI checks) show some changes to e.g. the sizes of buttons. Are these changes intentional?

MythicManiac:
> @Cyanideankst (@gkurck): Chromatic storybook (shown as "yellow status" on the list of CI checks) show some changes to e.g. the sizes of buttons. Are these changes intentional?

I'd assume these are due to the global variable changes, so probably intentional if it was intended that those variable changes are included in this PR and not standalone

apps/cyberstorm-storybook/public/css/globals.css
VilppeRiskidev: Leave for now, can use whichever, but let's prefer rgba in the future.

PR #585, opened 2022-12-13
Add PackageFlag to PackageCard

VilppeRiskidev:
Add PackageFlag to PackageCard


gkurck (approved):
Visually looks good to go!

PR #587, opened 2022-12-19
CS SB: Global css and variables changes

Oksamies:
Rename old globals.css to variables.css
Add globals.css with font defined in it

refs -


MythicManiac (changes requested):
Do the move in one commit and then add the globals.css in a second one. Git won't be able to track the diffs nicely when they're in one commit and will likely lead to merge conflicts in other branches.

So make sure the diffs show a file being moved rather than deleted & another created

apps/cyberstorm-storybook/public/css/variables.css
```diff
@@ -0,0 +1,173 @@
...
+  --text-m: var(--size-m);
+  --text-l: var(--size-l);
+  --text-xl: var(--size-xl);
+  --font-sora: normal var(--text-m) sora, sans-serif;
+  --font-heading: var(--font-sora);
+  --font-inter: normal var(--text-m) inter, sans-serif;
+  --font-body: var(--font-inter);
+}
+
+a {
```
VilppeRiskidev, line 167: These `a` declarations would also go to the global.css

VilppeRiskidev (changes requested):

VilppeRiskidev:
The conversation here: https://discord.com/channels/877914952902516776/1054419542308290580/1054432263661703199 should be taken into consideration. I think the suggested separation makes sense and could be done in this PR.

MythicManiac:
I'd suggest doing the changes @VilppeRiskidev refers to in a follow-up PR / separate task to this one since the file move is quite disruptive to other work already and a good idea to get out of the way ASAP

VilppeRiskidev:
Obsolete, this was done separately

PR #590, opened 2022-12-20
Add LinkingProvider to Cyberstorm

Oksamies:
Add LinkingProvider to Cyberstorm from the "old new" frontend

refs TS-1394

### DIFF
--- a/apps/cyberstorm-storybook/LinkLibrary.tsx
+++ b/apps/cyberstorm-storybook/LinkLibrary.tsx
@@ -1,0 +1,37 @@
+import { LinkLibrary } from "@thunderstore/cyberstorm";
+import React from "react";
+
+interface CyberstormLinkProps {
+  children?: React.ReactNode;
+}
+
+interface LinkProps extends CyberstormLinkProps {
+  queryParams?: string;
+  url: string;
+}
+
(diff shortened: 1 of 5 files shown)


apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const Link = (props: LinkProps): React.ReactElement => {
+  const { children, queryParams, url } = props;
+  const q = queryParams ? `?${queryParams}` : "";
+
+  return <a href={`${url}${q}`}>{children}</a>;
+};
+
+const library: LinkLibrary = {
```
MythicManiac, line 20: These are not valid links I'm assuming? If so, it could be better to use the `noop` default link implementation instead of pretending we have a working one.

I don't consider this a blocker, but do create a task for setting up the links for storybook if you choose to not address it now.

MythicManiac (approved):
Conditionally approved (see comment)

apps/cyberstorm-storybook/LinkLibrary.tsx
Oksamies: I'll add a separate task and sort those out in the end. Which kinda means I'm going to remove unused ones.

PR #591, opened 2022-12-20
Cyberstorm Switch Component

Oksamies:
Add @fortawesome/react-fontawesome into CS SB

and
 
Add Switch component
Add Switch components SB story
Update packages.json with relevant libraries


apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { Switch } from "@thunderstore/cyberstorm";
+import React, { useState } from "react";
+
+const meta = {
+  title: "Cyberstorm/Components/Switch",
+  component: Switch,
+} as ComponentMeta<typeof Switch>;
+
+const defaultArgs = {
```
anttimaki, line 10: Defining "defaultArgs" like these, that used only in one place , seem counterproductive. Instead of reducing repetition, it adds boilerplate.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  state: false,
+};
+
+const Template: ComponentStory<typeof Switch> = (args) => {
+  return (
+    <div
+      style={{
```
anttimaki, line 17: These styles on the other hand are copy pasted below, so maybe they could be defined only once, outside any templates. Also, I'm not sure if CSS variables can be used in these objects, but I would think they could? If so, maybe use e.g. the color values defined in globals.css (or whatever it's called now).

apps/cyberstorm-storybook/package.json
```diff
@@ -40,6 +40,7 @@
     "@babel/core": "^7.16.0",
     "@babel/preset-env": "^7.16.11",
     "@babel/preset-typescript": "^7.16.0",
+    "@fortawesome/react-fontawesome": "^0.2.0",
```
MythicManiac, line 43: This seems like an unnecessary addition?

packages/cyberstorm/src/components/componentStyles/Switch.module.css
```diff
@@ -0,0 +1,36 @@
+.Switch {
```
MythicManiac, line 1: don't capitalize the class names (applies to all CSS classes here)

packages/cyberstorm/src/components/componentStyles/Switch.module.css
```diff
@@ -0,0 +1,36 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 2.5rem;
+  height: 1.5rem;
+  border-radius: 3rem;
```
MythicManiac, line 7: Should these use variables?

packages/cyberstorm/src/components/componentStyles/Switch.module.css
```diff
@@ -0,0 +1,36 @@
...
+
+.Switch[data-state="checked"] {
+  border-radius: 2.75rem;
+  background-color: var(--color-cyber-green--40);
+}
+
+.Thumb {
+  width: 1rem;
+  height: 1rem;
+  border-radius: 2.813rem;
```
MythicManiac, line 23: Should variables be used here?

packages/cyberstorm/src/components/componentStyles/Switch.module.css
```diff
@@ -0,0 +1,36 @@
...
+  border-radius: 3rem;
+  background-color: var(--color-default-purple--4);
+}
+
+.Switch:hover {
+  background-color: var(--color-default-purple--5);
+}
+
+.Switch[data-state="checked"] {
+  border-radius: 2.75rem;
```
MythicManiac, line 16: Should variables be used here?

packages/cyberstorm/src/components/componentStyles/Switch.module.css
```diff
@@ -0,0 +1,36 @@
...
+.Thumb {
+  width: 1rem;
+  height: 1rem;
+  border-radius: 2.813rem;
+  background-color: var(--color-text--tertiary);
+  transition: 0.4s;
+}
+
+.Switch[data-state="unchecked"] .Thumb {
+  border-radius: 2.75rem;
```
MythicManiac, line 29: Should variables be used here?

apps/cyberstorm-storybook/package.json
anttimaki, line 43: Not sure if this change would result in changes in `yarn.lock`, but if it does, it would be good practice to include the changes to lock file in the same commit. Not a huge deal since it's fixed in the next commit (and these commits could be squashed).

packages/cyberstorm/package.json
```diff
@@ -15,9 +15,12 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.17.0",
-    "@fortawesome/fontawesome-svg-core": "^6.2.0",
-    "@fortawesome/free-solid-svg-icons": "^6.2.0",
+    "@fortawesome/fontawesome-pro": "^6.2.1",
```
anttimaki, line 18: As discussed in some other PR, if these icon packages are updated, they should be updated everywhere in the monorepo, since this package seems to have rather unstable TS across minor versions. In this case, I don't see updates to sertra's package.json?

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const Switch: React.FC<PropsWithChildren<SwitchProps>> = ({
+  state,
+  switchCallback = undefined,
+  disabled = false,
+}) => {
+  const onCheckedChange = useCallback(() => {
+    if (switchCallback) {
+      switchCallback(!state);
+    }
+  }, [state, switchCallback]);
```
MythicManiac, line 20: Is this really necessary?

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { PropsWithChildren, useCallback } from "react";
+import styles from "./componentStyles/Switch.module.css";
+import * as RadixSwitch from "@radix-ui/react-switch";
+
+export interface SwitchProps {
+  state: boolean;
+  switchCallback?: (state: boolean) => void;
```
MythicManiac, line 7: This should probably be called something along the lines of `onChange` or whatever is the most standard. Really, the props here should probably emulate the props that exist in the normal checkbox input in React so that this component is easy to use in forms if necessary.

MythicManiac (changes requested):

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <div>
+        <Switch {...args} />
+      </div>
+    </div>
+  );
+};
+
+const SameStateSwitchTemplate: ComponentStory<typeof Switch> = () => {
+  const [state, setState] = useState(false);
+  const switchCallback = (x: boolean) => {
```
anttimaki, line 34: Seems like a rather pointless wrapper. Why not just use `setState` as the prop?

MythicManiac (commented):
Also why is the fontawesome setup in this PR? Doesn't seem like it even uses it

packages/cyberstorm/src/components/Switch.tsx
anttimaki, line 7: Related to this earlier comment:
> Seems like a rather pointless wrapper. Why not just use setState as the prop?

the callback could then use typings imported from `React`: `Dispatch<SetStateAction<boolean>>`. Unless there's some idea that we want the toggle to do something else than actually toggle a state 🤷🏻

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+  disabled?: boolean;
+}
+
+export const Switch: React.FC<PropsWithChildren<SwitchProps>> = ({
+  state,
+  switchCallback = undefined,
```
anttimaki, line 13: Passing undefined as the default value is pointless in JS since missing keys in objects are undefined, and missing function parameters are undefined.

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+  disabled = false,
+}) => {
+  const onCheckedChange = useCallback(() => {
```
anttimaki, line 16: I'm not sure I see the point of using `useCallBack` here, since the memoization probably adds more overhead than we save with it?

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+    if (switchCallback) {
```
anttimaki, line 17: With "optional chaining" this can be shortened to `switchCallback?.(!state);`

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+      switchCallback(!state);
+    }
+  }, [state, switchCallback]);
+
+  return (
+    <RadixSwitch.Root
+      className={styles.Switch}
+      disabled={disabled}
+      onCheckedChange={onCheckedChange}
```
anttimaki, line 26: Depending on which of my comments you end up following, this could be either
`onCheckedChange={() => switchCallback?.(!state)}`
or
`onCheckedChange={() => switchCallback?.((x) => !x)}`

packages/cyberstorm/src/components/Switch.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+  const onCheckedChange = useCallback(() => {
+    if (switchCallback) {
+      switchCallback(!state);
+    }
+  }, [state, switchCallback]);
+
+  return (
+    <RadixSwitch.Root
+      className={styles.Switch}
```
anttimaki, line 24: According to our naming conventions this should be `styles.root`, and AFAIK the keys should be lowercase here and elsewhere?

packages/cyberstorm/src/components/componentStyles/Switch.module.css
anttimaki, line 23: wat

anttimaki (changes requested):
Nothing too major this time but given the large number of small suggestions I'd set this as "Request changes". Feel free to argument otherwise.

apps/cyberstorm-storybook/package.json
Oksamies: I'm not sure at this point. Storybook doesn't work without this addition, for some reason. Although this doesn't belong in this PR. I'll make separate PR out of all the package version changes.

packages/cyberstorm/src/components/Switch.tsx
Oksamies: No. Just realized that also.

apps/cyberstorm-storybook/package.json
Oksamies: Yeah, I'm not sure either. I'll remove this and check all the package versions relevant to my components in another separate PR.

packages/cyberstorm/package.json
Oksamies: https://github.com/thunderstore-io/thunderstore-ui/pull/591#discussion_r1054030900 Going to be handled within this checking I'm doing.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
Oksamies: Very true, it was pointless. Still not fully familiar, how data is passed to `onChange` function usually. (Radix passes the checked/unchecked bool in the `onCheckedChange`)

packages/cyberstorm/src/components/Switch.tsx
Oksamies: There was no idea 😅 I've updated the typing 👍

packages/cyberstorm/src/components/Switch.tsx
Oksamies: IIRC there is an performance benefit, but then again I'm not 100% sure how the `useCallback` works. I've now left it out, in the name of simplifying code and me not knowing for sure.

apps/cyberstorm-storybook/package.json
Oksamies: Though I'll leave this here for now and try to get this same change merged into master before this.

anttimaki: Seems it's still there? And yarn.lock wasn't updated in the latest commit?

packages/cyberstorm/src/components/Switch.tsx
anttimaki: Recommended reading: https://kentcdodds.com/blog/usememo-and-usecallback

anttimaki (commented):
Apart from `package.json` stuff my concerns have been addressed.

gkurck (changes requested):
Visually looks good!

As discussed in voice, I commented a couple of additions as to conform to the designs.

PR #592, opened 2022-12-20
Refactor the Breadcrumbs component

MythicManiac:
Significantly simplify the developer facing Breadcrumbs component API as
well as implementation

Also improve the storybook stories a bit

Refs TS-1316


packages/cyberstorm/src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import styles from "./componentStyles/Breadcrumb.module.css";
+import { IndexLink } from "./Links";
+
+type BreadcrumbsProps = PropsWithChildren<{
+  excludeHome?: boolean;
+}>;
+
+export const Breadcrumbs: React.FC<BreadcrumbsProps> = (props) => {
+  let nodes: ReactNode[] = !props.excludeHome ? [<DefaultHomeCrumb />] : [];
```
Oksamies, line 13: Linter throwing this, on line 13 and 16. Though according to info here https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-key.md , I'm not sure if this can be fixed and should just be disabled on these lines.
image

Oksamies (approved):
Everything looks good to me, except the one linter throw on `Breadcrumbs.tsx` lines 13 and 16. Though it doesn't seem like a blocker to me, so I'm going approve this :shipit:

packages/cyberstorm/src/components/Breadcrumbs.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <div className={styles.root}>
+      {nodes.map((x, i) => (
+        <Breadcrumb key={i}>{x}</Breadcrumb>
+      ))}
+    </div>
+  );
+};
+
+export const Breadcrumb: React.FC<PropsWithChildren<{}>> = ({ children }) => {
```
anttimaki, line 27: Also linting complains on this line about using `{}` as a type.

anttimaki (commented):
Generally speaking, seems there's been a paradigm shift when moving from the old components to cyberstorm components. Whereas the old ones were more tightly defined the new ones use mostly all-optional props and just wraps stuff like breadcrumbs here. Not saying one approach is better than the other - it seems to work nicely for breadcrumbs. Just wondering if the decision has been intentional or if we're just drifter there?

Additionally:

* Styles don't seem to work properly on my local Storybook; links use some default serif font
* Wondering why you decided to remove the WithoutHome story?

packages/cyberstorm/src/components/Breadcrumbs.tsx
MythicManiac: This was copied from what the react type definitions do themselves in whatever version the type arg has a default value. We don't seem to have that version in use here, so just replicated how they had it done.

As far as the lint rule goes, we probably want to look into the reasoning of that, or at least I'm not fully confident yet it makes sense to avoid this syntax in every scenario possible. The alternatives are pretty verbose (unless you can use `unknown` directly). Using `any` broke the inferred type hint for the `children` variable here, would assume `unknown` to do the same, so you're left with something like `Record<string, any>` or such.

MythicManiac:
> Styles don't seem to work properly on my local Storybook; links use some default serif font

This is what it was like for me on the branch already, leaving that up to @Oksamies to deal with

> Wondering why you decided to remove the WithoutHome story?

Since it's now a documented parameter + toggleable on the story directly via a control, it felt redundant. The code example updates to reflect the toggle state. IMO the story page is a bit noisy as it is already.

packages/cyberstorm/src/components/Breadcrumbs.tsx
MythicManiac: AFAIK this probably does not have to be fixed, unless React does some weird stuff with them before even rendering. The key prop is used for React to map changes to existing elements in the DOM once they're mounted, and it's only required if there's more than one child in an element. If you look at how this component is rendered, you'll notice that the keys are set accordingly and this element never needs one since it gets wrapped by the Breadcrumb component.

packages/cyberstorm/src/components/Breadcrumbs.tsx
anttimaki: https://github.com/typescript-eslint/typescript-eslint/issues/2063

anttimaki: So I guess this *might* fall into "React Props" use case. 🤷🏻

MythicManiac: I'm assuming you mean this one here? Might be worth looking into

> ## If you are using React, and you want to define type Props = {}.
> This is technically safe in this instance, [because under the hood the {} type is passed into an intersection type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6fd37a55773b23e00a19418d9b5aad912087c982/types/react/index.d.ts#L501) (see the note at the end of this comment for why this is safe).

>
> However, there is no way for us to statically analyse and know that this is a safe usage.
 To work around this, consider reconfiguring the lint rule to match your repository's coding style.
 You can use the following config to allow it:

>
> ```json
> {
>   "rules": {
>     "@typescript-eslint/ban-types": [
>       "error",
>       {
>         "extendDefaults": true,
>         "types": {
>           "{}": false
>         }
>       }
>     ]
>   }
> }

> ```
> Consider using [an eslint overrides config](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) to limit the scope of this change to just react component files, to help ensure you're keeping your codebase as safe as possible.

PR #574, opened 2022-11-26
Add TextInput component

Oksamies:
refs TS-1317

PR #599, opened 2023-01-03
Cyberstorm DropDown UI Component

VilppeRiskidev:


packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  defaultOpen: boolean;
+  dropDownData?: DataItem[];
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultValue?: number;
+};
+
+type DataItem = {
+  label: string;
+  value: number;
+  reactElement?: ReactElement;
```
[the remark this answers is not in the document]
anttimaki: Additionally, based on how this is used in the stories, and the fact that we're trying to pass `colorScheme` props to this component in `mapDropDownData`, should the actual type be `typeof DropDownItem`? That would limit the what can be passed into the dropdown, but I don't see that as a bad thing, especially since `DropDownItem` is already quite flexible in what it can do.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+    primary: styles.content__primary,
+  }[scheme];
+};
+
+const mapDropDownData = (
+  dropDownData: DataItem[] | undefined,
+  colorScheme: "default" | "defaultDark" | "primary" | undefined,
+  setValue: React.Dispatch<React.SetStateAction<DataItem | null>>
+) => {
+  return dropDownData
```
anttimaki, line 94: Assuming the `reactElement` is typed as `typeof DropDownItem`, we could utilise different things to clean this up into:

```typescript
  const props = { colorScheme } as DropDownItemProps;

  return (dropDownData ?? []).map((dataItem, index) => (
    <DropdownMenu.Item onSelect={() => setValue(dataItem)} key={index} asChild>
      {React.isValidElement(dataItem.reactElement)
        ? React.cloneElement(dataItem.reactElement, props)
        : null}
    </DropdownMenu.Item>
  ));
```

As a plus I think it would have a better type safety. But use your own judgement if any or all of the changes make sense.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [value, setValue] = useState(initialValue);
+  const dropDownItemElements = mapDropDownData(
+    dropDownData,
+    colorScheme,
+    setValue
+  );
+
+  return (
+    <div className={styles.root}>
+      <DropdownMenu.Root defaultOpen={defaultOpen}>
```
anttimaki, line 40: Should we check that if `dropDownData` is `undefined` or an empty array, the dropdown can't be opened (by `defaultOpen` or by the user manually)?

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <DropdownMenu.Trigger asChild>
+          <div>
+            <Button
+              colorScheme={colorScheme}
+              rightIcon={icon}
+              label={value ? value.label : ""}
```
anttimaki, line 46: `{value?.label ?? ""}`

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { ReactElement, ReactNode, useState } from "react";
+import styles from "./componentStyles/DropDown.module.css";
+import { Button } from "./Button";
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+
+type DropDownProps = {
+  label?: string;
```
anttimaki, line 10: Is this used anywhere?

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+          </DropdownMenu.Content>
+        </DropdownMenu.Portal>
+      </DropdownMenu.Root>
+    </div>
+  );
+};
+
+DropDown.defaultProps = {
+  dropDownData: [],
+  defaultOpen: false,
```
anttimaki, line 69: Required prop, no need for a default.

packages/cyberstorm/src/components/DropDownItem.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { ReactNode } from "react";
+import styles from "./componentStyles/DropDownItem.module.css";
+
+export interface DropDownItemProps {
+  label?: string;
+  leftIcon?: ReactNode;
+  rightIcon?: ReactNode;
+  colorScheme?: "default" | "defaultDark" | "primary";
```
anttimaki, line 8: Since `DropDown` attempts to pass its own `colorScheme` to the options, it could be a good idea to define the color schemes in one of the components and import into the other. Since my other suggestions already require importing stuff from here to `DropDown`, it would make sense to define the schemes here (assuming my suggestions are actually implemented).

packages/cyberstorm/src/components/componentStyles/Button.module.css
```diff
@@ -22,7 +22,7 @@
 /* Styles for Button's text */
 .label {
   font: var(--font-heading);
-  font-weight: var(--font-weight-semi-bold);
+  font-weight: var(--font-weight-600);
```
anttimaki, line 25: Is this (and the other font-weight changes in other files) here on purpose? They seem unrelated to the PR. If they're here on purpose, please make sure what we want to actually use the numbers instead of text-based names (we might want to, I can't keep up with this stuff 😅 )

packages/cyberstorm/src/components/componentStyles/DropDown.module.css
```diff
@@ -0,0 +1,40 @@
+/* Common Button styles */
```
anttimaki, line 1: Not a button.

packages/cyberstorm/src/components/componentStyles/DropDown.module.css
```diff
@@ -0,0 +1,40 @@
...
+.root {
+  position: relative;
```
anttimaki, line 3: What is this needed for? There doesn't seem to be any children using `position`.

packages/cyberstorm/src/components/componentStyles/DropDownItem.module.css
```diff
@@ -0,0 +1,48 @@
...
+
+  gap: var(--gap-m);
+  min-width: 10rem;
+  padding: var(--padding-3xs) var(--padding-xs);
+  border-radius: var(--border-radius-3xs);
+  overflow: hidden;
+  color: var(--text-color);
+
+  font: var(--font-body);
+  font-size: 14px;
```
anttimaki, line 13: Is this intentionally hard coded?

anttimaki (changes requested):
Lots of comments but I think most of them are about code style or typescript improvements. Overall the approach looks good!

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+  faArrowDownAZ,
+  faArrowUpAZ,
+} from "@fortawesome/free-solid-svg-icons";
+
+const meta = {
+  title: "Cyberstorm/Components/DropDown",
+  component: DropDown,
+} as unknown as ComponentMeta<typeof DropDown>;
+
+const style = {};
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed in the next commit 👍

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+const defaultArgs = {
+  icon: (
+    <FontAwesomeIcon
+      fixedWidth={true}
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed in the next commit 👍

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+  ),
+};
+
+const dropDownData = [
+  {
+    label: "Newest",
+    value: 1,
+    reactElement: (
+      <DropDownItem
+        label={"New"}
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed in the next commit 👍

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+            className={"dropDownItemIcon"}
+          />
+        }
+      />
+    ),
+  },
+];
+
+const Template: ComponentStory<typeof DropDown> = (args) => (
+  <div style={style}>
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed in the next commit 👍

apps/cyberstorm-storybook/stories/components/DropDownItem.stories.tsx
```diff
@@ -0,0 +1,74 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
```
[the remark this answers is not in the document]
VilppeRiskidev: Fixed in the next commit 👍

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+
+type DropDownProps = {
+  label?: string;
+  icon?: ReactNode;
+  defaultOpen: boolean;
+  dropDownData?: DataItem[];
```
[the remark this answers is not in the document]
VilppeRiskidev: Yeah! Good suggestion, renaming in the next commit

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+type DataItem = {
+  label: string;
+  value: number;
+  reactElement?: ReactElement;
+};
+
+export const DropDown: React.FC<DropDownProps> = (props) => {
+  const { icon, colorScheme, defaultOpen, dropDownData, defaultValue } = props;
+
+  let initialValue = dropDownData ? dropDownData[0] : null;
```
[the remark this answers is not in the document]
VilppeRiskidev: Refactored in the next commit

packages/cyberstorm/src/components/DropDown.tsx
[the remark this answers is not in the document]
VilppeRiskidev: Makes sense. We can extend the DropDownItem if needed,
Idea was to allow passing any elements, but we can probably add a global "CyberstormComponent" type (which'd take colorScheme as a prop) in the future if needed. 
Anyway refactored, should be a little better now.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  icon?: ReactNode;
+  defaultOpen: boolean;
+  dropDownData?: DataItem[];
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultValue?: number;
+};
+
+type DataItem = {
+  label: string;
+  value: number;
```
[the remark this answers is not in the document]
VilppeRiskidev: Refactored, now takes numbers and/or strings.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+
+type DropDownProps = {
+  label?: string;
+  icon?: ReactNode;
+  defaultOpen: boolean;
+  dropDownData?: DataItem[];
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultValue?: number;
```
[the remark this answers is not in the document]
VilppeRiskidev: Renamed to defaultSelectedOptionKey.

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: @gkurck should this make the DropDown be disabled or something?

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: 👍 very nice

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: Not used, removed in the next commit 👍

packages/cyberstorm/src/components/DropDownItem.tsx
VilppeRiskidev: I'd let the DropDownitem handle it's props separately, to not depend too much on each other.
Default colorScheme is used if the passed one isn't valid.
I think in the future some list of colorSchemes might be implemented for all of the components and the shcemes would basicly be in line with each other. Feel free to change my mind.

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: Not really to do with this, but just overall a change that was nice to do. We discussed with @gkurck about using the numbers instead of xs, 3xs etc. Probably should be done separately, yeah, but this is a nice conversation starter ;D

packages/cyberstorm/src/components/componentStyles/DropDown.module.css
VilppeRiskidev: Removed 👍

packages/cyberstorm/src/components/componentStyles/DropDownItem.module.css
VilppeRiskidev: Definitely not intentional 🤮

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+  options?: DropDownOptions;
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultSelectedOptionKey?: number;
+};
+
+export type DropDownOption = {
+  label: string;
+  content?: ReactElement<DropDownItemProps>;
+};
+export type DropDownOptions = Map<number | string, DropDownOption>;
```
anttimaki, line 22: I think I prefer the earlier approach where `DropDownOption` included a `value`. Now the value is "hidden" as the key in the Map, making it sort of implicit rather than explicit. I guess this could be improved by typing this as `Map<DropDownOptionValue, DropDownOption>`, but that still feels awkward in unintuitive to *me* as a developer. I think it would also mean that we need to do (more) processing to get the values returned from the backend into a format that this component accepts. @MythicManiac could chip in with a second opinion since I'm not sure if it's just my "not invented by me" bias speaking.

To clarify, I suggest the beginning of the component should be something like this:
```typescript
export type DropDownOption = {
  value: string | number;
  content: typeof DropDownItem;
};

type DropDownProps = {
  colorScheme?: "default" | "defaultDark" | "primary";
  defaultOpen?: boolean;
  defaultValue?: string | number;
  icon?: ReactNode;
  options?: DropDownOption[];
};

export const DropDown: React.FC<DropDownProps> = (props) => {
  const { colorScheme, defaultOpen, defaultValue, icon, options } = props;

  const initialValue = options?.find((o) => o.value === defaultValue) ?? null;
  // ...rest of the component here...
```

Note that `DropDownOption.label` was yeeted since it wasn't used anymore.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <div className={styles.root}>
+      <DropdownMenu.Root defaultOpen={defaultOpen}>
+        <DropdownMenu.Trigger asChild>
+          <div>
+            <Button
+              colorScheme={colorScheme}
+              rightIcon={icon}
+              label={value?.label ?? ""}
```
anttimaki, line 52: Took a second review to connect the dots here. This shouldn't be the label of the `value` (the `DropDownItem`), but the label of the `DropDown` itself, right? The label that was removed from the `DropDown`'s as per my earlier comment since it wasn't used for anything.

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+            />
+          </div>
+        </DropdownMenu.Trigger>
+
+        <DropdownMenu.Portal>
+          <DropdownMenu.Content
+            align={"start"}
```
anttimaki, line 59: `align="start"` 🥺

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  if (
+    options &&
+    defaultSelectedOptionKey &&
+    options.get(defaultSelectedOptionKey)
+  ) {
+    initialValue = options.get(defaultSelectedOptionKey) ?? null;
+  }
+
+  const [value, setValue] = useState(initialValue);
```
anttimaki, line 39: Don't know how I missed this in the previous review, but do we actually do anything with the `value`? I.e. what happens the value changes? AFAIK currently the `DropDown`'s internal state is updated, but then what? Does the parent page/component have anyway to access the value or detect the change? Does Radix handle this somehow behind the scenes?

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const getContentStyle = (scheme: DropDownProps["colorScheme"] = "default") => {
+  return {
+    default: styles.content__default,
+    defaultDark: styles.content__defaultDark,
+    primary: styles.content__primary,
+  }[scheme];
+};
+
+const mapDropDownData = (
```
anttimaki, line 93: Iff my comments above about ditching the Map-based approach is implemented, this function also cleans up to simple `.map()`. (IIRC there was an example about it on my last review's comments.)

anttimaki (changes requested):
Just mentioning this here, Storybook shows that strong focus outline on dropdown's options on FF again. I remember this was discussed last year with some other component, but can't recall if we decided to do something about it.

packages/cyberstorm/src/components/componentStyles/Button.module.css
anttimaki: Not sure if your comment means that this should wait gkurck's input so I'll leave this comment unresolved for now. Mark resolved at your discretion.

anttimaki:
> Just mentioning this here, Storybook shows that strong focus outline on dropdown's options on FF again. I remember this was discussed last year with some other component, but can't recall if we decided to do something about it.

Well seems the other component was actually one of Oksamies, see https://github.com/thunderstore-io/thunderstore-ui/pull/574#issuecomment-1371123460

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+} from "@fortawesome/free-solid-svg-icons";
+
+const meta = {
+  title: "Cyberstorm/Components/DropDown",
+  component: DropDown,
+} as unknown as ComponentMeta<typeof DropDown>;
+
+const defaultArgs = {
+  icon: (
+    <FontAwesomeIcon fixedWidth icon={faChevronDown} className="dropDownIcon" />
```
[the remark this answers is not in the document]
VilppeRiskidev: @gkurck asked to add them, but we might've (after that) decided to not use placeholder classes 🤔

apps/cyberstorm-storybook/stories/components/DropDownItem.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+const meta = {
+  title: "Cyberstorm/Components/DropDownItem",
+  component: DropDownItem,
+} as ComponentMeta<typeof DropDownItem>;
+
+const defaultArgs = {
+  label: "A-Z",
+  leftIcon: (
+    <FontAwesomeIcon
+      fixedWidth={true}
```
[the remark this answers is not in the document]
VilppeRiskidev: 👍

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+    </div>
+  );
+};
+
+DropDown.defaultProps = {
+  defaultOpen: false,
+  colorScheme: "default",
+  icon: (
+    <FontAwesomeIcon
+      fixedWidth={true}
```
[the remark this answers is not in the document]
VilppeRiskidev: fixed all of them now (also in other files)

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+import { DropDownItemProps } from "./DropDownItem";
+
+type DropDownProps = {
+  icon?: ReactNode;
+  defaultOpen?: boolean;
+  options?: DropDownOptions;
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultSelectedOptionKey?: number;
```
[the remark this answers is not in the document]
VilppeRiskidev: 👍

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: image
This label is this one^, so it must change when the selected value changes. Also this is not the same as dropdownitem's title 
image
 (new vs. newest)

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: 👍

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: Not yet used, but when the component is used on a form etc. the setValue could instead be a callback function..... anyways I don't quite know what's the best way for managing the state.
The DropDown component has to update (label on button changes) based on the selected value and the parent component should also be aware of the selected value (state).

VilppeRiskidev: I mean I don't know if it is good to implement the state management on the storybook side, but also that would be the logical place to manage it.

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Button } from "./Button";
+import { CustomSelectItem, SelectItemProps } from "./CustomSelectItem";
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import * as Select from "@radix-ui/react-select";
+import { SelectItem } from "@radix-ui/react-select";
+
+export type SelectOption = {
+  value: string;
```
anttimaki, line 12: This might be asked-and-answered, but would `string | number` make more sense? If changed, do change `SelectProps.value` too.

packages/cyberstorm/package.json
```diff
@@ -18,6 +18,8 @@
     "@fortawesome/fontawesome-svg-core": "^6.2.0",
     "@fortawesome/free-solid-svg-icons": "^6.2.0",
     "@fortawesome/react-fontawesome": "^0.2.0",
+    "@radix-ui/react-dropdown-menu": "^2.0.1",
+    "@radix-ui/react-select": "^1.1.2",
```
anttimaki, line 22: Running `yarn.install` seems to alter `yarn.lock` on my machine. Did you remember to update `yarn.lock` in this PR after react-select was added?

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+  content: typeof CustomSelectItem;
+  displayValue?: string;
+};
+
+type SelectProps = {
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultOpen?: boolean;
+  icon?: ReactNode;
+  options?: SelectOption[];
+  setValue: CallableFunction;
```
anttimaki, line 22: I think it would make more sense name the prop as `onChange`, since that makes more sense to the developer using the component. Unless there's already an establlished practice in Cyberstorm to name the props as setFoo.

I would prefer `Dispatch<SetStateAction<string>>;` (or `string|number` if numbers are supported) since it's more specific and conveys the intent more clearly to the developer. Unless there's some specific use case that demands using the lenient `CallableFunction`.

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+    ? mapSelectData(options, colorScheme)
+    : null;
+
+  return (
+    <div className={styles.root}>
+      <Select.Root
+        defaultOpen={defaultOpen}
+        value={value}
+        onValueChange={(newValue) =>
+          setValue !== undefined ? setValue(newValue) : null
```
anttimaki, line 39: Per the current typing, `setValue` can't be undefined, so `onValueChange={setValue}` should probably suffice.

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+        }
+      >
+        <Select.Trigger asChild>
+          <div>
+            <Button
+              colorScheme={colorScheme}
+              rightIcon={icon}
+              label={
+                options?.find((o) => o.value === value)?.displayValue ?? ""
```
anttimaki, line 48: The default value could be "Choose" oslt 🤷🏻 ?

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+              {selectItemElements}
+            </Select.Group>
+          </Select.Content>
+        </Select.Portal>
+      </Select.Root>
+    </div>
+  );
+};
+
+CustomSelect.defaultProps = {
```
anttimaki, line 68: Should this include `options: [],`

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const mapSelectData = (
+  options: SelectOption[],
+  colorScheme: "default" | "defaultDark" | "primary" | undefined
+) => {
+  const props = { colorScheme } as SelectItemProps;
+
+  return options.map((option, index) => {
+    return (
```
anttimaki, line 91: This return could be omitted, as could the `div` inside since CustomSelectOption in itself is already a div.

packages/cyberstorm/src/components/DropDown.tsx
anttimaki: Yes I think it's a good idea to implement the management outside the component (e.g. in Storybook) unless we have a reason to do it otherwise.

anttimaki (changes requested):
I guess I was pretty focused on the code in front of me to fail to see that some of my comments were due to the component being kinda like select and dropdown at the same time. The change to two separate components brought some new issues though.

As discussed in a meeting this week, I add comments quite liberally. Doesn't mean all of them need to be fulfilled. If high-work-low-impact changes is blocking the progress, feel free to ignore them (or create a new ticket about them). Of course if a requested change makes sense and is quick to implement, IMO it makes sense to fix them now so we don't have to worry about them later.

In this case I think the typing of `SelectOption.content` could be seen as a blocker, though.

apps/cyberstorm-storybook/stories/components/CustomSelect.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+          />
+        }
+      />
+    ),
+  },
+];
+
+const Template: ComponentStory<typeof CustomSelect> = (args) => {
+  const [value, setValue] = useState(
+    args.defaultValue ??
```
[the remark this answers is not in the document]
VilppeRiskidev: This is going to be redone when building the website (as storybook acts as an example of usage).

It currently is used to show an example of how to set a default value for the Select.

`const DarkSelect = Template.bind({});
DarkSelect.args = {
  ...defaultArgs,
  defaultValue: "3",
  options: options,
  colorScheme: "defaultDark",
};`

So yeah, not a prop in the component, but just an argument passed when making the story.
I'll add a delete to not pass it to the component.

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Button } from "./Button";
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import * as DropdownMenu from "@radix-ui/react-dropdown-menu";
+import { SelectItemProps } from "./CustomSelectItem";
+
+type DropDownProps = {
+  colorScheme?: "default" | "defaultDark" | "primary";
+  defaultOpen?: boolean;
```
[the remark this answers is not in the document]
VilppeRiskidev: Radix's dropdown has a defaultOpen (as do many other libraries). I think it's nice to have. 🤷

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,77 @@
...
+const parseContent = (
+  content: ReactNode[] | undefined,
+  colorScheme: "default" | "defaultDark" | "primary" | undefined
+) => {
+  const props = { colorScheme } as SelectItemProps;
+  return content?.map((item, index) => {
+    return (
+      <DropdownMenu.Item key={index} asChild>
+        <div>
+          {React.isValidElement(item) ? React.cloneElement(item, props) : null}
```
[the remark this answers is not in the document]
VilppeRiskidev: yeah, no. 2 makes more sense. In my opinion in most cases it's better to allow doing "stupid" things. As a developer I'd kind of assume passing the colorScheme would not work if not using cyberstorm components.

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,77 @@
...
+  label: "...",
+};
+
+const parseContent = (
+  content: ReactNode[] | undefined,
+  colorScheme: "default" | "defaultDark" | "primary" | undefined
+) => {
+  const props = { colorScheme } as SelectItemProps;
+  return content?.map((item, index) => {
+    return (
```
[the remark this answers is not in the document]
VilppeRiskidev: 👍

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: It would, but Radix only allows strings for values, so I'll go with that.

packages/cyberstorm/package.json
VilppeRiskidev: Should be updated in the next commit

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: 👍

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: Thought about this and didn't come up with anything so I left it as blank, but "Choose" sounds good.

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: sure 👍

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: Return omitted.
`<div> `can't be omitted, Radix's `asChild` doesn't like custom components, so those have to be wrapped.

apps/cyberstorm-storybook/stories/components/CustomSelect.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+  icon: (
+    <FontAwesomeIcon fixedWidth icon={faChevronDown} className="selectIcon" />
+  ),
+};
+
+const options: SelectOption[] = [
+  {
+    value: "1",
+    displayValue: "Newest",
+    content: (
```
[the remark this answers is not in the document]
VilppeRiskidev: changed the type to `ReactElement<SelectItemProps>`

apps/cyberstorm-storybook/stories/components/Button.stories.tsx
```diff
@@ -12,11 +12,7 @@ const meta = {
 const defaultArgs = {
   label: "Categories",
   rightIcon: (
-    <FontAwesomeIcon
-      fixedWidth={true}
-      icon={faChevronDown}
-      className={"buttonIcon"}
-    />
+    <FontAwesomeIcon fixedWidth icon={faChevronDown} className={"buttonIcon"} />
```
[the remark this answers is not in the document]
VilppeRiskidev: Did exactly this 👍

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: 👍

packages/cyberstorm/src/components/componentStyles/Button.module.css
VilppeRiskidev: This has been written down and will be handled separately, I'll leave this in, as it will be gone over and changed back if needed.

apps/cyberstorm-storybook/stories/components/DropDown.stories.tsx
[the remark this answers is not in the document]
VilppeRiskidev: This has been written down and will be handled separately.

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const CustomDropDown: React.FC<DropDownProps> = (props) => {
+  const { colorScheme, defaultOpen, icon, content, label } = props;
+
+  return (
+    <div className={styles.root}>
+      <DropdownMenu.Root defaultOpen={defaultOpen}>
+        <DropdownMenu.Trigger asChild>
+          <div>
+            <Button label={label} colorScheme={colorScheme} rightIcon={icon} />
```
gkurck, line 26: I'm not keen on hardcoding a `Button` as the trigger component as the dropdown can be applied to a variety of elements. Would it be feasible to be able to provide the trigger element independently as a consumer?

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,75 @@
...
+          </div>
+        </DropdownMenu.Trigger>
+
+        <DropdownMenu.Portal>
+          <DropdownMenu.Content
+            align="start"
+            sideOffset={8}
+            className={`${styles.content} ${getContentStyle(colorScheme)}`}
+          >
+            <DropdownMenu.Group className={styles.group}>
```
gkurck, line 36: Is nesting everything inside a group by default even if the items don't belong in a group by nature intentional? This will also prevent any additional grouping of items.

packages/cyberstorm/src/components/CustomSelect.tsx
```diff
@@ -0,0 +1,108 @@
...
+              }
+            />
+          </div>
+        </Select.Trigger>
+
+        <Select.Portal>
+          <Select.Content
+            className={`${styles.content} ${getContentStyle(colorScheme)}`}
+          >
+            <Select.Group className={styles.group}>
```
gkurck, line 66: Same thing here, is nesting everything inside a group by default even if the items don't belong in a group by nature intentional? This will also prevent any additional grouping of items.

packages/cyberstorm/src/components/CustomDropDown.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export const CustomDropDown: React.FC<DropDownProps> = (props) => {
+  const { colorScheme, defaultOpen, icon, content, label } = props;
+
+  return (
+    <div className={styles.root}>
+      <DropdownMenu.Root defaultOpen={defaultOpen}>
+        <DropdownMenu.Trigger asChild>
+          <div>
```
gkurck, line 25: [Ref comment](https://github.com/thunderstore-io/thunderstore-ui/pull/599#discussion_r1066863646)

> `<div>` can't be omitted, Radix's asChild doesn't like custom components, so those have to be wrapped.

[This issue on Radix GitHub](https://github.com/radix-ui/primitives/issues/953) tells that the components used with `asChild` need to spread DOM props to work. Is this something we should and can do?

This applies to all components built with Radix, including `CustomSelect` in this PR.

packages/cyberstorm/src/components/componentStyles/CustomSelectItem.module.css
```diff
@@ -0,0 +1,50 @@
...
+  gap: var(--gap-m);
+  min-width: 10rem;
+  padding: var(--padding-3xs) var(--padding-xs);
+  border-radius: var(--border-radius-3xs);
+  overflow: hidden;
+  color: var(--text-color);
+
+  font: var(--font-body);
+  font-size: var(--text-s);
+  background-color: var(--bg-color);
```
gkurck, line 14: I don't think this should have `background-color` set by default as it should let the `background-color` of the parent container through. It also creates a visual z-index problem with the styling `reset.css` applies to `:focus-visible` -state.

packages/cyberstorm/src/components/componentStyles/CustomSelectItem.module.css
```diff
@@ -0,0 +1,50 @@
...
+  cursor: pointer;
+
+  --text-color: var(--color-white);
+  --bg-color: var(--color-surface--3);
+}
+
+.label {
+  font-weight: var(--font-weight-600);
```
gkurck, line 22: Should be `500` instead as per the design.

packages/cyberstorm/src/components/componentStyles/CustomSelectItem.module.css
```diff
@@ -0,0 +1,50 @@
+.root {
+  display: flex;
+  flex-direction: row;
+
+  gap: var(--gap-m);
+  min-width: 10rem;
+  padding: var(--padding-3xs) var(--padding-xs);
```
gkurck, line 7: I'd change the vertical padding to `--padding-2xs` to be more in line with the designs.

gkurck (changes requested):
Added some css and radix usage -related comments, one of which might solve the need for `<div>` when using components with `asChild`.

packages/cyberstorm/src/components/CustomDropDown.tsx
VilppeRiskidev: I think it can be done, but I don't think we should do it right now, Let's write this down for future work. Does the design currently have any dropdowns that don't use button as the trigger?

packages/cyberstorm/src/components/CustomDropDown.tsx
VilppeRiskidev: True, removed the unnecessary layer of grouping.

packages/cyberstorm/src/components/CustomDropDown.tsx
gkurck: I don't think any of them currently use a button. Looking at the header, it has two instances of a dropdown, one with the link `Developers` and one with the avatar.

packages/cyberstorm/src/components/componentStyles/CustomSelectItem.module.css
```diff
@@ -0,0 +1,50 @@
+.root {
```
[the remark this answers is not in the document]
VilppeRiskidev: Added classes and definitions

packages/cyberstorm/src/components/CustomDropDown.tsx
VilppeRiskidev: Got it this time, should be OK now.

packages/cyberstorm/src/components/CustomSelect.tsx
gkurck: How about `Select`? :)

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -19,22 +19,27 @@ export interface ButtonProps {
 /**
  * Cyberstorm Button component
  */
-export const Button: React.FC<ButtonProps> = (props) => {
-  const { label, leftIcon, rightIcon, colorScheme, onClick } = props;
+export const Button: React.FC<ButtonProps> = React.forwardRef(
```
anttimaki, line 22: IIRC it was discussed that ref forwarding should be reserved for very specific conditions? Now it's hard coded to all Cyberstorm buttons?

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,66 @@
...
+  content?: ReactNode[];
+  trigger: ReactNode;
+};
+
+export const DropDown: React.FC<DropDownProps> = React.forwardRef(
+  (props, ref) => {
+    const { colorScheme, defaultOpen, content, trigger } = props;
+
+    return (
+      <div className={styles.root} {...props} ref={ref}>
```
anttimaki, line 19: On my IDE TypeScript complains about `ref`. Based on quick search this should use `React.forwardRef<HTMLDivElement, DropDownProps>`, but I'm not 100% certain about that so maybe do some proper research into the subject.

Also my Monday-morning brain can't quite recall why we're dealing with the refs in the first place? Anyhow settings the ref to the outermost wrapper seems odd, since that should be accessible by parents anyway? Isn't the point of the ref to attach it to some leaf component that can't be otherwise accessed from the outside? Should Storybook contain some example where the ref is actually used in a way that provides some value?

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,66 @@
...
+DropDown.defaultProps = {
+  defaultOpen: false,
+  colorScheme: "default",
+};
+
+const parseContent = (
+  content: ReactNode[] | undefined,
+  colorScheme: "default" | "defaultDark" | "primary" | undefined
+) => {
+  const props = { colorScheme } as SelectItemProps;
```
anttimaki, line 52: Since content is now just `ReactNode`s, this row seems pointless. It could be inlined on L55.

packages/cyberstorm/src/components/MetaItem.tsx
```diff
@@ -12,7 +12,7 @@ export interface MetaItemProps {
  * Used for displaying a single data point (e.g. an amount
  * of likes or a size of a package) with an icon next to it
  */
-export const MetaItem: React.FC<MetaItemProps> = (props) => {
+export const MetaItem: React.FC<MetaItemProps> = React.forwardRef((props) => {
```
anttimaki, line 15: I think we're going a bit overboard with the refs here (or maybe I'm missing something)? Maybe hit me up on Discord? Anyways the ref isn't even used on this component.

packages/cyberstorm/src/components/Select.tsx
```diff
@@ -0,0 +1,100 @@
...
+    ? mapSelectData(options, colorScheme)
+    : null;
+
+  return (
+    <div {...props} ref={ref} className={styles.root}>
+      <RadixSelect.Root
+        defaultOpen={defaultOpen}
+        value={value}
+        onValueChange={onChange}
+        disabled={options.length == 0}
```
anttimaki, line 47: Prefer `===`

packages/cyberstorm/src/components/SelectItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+  (props, ref) => {
+    const { label, leftIcon, rightIcon, colorScheme } = props;
+
+    return (
+      <div
+        {...props}
+        ref={ref}
+        className={`${styles.root} ${getStyle(colorScheme)}`}
+      >
+        <div className={`${styles.icon} ${getStyle(colorScheme)}`}>
```
anttimaki, line 24: Are the divs rendered on purpose even if left/rightIcon are undefined?

anttimaki (changes requested):
For future reference, when I sugggest some stylistic changes (like getting rid of `={"`), I mean it only for the files already in the scope of the PR, not any old occurances. Don't get me wrong, I love that they get fixed, but they should be done in a separate PR.

apps/cyberstorm-storybook/stories/components/CustomSelect.stories.tsx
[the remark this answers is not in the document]
anttimaki: Doesn't look like it? Is there a specific reason why you went with `ReactNode` instead?

packages/cyberstorm/src/components/CustomSelect.tsx
anttimaki: Leaving unresolved due to gkurck's comment. The default value is currently "Choose" as I suggested.

packages/cyberstorm/src/components/CustomDropDown.tsx
anttimaki: Also looking at the `Developers` example, the trigger and the drop down list use different color schemes, which isn't supported by the current implementation.

anttimaki:
image
Rementioning this just in case, stuff looks wack in FF.

Also the primary SelectItem story is kinda invisible unless hovered, I assume that's not on purpose?

apps/cyberstorm-storybook/stories/components/CustomSelect.stories.tsx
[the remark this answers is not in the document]
VilppeRiskidev: This seems outdated, the SelectOption is now more simple.

packages/cyberstorm/src/components/CustomDropDown.tsx
VilppeRiskidev: Adding the support in the next commit

apps/cyberstorm-storybook/stories/components/CustomDropDown.stories.tsx
```diff
@@ -0,0 +1,107 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
```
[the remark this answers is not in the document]
VilppeRiskidev: oopsie, renamed...

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: All components now support ref forwarding. What would be a solution if we want to only use it when needed? 2 different components? :D

packages/cyberstorm/src/components/DropDown.tsx
VilppeRiskidev: If you can figure a way to not use refs at all, I'll gladly use that. I couldn't, Radix wants the ref for Select's items, and the trigger etc.

packages/cyberstorm/src/components/CustomSelect.tsx
VilppeRiskidev: It is now `Select` as suggested.

packages/cyberstorm/src/components/SelectItem.tsx
VilppeRiskidev: Fixed

anttimaki (changes requested):
This review is done with the assumption that a review for how we should use refs in general is done separately. IIRC @MythicManiac mentioned that they might look into it?

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -19,22 +24,30 @@ export interface ButtonProps {
...
-      onClick={onClick}
-    >
-      {leftIcon}
-      {label ? <div className={styles.label}>{label}</div> : null}
-      {rightIcon}
-    </button>
-  );
-};
+    const fallbackRef = useRef(null);
+    const ref = forwardedRef || fallbackRef;
```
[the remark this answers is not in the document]
VilppeRiskidev: Writing this down for future work

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export const DropDown: React.FC<DropDownProps> = (props) => {
+  const {
+    colorScheme,
+    defaultOpen,
+    content,
+    trigger,
+    triggerColorScheme,
+    ...rest
```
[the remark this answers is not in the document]
VilppeRiskidev: Refactored into not passing other props than the colorScheme to the trigger, as that wasn't needed. Is this ok now? 🤔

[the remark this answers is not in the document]
VilppeRiskidev: > 

Actually yeah, the whole ...rest wasn't needed LOL

packages/cyberstorm/src/components/DropDown.tsx
```diff
@@ -0,0 +1,75 @@
...
+  } = props;
+  const triggerProps = Object.assign({}, rest, {
```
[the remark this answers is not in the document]
VilppeRiskidev: no longer needed

packages/cyberstorm/src/components/MenuItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+  rightIcon?: ReactNode;
+  colorScheme?: "default" | "defaultDark" | "primary";
+}
+
+/**
+ * Cyberstorm MenuItem component
+ */
+export const MenuItem: React.FC<MenuItemProps> = React.forwardRef(
+  (props, forwardedRef) => {
+    const { label, leftIcon, rightIcon, colorScheme, ...rest } = props;
```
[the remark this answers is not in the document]
VilppeRiskidev: the `...rest` here contains:
```
{
    "role": "option",
    "aria-labelledby": "radix-5",
    "aria-selected": false,
    "data-state": "unchecked",
    "tabIndex": -1,
    "data-radix-collection-item": ""
}
``` 
which is required to make the selecting work

packages/cyberstorm/src/components/MenuItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+      >
+        {leftIcon ? (
+          <div className={`${styles.icon} ${getStyle(colorScheme)}`}>
+            {leftIcon}
+          </div>
+        ) : null}
+        {label ? <div className={styles.label}>{label}</div> : null}
+        {rightIcon ? (
+          <div className={`${styles.icon} ${getStyle(colorScheme)}`}>
+            aaa
```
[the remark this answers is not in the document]
VilppeRiskidev: 😆 omg, removed

packages/cyberstorm/src/components/DropDown.tsx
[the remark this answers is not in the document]
anttimaki: Is the `...rest` needed in the other components where it was added in this PR?

packages/cyberstorm/src/components/MenuItem.tsx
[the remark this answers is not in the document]
anttimaki: And these are passed automatically by Radix?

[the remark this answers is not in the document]
VilppeRiskidev: yup, the example was copy-pasted from when you do console.log(rest); before render. (also, in the DropDown when `rest` was passed, button got disabled automatically by radix when content was empty)

packages/cyberstorm/src/components/DropDown.tsx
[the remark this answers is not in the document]
VilppeRiskidev: I don't actually know, have to dig a bit deeper. It is used for Radix, but maybe it is not needed for everything. for leaf components it doesn't cause any harm, but for others......

packages/cyberstorm/src/components/CustomDropDown.tsx
gkurck: I'm resolving this since this has been worked on and there's another, more up-to-date discussion involving this in another comment.

packages/cyberstorm/src/components/componentStyles/DropDown.module.css
```diff
@@ -0,0 +1,35 @@
+.root {
```
gkurck, line 1: Adding here as there's no relevant class to attach this to.

While removing the background-color from `MenuItem` had the effect of nullifying the visual z-index issue with the `:focus-visible` -ring, the problem still persists with any custom dropdown items for `DropDown` which have a background-color. However, this is easily fixed by setting a bigger z-index for the item being hovered. This can be targeted with the data-attribute Radix sets automatically, `[data-highlighted]`.

The question of what to do to the ring itself in the long run is another question, but for now this solutions is a good exercise on the matter.

gkurck (approved):
Visually LGTM apart from the minor issue I highlighted (no pun intented) below.

There are some visual inconsistencies to work on other external components to this PR - namely the button, but that's a known issue and a story for another time.

apps/cyberstorm-storybook/stories/components/PackageCard.stories.tsx
```diff
@@ -7,9 +7,7 @@ const meta = {
   component: PackageCard,
 } as ComponentMeta<typeof PackageCard>;
 
-const defaultArgs = {
-  packageCardStyle: "default",
-};
+const defaultArgs = {};
```
anttimaki, line 10: I don't see much point in having an empty object as the defaults.

packages/cyberstorm/src/components/PackageCard.tsx
```diff
@@ -132,6 +134,7 @@ function getPackageFlags(
   if (isPinned) {
     flagList.push(
       <PackageFlag
+        key={1}
```
anttimaki, line 137: React recommends using "meaningful" strings as keys, so here we could use the label texts. Although it might mostly be based on using indexes as keys, where change in order of things might degrade performance, and thus might not apply to manually set numbers. (But then again, if order of these things are changed manually, you'd have to change the keys too or they would be illogical.)

packages/cyberstorm/src/components/Button.tsx
```diff
@@ -19,11 +24,19 @@ export interface ButtonProps {
 /**
  * Cyberstorm Button component
  */
-export const Button: React.FC<ButtonProps> = (props) => {
-  const { label, leftIcon, rightIcon, colorScheme, onClick } = props;
+export const Button: React.FC<ButtonProps> = React.forwardRef<
+  HTMLButtonElement,
+  ButtonProps
+>((props: PropsWithChildren<ButtonProps>, forwardedRef) => {
+  const { label, leftIcon, rightIcon, colorScheme, onClick, ...rest } = props;
```
anttimaki, line 31: Continuing the `...rest` discussion: the name of the variable doesn't convey much information about the reasons it's used, as I think is proven by me needing to ask what it's for 😅 . Since the main point is for passing Radix props, and at the same time React.FC props, would something like `radixFCProps` make sense? Or `forwardedProps` to be generic?

anttimaki (approved):
Some minor stuff, I don't consider any of them blockers.

packages/cyberstorm/src/components/Button.tsx
VilppeRiskidev: I like forwardedProps 👍

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: switched to "flag_pinned" etc. also adde "category_" prefix for category tag's key

apps/cyberstorm-storybook/stories/components/PackageCard.stories.tsx
VilppeRiskidev: Removed (also removed from Tag)

anttimaki (approved):

packages/cyberstorm/src/components/componentStyles/MenuItem.module.css
```diff
@@ -18,25 +18,29 @@
   --bg-color: transparent;
 }
 
+.root[data-highlighted] {
+  z-index: 999;
+}
```
gkurck, line 23: I'm not sure if it's best to include the data attribute from Radix here in a component that doesn't rely on Radix at all and could potentially used without any involvement from Radix. Would this be better suited to reside in `DropDown` and `Select`?

packages/cyberstorm/src/components/componentStyles/MenuItem.module.css
```diff
@@ -18,25 +18,29 @@
...
+
 .label {
   font-weight: var(--font-weight-500);
   line-height: var(--line-height-l);
 }
 
 /* Different MenuItem color schemes */
-.selectItem__default:hover {
+.selectItem__default[data-highlighted] {
```
gkurck, line 31: I don't think this change is necessary, since we specifically want the style changes on hover, not when another component might or might not add a data attribute.

This applies to the changes below as well.

gkurck (commented):
I gave my opinion on the data attribute related changes to `MenuItem`. What do you think?

packages/cyberstorm/src/components/PackageCard.tsx
VilppeRiskidev: I did this fix

packages/cyberstorm/src/components/componentStyles/MenuItem.module.css
VilppeRiskidev: Reverted these.

anttimaki (approved):

packages/cyberstorm/src/components/componentStyles/MenuItem.module.css
VilppeRiskidev: Created a clickup task about this

VilppeRiskidev: Fixed

PR #603, opened 2023-01-16
Refactor global.css to use numberic variable names

VilppeRiskidev:


anttimaki:
Well technically I see no problems with the changes. But I do say that something like
```css
padding: var(--size-0-250) var(--size-0-625);
```
doesn't look developer friendly to me. That being said, if it's the direction we want to take, I won't veto it.

MythicManiac:
I feel like non-semantic variable names like this more or less eliminates the purpose of variables entirely? The only thing I can think of that it really gives is potential to override them based on context (which we probably don't need to do), and makes it easier to track that the values we have pre-defined are used as opposed to just being random values.

In this naming scheme, they aren't really variables, but just a pre-defined set of constants that we're fine with using. If we were to change some of the values, the variable naming would immediately stop making sense.

PR #606, opened 2023-01-30
Separate global css files from the cyberstorm storybook

VilppeRiskidev:


apps/cyberstorm-storybook/.storybook/preview.js
```diff
@@ -1,5 +1,6 @@
 import { LinkLibrary } from "../LinkLibrary";
 import { LinkingProvider } from "@thunderstore/cyberstorm";
+import "@thunderstore/cyberstorm-styles/dist/thunderstore-cyberstorm-styles.esm";
```
anttimaki, line 3: Would this work with just `import "@thunderstore/cyberstorm-styles";`? The user of the package shouldn't be this aware of the internal structure?

apps/cyberstorm-storybook/package.json
```diff
@@ -41,6 +41,7 @@
     "@babel/preset-env": "^7.16.11",
     "@babel/preset-typescript": "^7.16.0",
     "@thunderstore/cyberstorm": "^0.1.0",
+    "@thunderstore/cyberstorm-styles": "^0.0.0",
```
anttimaki, line 44: Nitpick: semver should start from 0.1.0.

packages/cyberstorm-styles/src/index.tsx
```diff
@@ -0,0 +1,3 @@
+import "./styles/globals.css";
```
anttimaki, line 1: The original tickets (TS-1214 & TS-1215) mention that user should be able to import globals and reset separately. Is there a feasible way to achieve this?

anttimaki (commented):
Did you manage to use this in a Next.js app, i.e. `apps/nextjs` or `apps/sertra`? On a quick try I encountered some errors, which I didn't explore further. Note: such changes should NOT be included in this PR. I'm merely interested in how easy this implementation will be to use in it's actual use case.

apps/cyberstorm-storybook/.storybook/preview.js
VilppeRiskidev: fixed 👍

apps/cyberstorm-storybook/package.json
VilppeRiskidev: Fixed 👍

PR #655, opened 2023-04-17
Cyberstorm nextjs fixings

Oksamies:


apps/cyberstorm-nextjs/next.config.js
```diff
@@ -1,6 +1,9 @@
+const withPreconstruct = require("@preconstruct/next");
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  transpilePackages: ["react"],
```
MythicManiac, line 6: Is this line actually necessary?

apps/cyberstorm-nextjs/pages/_document.tsx
```diff
@@ -1,5 +1,15 @@
...
+      <Html lang="en">
+        <Head></Head>
+        <body>
+          <Main />
+          <NextScript />
+        </body>
+      </Html>
+    );
+  }
 }
```
MythicManiac, line 15: Why is this being turned into a class based function?

MythicManiac (approved):
Overall looks good, though the build has to be fixed of course and I had a few comments/questions

apps/cyberstorm-nextjs/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "cyberstorm-nextjs",
+  "name": "@thunderstore/cyberstorm-nextjs",
```
anttimaki, line 2: 👍🏻

apps/nextjs/package.json
```diff
@@ -17,19 +17,18 @@
     "@thunderstore/cyberstorm": "^0.1.0",
     "@thunderstore/dapper": "^0.1.0",
     "@thunderstore/hooks": "^0.1.0",
-    "@types/react": "^17.0.38",
+    "@types/react": "18.0.35",
```
anttimaki, line 20: Could it maybe be enough to update React versions of `apps/cyberstorm-nextjs` and `packages/cyberstorm` (and perhaps `apps/cyberstorm-storybook` to match? Not that I'm against updating the others, but seems like here they are just updated without any proper testing or thought?

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
```diff
@@ -9,6 +9,10 @@ type BreadCrumbsProps = PropsWithChildren<{
   excludeHome?: boolean;
 }>;
 
+type BreadCrumbProps = PropsWithChildren<{
+  key?: number;
```
anttimaki, line 13: Is this React's key-prop? Should it also support strings? And should it actually be used in `BreadCrumb`?

apps/cyberstorm-nextjs/next.config.js
anttimaki: Alternative take - should it include `@thunderstore/cyberstorm`? (I just read [this note](https://nextjs.org/blog/next-13-1#built-in-module-transpilation-stable) quickly, didn't dig any deeper, so this is completely unloaded question.)

anttimaki (commented):
Felt like there weren't too many cooks yet.

apps/nextjs/package.json
MythicManiac: It's probably best to keep them in sync if we can help it, and I'm pretty sure that's also a requirement on some level given the cross-package dependencies.

apps/cyberstorm-nextjs/next.config.js
MythicManiac: I'm assuming preconstruct takes care of that with the `withPreconstruct` wrapper, and that this line was a leftover from when @Oksamies was manually trying to declare the packages here, thus the comment

apps/nextjs/package.json
Oksamies: The problem was that I couldn't have multiple versions of these packages exist at the same time. Or at least not with such a big version difference (17vs18) or it wouldn't work. I can try if it works with just what @anttimaki mentioned 🤷

MythicManiac: > I can try if it works with just what @anttimaki mentioned

no need to, better to keep them in sync as mentioned

apps/cyberstorm-nextjs/next.config.js
anttimaki: Well it seems that I've added [this line](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/nextjs/next.config.js#L2) when making the cyberstorm components work with the old nextjs app, and I assume transpilePackages does somewhat the same thing. But I want to strong emphasise that this is all assumptions.

MythicManiac: Yeah I'm just asking if @Oksamies tried to run the project without this config line, since it seems like what he originally tried to do with it didn't end up being necessary after all (based on Discord screenshots).

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
anttimaki: What I meant was that the key, if given, is not actually passed to the span on [L38](https://github.com/thunderstore-io/thunderstore-ui/pull/655/files/f34b4dd648b69d4aff7a5aaf5044711d69c0eed1#diff-389a5ab4e4f2ec027c58d319f053029ac8a24b8227c218f4689ecff407e749d0R38). So when I use this component and pass the key, it gets yeeted and I get odd errors about not providing a key, although I did provide one.

Now that I think of it, React's types probably provide something that already provides the key, and maybe we should use that instead of defining it manually. Can't recall what that type might be from the top of my mind.

apps/cyberstorm-nextjs/next.config.js
Oksamies: Ye, I tried running it without the line. It gave an weird React related error. That error lead me to an github thread, which listed this as an solution.

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
Oksamies: I think I followed what you meant 🤔 I've added the key to the span 👍

VilppeRiskidev: key is a default prop. You don't need to define it in the component's props, just pass it to the component. I fixed this.

PR #656, opened 2023-04-19
Cyberstorm copybutton

Cyanideankst:


packages/cyberstorm/src/components/CopyButton/CopyButton.module.css
```diff
@@ -0,0 +1,11 @@
+.copy {
+  color: var(--color-text--tertiary);
+}
+
+.checkmark {
+  color: var(--color-cyber-green--50);
+}
+
+.copy:hover {
```
anttimaki, line 9: It would be clearer if this would come after `.copy` instead of `.checkmark`, to have `.copy`-related stuff close to each other.

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type CopyButtonProps = {
+  text: string;
+};
+type CopiedValue = string | null;
+type CopyFn = (text: string) => Promise<void>;
+
+function useCopyToClipboard(): CopyFn {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [copiedText, setCopiedText] = useState<CopiedValue>(null);
```
anttimaki, line 17: If `copiedText` is never used (as per the eslint comment above), why set it's value at all? Seems this isn't actually used in any way?

If you yeet this `useState`, this hook just returns a function. In which case it doesn't need to be a hook? Why not just define `copy` as a regular function?

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const copy: CopyFn = async (text) => {
+    if (!navigator?.clipboard) {
+      console.warn("Clipboard not supported");
```
anttimaki, line 21: If the clipboard can't be used, maybe we should just `return;` here after logging the warning?

Although it would be more user-friendly if the `CopyButton` would also do this check, and just return/render `null` if the copy functionality isn't available. Why show a button at all if clicking it only logs some errors in the console?

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }
+  };
+
+  return copy;
+}
+
+export const CopyButton = React.forwardRef<HTMLButtonElement, CopyButtonProps>(
+  (props: PropsWithChildren<CopyButtonProps>) => {
+    const copy = useCopyToClipboard();
+    const [stateValue, setState] = useState(false);
```
anttimaki, line 40: Something like `const [wasCopied, setWasCopied] = useState(false);` could be more descriptive naming?

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    useEffect(() => {
+      setTimeout(() => setState(false), 2000);
+    }, [stateValue]);
+    return (
+      <Tooltip content="Copy">
+        <button
```
anttimaki, line 46: Would we want to use Cyberstorm's button component here, to allow styling the copy button like other buttons? I haven't checked the designs so I've no idea what we're aiming for. Just thinking aloud.

anttimaki (commented):
"Asiahan ei tietysti minulle kuulu, mutta" just some thoughts to consider.

packages/cyberstorm/src/components/Tooltip/Tooltip.tsx
```diff
@@ -24,7 +26,7 @@ export const Tooltip: React.FC<PropsWithChildren<TooltipProps>> = ({
 }) => {
   return (
     <RadixTooltip.Root>
-      <RadixTooltip.Trigger>{children}</RadixTooltip.Trigger>
+      <RadixTooltip.Trigger asChild={asChild}>{children}</RadixTooltip.Trigger>
```
VilppeRiskidev, line 29: are there some cases where asChild would actually be false? IF not maybe just hard code it and remove the prop, otherwise leave as is.

VilppeRiskidev (commented):
Aside from the one comment I left and what Antti noted, seems ok 👌

packages/cyberstorm/src/components/Tooltip/Tooltip.tsx
VilppeRiskidev: this is fixed, resolve conversation when you've looked at the commit 👌

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
VilppeRiskidev: Usually yes, that would be nice, but in this case the copybutton is its' own thing and differs from the cyberstorm Button's usage enough to be best done this way IMO.

packages/cyberstorm/src/components/CopyButton/CopyButton.tsx
Cyanideankst: Used wasRecentlyCopied here since it is bit more descriptive.

PR #688, opened 2023-05-25
Clean up MetaInfoItemList

anttimaki:
To share why I think these changes are good ones and to promote learning within the team, here's a bit more verbose description.

In PackageDetailLayout:

- metaInfoData dummy data object defined "key" prop, which would suggest that the value is returned from the backend. This is unlikely, since the keys are implementation detail required by React, and backend doesn't even (need to) know that React is used on frontend. Usually you want to generate the keys on the component that does the looping from what data you have available already.
- The dummy data object also wrapped values of "content" prop with React fragments. The real data is more likely to contain strings from the backend, and since they're valid ReactNode's too, casting them into fragments is unnecessary.

In MetaInfoItemList:

- Import MetaInfoItemProps from MetaInfoItem component to avoid needing to redefine them here. This should help keep the components in sync if someone edits MetaInfoItem component.
- Rewrote the component docstring, assuming it was copy-pasted from somewhere.
- When `const metaItemList` was defined, a wrapping div was used for the "key" prop - this was unnecessary, since we can pass the key directly to MetaInfoItem component. Also the value of key is now based on the label of the item and a running integer from the map-loop, instead of being a value passed from the backend.
- "label" and "content" props were passed to MetaInfoItem component explicitly, but I changed it to use `{...metaInfoItem}` approach instead, which passes all the props expected by the component that are found in the object used. While the previous approach 100% works and can be more informative, I like this approach since it's shorter and doesn't require updating if more props are added.
- In the markup returned from MetaInfoItemList, the root className was set into an empty div instead of the actual root div, so it was moved to root div instead. This seemed to have no effect on how the component actually looks, suggesting that it's unnecessary, but I decided to keep it to be consistent with other components.
- The changes mentioned above made the markup much shorter, so it was possible to yeet `const metaItemList` definition and do it "inline", and also remove the `return` statements, so the result is even shorter but still readable.

Also it's good to have the ref to the ticket in the commits message (too), instead of (only) in the PR comments. Like this:

Refs TS-1595


VilppeRiskidev:
using defaultProps is deprecated nowadays, so I fixed that.

Cyanideankst (approved):

PR #757, opened 2023-08-01
@thunderstore/cyberstorm: fix SSR for CopyButton

anttimaki:
Navigator is not defined on server side, so any checks on its existence needs to be done within a useEffect hook to trigger it only on client side.

Refs TS-1778


VilppeRiskidev (approved):
LGTM 👍🏻

PR #763, opened 2023-08-04
Add login page

VilppeRiskidev:


packages/cyberstorm/src/components/Layout/LoginLayout/LoginLayout.tsx
```diff
@@ -0,0 +1,123 @@
+"use client";
```
MythicManiac, line 1: This covers too much code right now

VilppeRiskidev: Nice catch

VilppeRiskidev:
The background images will be refactored separately from jpg/png into html/css format.

PR #771, opened 2023-08-16
Use Dapper calls instead of directly calling the dummy data generator

VilppeRiskidev:


MythicManiac:
Looks good so far

packages/cyberstorm/src/components/Header/Header.tsx
```diff
@@ -64,65 +66,69 @@ export function Header() {
...
+              />
+            }
           />
-        }
-      />
-    </a>,
-  ];
+        </a>,
+      ]
+    : [];
```
MythicManiac, line 131: This should be turned into a standalone component or function instead of using a massive multi-line ternary operator.

packages/dapper/src/dapper.ts
```diff
@@ -112,6 +122,9 @@ export class Dapper implements DapperInterface {
   getCommunities = NotImplemented;
   getCommunity = NotImplemented;
   getPackage = NotImplemented;
+  getPackageDependencies = NotImplemented;
+  getServiceAccount = NotImplemented;
   getTeam = NotImplemented;
   getUser = NotImplemented;
+  getUserSettings = NotImplemented;
```
MythicManiac, line 129: Is this not missing some of the newly added functions? `getServiceAccountList` and `getTeamList` were added to the interface but not here.

MythicManiac (approved):
Overall this fulfills the goal of the task, nice work.

There were/are some code quality and structure issues I wanted to comment on, but decided not to as they were out of scope of the changes here and I don't want to have this blocked unnecessarily.

After this is merged, we can start bringing the real-data dapper implementation to a state which works with the project. Although some refactoring of the dapper module might be needed first (such as splitting the different implementations to standalone modules).

PR #774, opened 2023-08-20
Add Jest, a fetch function and a test to api pkg

Oksamies:
refs TS-1799

### DIFF
--- a/packages/thunderstore-api/babel.config.js
+++ b/packages/thunderstore-api/babel.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  presets: [
+    ["@babel/preset-env", { targets: { node: "current" } }],
+    "@babel/preset-typescript",
+  ],
+};
--- a/packages/thunderstore-api/jest.config.ts
+++ b/packages/thunderstore-api/jest.config.ts
@@ -1,0 +1,201 @@
+/*
+ * For a detailed explanation regarding each configuration property and type check, visit:
+ * https://jestjs.io/docs/configuration
+ */
+
+const esModules = [
+  "node-fetch",
+  "data-uri-to-buffer",
+  "fetch-blob",
+  "formdata-polyfill",
+].join("|");
+
+export default {
+  // All imported modules in your tests should be mocked automatically
+  // automock: false,
+
+  // Stop running tests after `n` failures
+  // bail: 0,
+
+  // The directory where Jest should store its cached dependency information
+  // cacheDirectory: "C:\\Users\\askos\\AppData\\Local\\Temp\\jest",
+
+  // Automatically clear mock calls, instances and results before every test
+  clearMocks: true,
+
+  // Indicates whether the coverage information should be collected while executing the test
+  collectCoverage: true,
+
+  // An array of glob patterns indicating a set of files for which coverage information should be collected
+  // collectCoverageFrom: undefined,
+
+  // The directory where Jest should output its coverage files
+  coverageDirectory: "coverage",
+
+  // An array of regexp pattern strings used to skip coverage collection
+  // coveragePathIgnorePatterns: [
+  //   "\\\\node_modules\\\\"
+  // ],
+
+  // Indicates which provider should be used to instrument code for coverage
+  // coverageProvider: "babel",
+
+  // A list of reporter names that Jest uses when writing coverage reports
+  // coverageReporters: [
+  //   "json",
+  //   "text",
+  //   "lcov",
+  //   "clover"
+  // ],
+
+  // An object that configures minimum threshold enforcement for coverage results
+  // coverageThreshold: undefined,
+
+  // A path to a custom dependency extractor
+  // dependencyExtractor: undefined,
+
+  // Make calling deprecated APIs throw helpful error messages
+  // errorOnDeprecated: false,
+
+  // Force coverage collection from ignored files using an array of glob patterns
+  // forceCoverageMatch: [],
+
+  // A path to a module which exports an async function that is triggered once before all test suites
+  // globalSetup: undefined,
+
+  // A path to a module which exports an async function that is triggered once after all test suites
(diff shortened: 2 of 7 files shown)


Oksamies:
Although this isn't an draft PR, all of this is an suggestion since I don't have a strong understanding on how to implement.

This PR still adds `"node-fetch": "^3.3.2"`, since I couldn't figure out any other way to get fetch to work with Jest. Suggestions appreciated.

packages/thunderstore-api/babel.config.js
anttimaki, line 1: It could be nice for the next guy if e.g. commit message would mention why this is needed here. So when a year from now they wonder if it's still needed, they can at least check if the most obvious thing breaks,

packages/thunderstore-api/jest.config.ts
anttimaki, line 1: Just noting that this seems quite a different from the contents of Dapper's jest config. I'm not saying this is worse than what Dapper has - for all I know Dapper's approach can be outdated - but generally speaking if there's no reason to have two different ways of doing this setup, I'd personally prefer just one approach. (Note that this might mean updating Dapper's config instead of this package's - I'm not up to speed with what's best practice in this regard.)

packages/thunderstore-api/package.json
```diff
@@ -11,8 +11,19 @@
   ],
   "scripts": {
     "build": "tsc",
-    "dev": "tsc --watch"
+    "dev": "tsc --watch",
+    "test": "jest"
```
anttimaki, line 15: Could this also be `jest --watch`? It would be nice to have this running in the background and the tests breaking as soon as something is up, instead of at the branch pushing or commit time.

packages/thunderstore-api/package.json
```diff
@@ -11,8 +11,19 @@
...
-  "devDependencies": {}
+  "devDependencies": {
+    "@babel/core": "^7.22.10",
+    "@babel/preset-env": "^7.22.10",
+    "@babel/preset-typescript": "^7.22.5",
+    "@jest/globals": "^29.6.2",
+    "babel-jest": "^29.6.2",
+    "node-fetch": "^3.3.2",
+    "node-notifier": "^10.0.1",
+    "ts-jest": "^29.1.1",
```
anttimaki, line 26: At least this dependency uses different version than the one already in this monorepo. I'm not sure about the others. IIRC we prefer the "only one version in monorepo" approach?

packages/thunderstore-api/src/client.ts
```diff
@@ -0,0 +1,35 @@
+import fetch from "node-fetch";
+
+class CommunityListResponse {
+  pagination: {
+    next_link: string | unknown;
```
anttimaki, line 5: From Typescript's perspective, having `string | unknown` is the same thing as just having `unknown`. So at best this is type hinting for the human reader. Not sure about the background and aims of this ticket (the ticket itself is little help in the matter) so can't say if that's fine or not.

packages/thunderstore-api/src/client.ts
```diff
@@ -0,0 +1,35 @@
...
+  return fetch(path, {
+    headers: {
+      "Content-Type": "application/json",
+      Accept: "application/json",
+    },
+  });
+}
+
+export async function fetchCommunityList() {
+  const data = await apiFetch(
```
anttimaki, line 27: I'm not a fan of the way `await` and `then`s are mixed here.

packages/thunderstore-api/src/tests/client.test.ts
```diff
@@ -0,0 +1,20 @@
+import { describe, expect, test } from "@jest/globals";
+import { fetchCommunityList } from "../client";
+
+const ASD = {
```
anttimaki, line 4: I'm sure you can do better than "ASD".

packages/thunderstore-api/src/client.ts
MythicManiac: The intention was to have the function return `knowntype | unknown` as that would force validation (or explicit ignorance of validation by casting) at the caller's end. The way it's implemented here does not really make sense though.

packages/thunderstore-api/jest.config.ts
MythicManiac: I don't see any reason why should we deviate either, and didn't remember jest was already configured to dapper. We should probably stick to that way of doing things if it works as this is more complex, I was pretty sure there should be a way to get typescript to work with jest _without_ using babel, but their docs seemed to suggest otherwise.

In any case, as it seems possible to run jest without babel, we should yeet that dependency. Moreover I'm 90% certain we don't need `node-fetch` either if you just run the tests with a node version of 17 or higher.

MythicManiac: Well turns out the docs do mention `ts-jest` just fine, it's just not the first option in the list: https://jestjs.io/docs/getting-started#via-ts-jest

So this should be changed to that configuration, could probably also update the jest version used by Dapper

packages/thunderstore-api/babel.config.js
Oksamies: Ah, true, thanks for the point out 💯

packages/thunderstore-api/jest.config.ts
anttimaki: The `node-fetch` issue seems a hairy one. While Node 18 provides fetch, the matching version of @types/node does not (there's a long conversation regarding this on their GitHub page). However, removing `testEnvironment: "node"` from Jest config seems to fix the issue, although "node" is the default value for that setting. I guess `preset: "ts-jest"` in the same config might solve this by targeting something like browser, which wouldn't be ideal I guess. Or maybe they have some different approach altogether. However, this seems to fix the issue, at least for now, so I'll go with that one.

anttimaki:
Similar changes were merged to master from #777, closing this one.

PR #777, opened 2023-08-22
Add @thunderstore/thunderstore-api

anttimaki:
This is an alternative for https://github.com/thunderstore-io/thunderstore-ui/pull/774, addressing the feedback I provided there. The actual implementation is as simple as in the old PR to keep the PR small and get feedback early.

Also the changes target master branch rather than the existing WIP branch. The refactoring needs for the Dapper package seemed like a shapeless plop, and I was hard time deciding where to start from, so I decided to keep it simple and just replicate the old PR. The router changes from the WIP branch can be cherry picked later on top of these changes.

### DIFF
--- a/packages/thunderstore-api/README.md
+++ b/packages/thunderstore-api/README.md
@@ -1,0 +1,8 @@
+# thunderstore-api
+
+Thunderstore API client
+
+## Scripts
+
+- `yarn run build`: Builds the project
+- `yarn run dev`: Builds the project & watches files for changes, triggering a rebuild
--- a/packages/thunderstore-api/README.md
+++ b/packages/thunderstore-api/README.md
@@ -6,8 +6,8 @@ Thunderstore API client
 
 - `yarn run build`: Builds the project
 - `yarn run dev`: Builds the project & watches files for changes, triggering a rebuild
-- `yarn run jest`: run test suite once
-- `yarn run test`: runs test suite & watches files for changes, triggering rerun
+- `yarn run test`: Runs the test suite once
+- `yarn run watch`: Runs the test suite & watches files for changes, triggering a rerun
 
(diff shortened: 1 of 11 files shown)


packages/thunderstore-api/README.md
```diff
@@ -0,0 +1,13 @@
+# thunderstore-api
+
+Thunderstore API client
+
+## Scripts
+
+- `yarn run build`: Builds the project
+- `yarn run dev`: Builds the project & watches files for changes, triggering a rebuild
+- `yarn run jest`: run test suite once
+- `yarn run test`: runs test suite & watches files for changes, triggering rerun
```
MythicManiac, line 10: I feel like conventionally this behavior would be behind a command named `watch` or such. Or at the very least the `test` command itself should be a one-off run you can use e.g. on CI pipelines.

packages/thunderstore-api/src/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * TODO: this is a naive implementation. Actual process should include:
+ * - Typesafe parsing of URLs from parameters
+ * - Using env variable to define backend
```
MythicManiac, line 4: I'd prefer the function itself here does not depend on any globals, but rather takes the necessary configs as parameters somehow (probably as an object with multiple potential keys to make partial re-use easier for callers.)

MythicManiac (approved):
A few minor comments but overall looks good 👍

packages/thunderstore-api/README.md
anttimaki: Changed the usage across the monorepo in a separate commit.

packages/thunderstore-api/src/index.ts
anttimaki: Updated the comment so I'll remember this.

PR #783, opened 2023-08-24
Add actual page count figuring to pkg list

Oksamies:
Add current page number and total object count to filters in PackageSearchLayout, so that they can be used to figure out how many pages there should be and what page is currently active PageSize and totalCount are still not configurable, but probably should. Implementation of all the configurable stuff needs to be unified

refs TS-1781

### DIFF
--- a/packages/cyberstorm/src/components/Layout/PackageListings/PackageListings.tsx
+++ b/packages/cyberstorm/src/components/Layout/PackageListings/PackageListings.tsx
@@ -79,6 +79,8 @@ export default function PackageListings(props: PackageListingsProps) {
     if (!isEqual(updatedAvailableCategories, filters.availableCategories)) {
       filters.setAvailableCategories(updatedAvailableCategories);
     }
+
+    filters.setPageObjectCount(datas.length);
   }, [
     datas,
     filters?.availableCategories,
--- a/packages/cyberstorm/src/components/Layout/PackageSearchLayout/PackageSearchLayout.tsx
+++ b/packages/cyberstorm/src/components/Layout/PackageSearchLayout/PackageSearchLayout.tsx
@@ -47,11 +47,17 @@ export class Filters {
   setKeywords: React.Dispatch<React.SetStateAction<string[]>>;
   availableCategories: CategoriesProps;
   setAvailableCategories: React.Dispatch<React.SetStateAction<CategoriesProps>>;
+  page: number;
+  setPage: React.Dispatch<React.SetStateAction<number>>;
+  pageObjectCount: number;
+  setPageObjectCount: React.Dispatch<React.SetStateAction<number>>;
 
   constructor() {
     [this.keywords, this.setKeywords] = useState<string[]>([]);
     [this.availableCategories, this.setAvailableCategories] =
       useState<CategoriesProps>({});
+    [this.page, this.setPage] = useState<number>(1);
+    [this.pageObjectCount, this.setPageObjectCount] = useState<number>(1);
   }
 }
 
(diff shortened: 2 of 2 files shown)


packages/cyberstorm/src/components/Layout/PackageListings/PackageListings.tsx
```diff
@@ -79,6 +79,8 @@ export default function PackageListings(props: PackageListingsProps) {
     if (!isEqual(updatedAvailableCategories, filters.availableCategories)) {
       filters.setAvailableCategories(updatedAvailableCategories);
     }
+
+    filters.setPageObjectCount(datas.length);
```
anttimaki, line 83: Why do this here? This effect is for updating categories, and page count is totally unrelated. I'd do this in a separate effect, or the one that fetches the data above.

Also when adding stuff to effects, update the dependency list argument as well.

packages/cyberstorm/src/components/Layout/PackageSearchLayout/PackageSearchLayout.tsx
```diff
@@ -47,11 +47,17 @@ export class Filters {
...
+  page: number;
+  setPage: React.Dispatch<React.SetStateAction<number>>;
+  pageObjectCount: number;
+  setPageObjectCount: React.Dispatch<React.SetStateAction<number>>;
 
   constructor() {
     [this.keywords, this.setKeywords] = useState<string[]>([]);
     [this.availableCategories, this.setAvailableCategories] =
       useState<CategoriesProps>({});
+    [this.page, this.setPage] = useState<number>(1);
```
anttimaki, line 59: I don't like this. You're supposed to use hooks in very specific places, and a class isn't one of them. This was mentioned somewhere in my commit message oslt when I refactored this stuff, I think this whole `// TODO: OVERKILL???` part should be refactored. Not that this makes things much worse since we already use hooks in this class, but it's more tech debt anyway.

packages/cyberstorm/src/components/Layout/PackageSearchLayout/PackageSearchLayout.tsx
```diff
@@ -47,11 +47,17 @@ export class Filters {
   setKeywords: React.Dispatch<React.SetStateAction<string[]>>;
   availableCategories: CategoriesProps;
   setAvailableCategories: React.Dispatch<React.SetStateAction<CategoriesProps>>;
+  page: number;
+  setPage: React.Dispatch<React.SetStateAction<number>>;
+  pageObjectCount: number;
```
anttimaki, line 52: ~Why not just `pageCount`?~
Edit: oh wait this has nothing to do with pages, which makes the naming even more confusing. Why not use `totalCount` to match the consuming component's naming convention?

anttimaki (changes requested):
Why should the total count of objects be configurable?

packages/cyberstorm/src/components/Layout/PackageSearchLayout/PackageSearchLayout.tsx
Oksamies: True, I'll try to refactor the provider when I get back home 👍

packages/cyberstorm/src/components/Layout/PackageListings/PackageListings.tsx
anttimaki: Checked this component a bit more while working on Dapper refactoring. If I'm not mistaken this approach won't work anyway since `datas.length` contains only the returned packages, i.e. is limited by the requested "page size" (20 packages per page). Backend needs to return the total number of packages separately, which it currently does not do (at least according to Dapper interface/implementations).

PR #791, opened 2023-08-29
@thunderstore/dapper: add second entrypoint for schemas

anttimaki:


VilppeRiskidev (approved):
Very nice

VilppeRiskidev:
Chromatic build fails, the code changes themselves look ok

anttimaki:
> Chromatic build fails, the code changes themselves look ok

Yeah, test build as well, and it seems that the failures are just the ones I mention in commit message of the second commit: importing stuff that wasn't exported by Dapper. It would be nice if IDE could catch these, but better in CI than not at all I guess. I'll add a fix for these.

PR #793, opened 2023-08-30
Add authentication to Cyberstorm

Oksamies:
Add minimal authentication to Cyberstorm
That is coupled with old sites Django sessions
Add auth and redirection of unauthorized requests to "premium" page
into the middleware of nextjs
Add .env variables for disabling auth in development
Add .env variable for auth encryption secret

refs TS-1810

### DIFF
--- a/apps/cyberstorm-nextjs/.env.development
+++ b/apps/cyberstorm-nextjs/.env.development
@@ -4,3 +4,7 @@ NEXT_PUBLIC_API_DOMAIN=http://thunderstore.localhost
 NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_DOMAIN/api/cyberstorm/
 CYBERSTORM_AUTH_ENABLED=
 CYBERSTORM_AUTH_SECRET=development
+CYBERSTORM_AUTH_BASE_URL=https://auth.thunderstore.io/auth/login
+CYBERSTORM_AUTH_DISCORD_URL=/discord/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
+CYBERSTORM_AUTH_GITHUB_URL=/github/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
+CYBERSTORM_AUTH_OVERWOLF_URL=/overwolf/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
--- a/apps/cyberstorm-nextjs/.env.development
+++ b/apps/cyberstorm-nextjs/.env.development
@@ -2,9 +2,10 @@
 # the security implications that entails.
 NEXT_PUBLIC_API_DOMAIN=http://thunderstore.localhost
 NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_DOMAIN/api/cyberstorm/
-CYBERSTORM_AUTH_ENABLED=
-CYBERSTORM_AUTH_SECRET=development
-CYBERSTORM_AUTH_BASE_URL=https://auth.thunderstore.io/auth/login
-CYBERSTORM_AUTH_DISCORD_URL=/discord/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
-CYBERSTORM_AUTH_GITHUB_URL=/github/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
-CYBERSTORM_AUTH_OVERWOLF_URL=/overwolf/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
+AUTH_ENABLED=
+AUTH_SECRET=development
+NEXT_PUBLIC_AUTH_BASE_URL=https://auth.thunderstore.io/auth/login
+NEXT_PUBLIC_AUTH_RETURN_URL=?next=https%3A%2F%2Fbeta.thunderstore.io%2F
+NEXT_PUBLIC_AUTH_DISCORD_URL=$NEXT_PUBLIC_AUTH_BASE_URL/discord/
+NEXT_PUBLIC_AUTH_GITHUB_URL=$NEXT_PUBLIC_AUTH_BASE_URL/github/
+NEXT_PUBLIC_AUTH_OVERWOLF_URL=$NEXT_PUBLIC_AUTH_BASE_URL/overwolf/
--- a/apps/cyberstorm-nextjs/.env.development
+++ b/apps/cyberstorm-nextjs/.env.development
@@ -2,3 +2,5 @@
 # the security implications that entails.
 NEXT_PUBLIC_API_DOMAIN=http://thunderstore.localhost
 NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_DOMAIN/api/cyberstorm/
+CYBERSTORM_AUTH_ENABLED=
+CYBERSTORM_AUTH_SECRET=development
--- a/apps/cyberstorm-nextjs/app/login/page.tsx
+++ b/apps/cyberstorm-nextjs/app/login/page.tsx
@@ -1,5 +1,11 @@
 import { BetaLoginLayout } from "@thunderstore/cyberstorm";
 
 export default function Page() {
-  return <BetaLoginLayout />;
+  return (
+    <BetaLoginLayout
+      discordAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_DISCORD_URL}`}
+      githubAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_GITHUB_URL}`}
+      overwolfAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_OVERWOLF_URL}`}
+    />
+  );
 }
--- a/apps/cyberstorm-nextjs/app/login/page.tsx
+++ b/apps/cyberstorm-nextjs/app/login/page.tsx
@@ -3,9 +3,9 @@ import { BetaLoginLayout } from "@thunderstore/cyberstorm";
 export default function Page() {
   return (
     <BetaLoginLayout
-      discordAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_DISCORD_URL}`}
-      githubAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_GITHUB_URL}`}
-      overwolfAuthUrl={`${process.env.CYBERSTORM_AUTH_BASE_URL}${process.env.CYBERSTORM_AUTH_OVERWOLF_URL}`}
+      discordAuthUrl={`${process.env.NEXT_PUBLIC_AUTH_DISCORD_URL}${process.env.NEXT_PUBLIC_AUTH_RETURN_URL}`}
+      githubAuthUrl={`${process.env.NEXT_PUBLIC_AUTH_GITHUB_URL}${process.env.NEXT_PUBLIC_AUTH_RETURN_URL}`}
+      overwolfAuthUrl={`${process.env.NEXT_PUBLIC_AUTH_OVERWOLF_URL}${process.env.NEXT_PUBLIC_AUTH_RETURN_URL}`}
     />
   );
 }
--- a/apps/cyberstorm-nextjs/middleware.ts
+++ b/apps/cyberstorm-nextjs/middleware.ts
@@ -1,0 +1,56 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import {
+  AUTH_COOKIE_NAME,
+  decode,
+  handleJWTGenerationAndReturnResponseOrRedirect,
+  redirectToLogin,
+} from "./utils/auth";
+import { errors } from "jose";
+
+export const config = {
+  matcher: [
+    /*
+     * Match all request paths except for the ones starting with:
+     * - _next/static (static files)
+     * - _next/image (image optimization files)
+     * - favicon.ico (favicon file)
+     */
+    "/((?!_next/static|_next/image|favicon.ico|images).*)",
+  ],
+};
+
+export async function middleware(request: NextRequest) {
+  const req = request;
+  const response = NextResponse.next();
+
+  // For the sake of keeping dev configuration easy and not confusing
+  if (!process.env.CYBERSTORM_AUTH_ENABLED) {
+    return response;
+  }
+
+  const JWTToken = request.cookies.get(AUTH_COOKIE_NAME);
+  if (JWTToken) {
+    try {
+      await decode(JWTToken.value);
+    } catch (e) {
+      if (e instanceof errors.JWTExpired) {
+        return handleJWTGenerationAndReturnResponseOrRedirect(req, response);
+      }
+      if (e instanceof errors.JWEDecryptionFailed) {
+        return new NextResponse(null, { status: 400 });
+      }
+      if (e instanceof errors.JWSSignatureVerificationFailed) {
+        return new NextResponse(null, { status: 403 });
+      }
+      return new NextResponse(null, { status: 500 });
+    }
+    return response;
+  } else {
+    if (request.cookies.get("sessionid")) {
+      return handleJWTGenerationAndReturnResponseOrRedirect(req, response);
+    } else {
+      return redirectToLogin(request, response);
+    }
+  }
+}
--- a/apps/cyberstorm-nextjs/middleware.ts
+++ b/apps/cyberstorm-nextjs/middleware.ts
@@ -25,7 +25,7 @@ export async function middleware(request: NextRequest) {
   const response = NextResponse.next();
 
   // For the sake of keeping dev configuration easy and not confusing
-  if (!process.env.CYBERSTORM_AUTH_ENABLED) {
+  if (!process.env.AUTH_ENABLED) {
     return response;
   }
 
--- a/apps/cyberstorm-nextjs/utils/auth.ts
+++ b/apps/cyberstorm-nextjs/utils/auth.ts
@@ -1,0 +1,131 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import hkdf from "@panva/hkdf";
+import { EncryptJWT, jwtDecrypt } from "jose";
+import { v4 as uuid } from "uuid";
+import { isRecord } from "./typeChecks";
+
+interface PartialUserStatus {
+  username: string;
+  subscription: { expires: string | null };
+}
+
+export const AUTH_COOKIE_NAME = "CyberstormAuth";
+
+const now = () => (Date.now() / 1000) | 0;
+
+function isPartialUserStatus(
+  userStatus: unknown
+): userStatus is PartialUserStatus {
+  return (
+    isRecord(userStatus) &&
+    typeof userStatus.username === "string" &&
+    isRecord(userStatus.subscription) &&
+    (typeof userStatus.subscription.expires === "string" ||
+      typeof userStatus.subscription.expires === null)
+  );
+}
+
+async function getDerivedEncryptionKey(secret: string | Buffer) {
+  return await hkdf(
+    "sha256",
+    secret,
+    "",
+    "Cyberstorm Generated Encryption Key",
+    32
+  );
+}
+
+async function encode(token: { username: string; expires: number }) {
+  if (process.env.CYBERSTORM_AUTH_SECRET) {
+    const encryptionSecret = await getDerivedEncryptionKey(
+      process.env.CYBERSTORM_AUTH_SECRET
+    );
+    return await new EncryptJWT(token)
+      .setProtectedHeader({ alg: "dir", enc: "A256GCM" })
+      .setIssuedAt()
+      .setExpirationTime(Math.min(now() + 900, token.expires))
+      .setJti(uuid())
+      .encrypt(encryptionSecret);
+  } else {
+    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+  }
+}
+
+export async function decode(token: string) {
+  if (process.env.CYBERSTORM_AUTH_SECRET) {
+    const encryptionSecret = await getDerivedEncryptionKey(
+      process.env.CYBERSTORM_AUTH_SECRET
+    );
+    const { payload } = await jwtDecrypt(token, encryptionSecret, {
+      clockTolerance: 15,
+    });
+    return payload;
+  } else {
+    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+  }
+}
+
+async function fetchSubStatus(sessionid: string | undefined): Promise<{
+  username: string | null;
+  subActive: boolean;
+  expires: number;
+}> {
+  if (sessionid) {
+    const res = await fetch(
+      "https://thunderstore.io/api/experimental/current-user/",
+      {
+        headers: {
+          authorization: `Session ${sessionid}`,
+          "Content-Type": "application/json",
+        },
+      }
+    );
+    if (res.status === 200) {
+      const data = await res.json();
+      if (isPartialUserStatus(data)) {
+        if (data.subscription.expires) {
+          const expiryTimeEpoch = Date.parse(data.subscription.expires);
+          if (!Number.isNaN(expiryTimeEpoch) && expiryTimeEpoch > Date.now()) {
+            return {
+              username: data.username,
+              subActive: true,
+              expires: expiryTimeEpoch / 1000, // jwtDecrypt Checks the time in seconds format
+            };
+          }
+        }
+      }
+    }
+  }
+  return { username: null, subActive: false, expires: 0 };
+}
+
+export function redirectToLogin(request: NextRequest, response: NextResponse) {
+  if (request.nextUrl.pathname !== "/login") {
+    return NextResponse.redirect(`${request.nextUrl.origin}/login`);
+  }
+  return response;
+}
+
+export async function handleJWTGenerationAndReturnResponseOrRedirect(
+  request: NextRequest,
+  response: NextResponse
+) {
+  if (!request.cookies.get("sessionid")) {
+    return redirectToLogin(request, response);
+  }
+
+  const subStatus = await fetchSubStatus(
+    request.cookies.get("sessionid")?.value
+  );
+  if (subStatus.username && subStatus.subActive) {
+    const newToken = await encode({
+      username: subStatus.username,
+      expires: subStatus.expires,
+    });
+    response.cookies.set(AUTH_COOKIE_NAME, newToken);
+    return response;
+  } else {
+    return redirectToLogin(request, response);
+  }
+}
--- a/apps/cyberstorm-nextjs/utils/auth.ts
+++ b/apps/cyberstorm-nextjs/utils/auth.ts
@@ -37,9 +37,9 @@ async function getDerivedEncryptionKey(secret: string | Buffer) {
 }
 
 async function encode(token: { username: string; expires: number }) {
-  if (process.env.CYBERSTORM_AUTH_SECRET) {
+  if (process.env.AUTH_SECRET) {
     const encryptionSecret = await getDerivedEncryptionKey(
-      process.env.CYBERSTORM_AUTH_SECRET
+      process.env.AUTH_SECRET
     );
     return await new EncryptJWT(token)
       .setProtectedHeader({ alg: "dir", enc: "A256GCM" })
@@ -48,21 +48,21 @@ async function encode(token: { username: string; expires: number }) {
       .setJti(uuid())
       .encrypt(encryptionSecret);
   } else {
-    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+    throw new Error("AUTH_SECRET is missing");
   }
 }
 
 export async function decode(token: string) {
-  if (process.env.CYBERSTORM_AUTH_SECRET) {
+  if (process.env.AUTH_SECRET) {
     const encryptionSecret = await getDerivedEncryptionKey(
-      process.env.CYBERSTORM_AUTH_SECRET
+      process.env.AUTH_SECRET
     );
     const { payload } = await jwtDecrypt(token, encryptionSecret, {
       clockTolerance: 15,
     });
     return payload;
   } else {
-    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+    throw new Error("AUTH_SECRET is missing");
   }
 }
 
--- a/packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
+++ b/packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
@@ -188,6 +188,11 @@
   line-height: var(--line-height--l);
 }
 
+.loginButton > a > div > svg {
+  width: 1rem;
+  height: 1rem;
+}
+
 @media (max-width: 85rem) {
   .advertisementTitle {
     display: flex;
--- a/packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
+++ b/packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
@@ -1,14 +1,20 @@
 import styles from "./BetaLoginLayout.module.css";
-import { Button } from "../../Button/Button";
+import { Button, PlainButton } from "../../Button/Button";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faDiscord, faGithub } from "@fortawesome/free-brands-svg-icons";
 import { OverwolfLogo, ThunderstoreLogo } from "../../../svg/svg";
 import { PrivacyPolicyLink, TermsOfServiceLink } from "../../Links/Links";
 
+interface Props {
+  discordAuthUrl?: string;
+  githubAuthUrl?: string;
+  overwolfAuthUrl?: string;
+}
+
 /**
  * Cyberstorm Beta Login Layout
  */
-export function BetaLoginLayout() {
+export function BetaLoginLayout(props: Props) {
   return (
     <div className={styles.root}>
       <div className={styles.login}>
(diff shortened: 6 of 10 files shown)


apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,129 @@
...
+      `https://thunderstore.io/api/experimental/current-user/`,
+      {
+        headers: {
+          authorization: `Session ${sessionid}`,
+          "Content-Type": "application/json",
+        },
+      }
+    );
+    if (res.status === 200) {
+      const currentUserStatus: UserStatusType = await res.json();
```
anttimaki, line 76: You're converting `any` to `UserStatusType` on blind faith here. Some type guard or at least using nullish chaining below would be welcome. E.g. the `Date.parse` below can have odd results otherwise.

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+  return { username: null, subActive: false, expires: 0 };
+}
+
+export function redirectToPremium(
+  request: NextRequest,
+  response: NextResponse
+) {
+  if (request.nextUrl.pathname !== "/premium") {
```
anttimaki, line 98: I'm a bit confused due context swapping to NextRequest and NextResponse. Why do we check the request path here? Does it contain the original path the user actually requested? Can some redirect already have changed this? Would it be better to just redirect user to the premium page no matter what?

apps/cyberstorm-nextjs/utils/auth.ts
anttimaki, line 1: Again here the if-elses seem unnecessarily deeply nested.

anttimaki (commented):
Reviewed as a regular PR but I guess in the sense this is throw-away code since the check won't be there forever, so use your judgment. Maybe fix at least the things the IDE complains about and ensure error handling is user friendly?

anttimaki:
The commit message still seems odd. Is it an attempt to write a haiku? :trollface:

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+  ],
+};
+
+export async function middleware(request: NextRequest) {
+  const req = request;
+  const response = NextResponse.next();
+  // Do this temporarily while in dev, since the django session might not be in the browser
+  // request.cookies.set("sessionid", "");
+
+  if (process.env.CYBERSTORM_AUTH_ENABLED) {
```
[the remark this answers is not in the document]
Oksamies: I lean more towards nesting when possible, but you are correct that this way it's more readable 👌

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+    // JWT exists?
+    if (request.cookies.has("CyberstormAuth")) {
+      // Validate JWT
+      try {
```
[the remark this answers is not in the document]
Oksamies: True true, didn't think of this. TBH, whatever reason; haven't had to use that many try catches in the past. Not even with Python 🤷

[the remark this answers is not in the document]
Oksamies: Guess it might be also that, just haven't used them, when should have 😅

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,129 @@
...
+  } else {
+    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+  }
+}
+
+export async function decode(
+  token: string
+): Promise<Record<string, unknown> | null> {
+  if (process.env.CYBERSTORM_AUTH_SECRET) {
+    if (!token) return null;
```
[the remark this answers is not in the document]
Oksamies: A left over type check I probably forgot. Removed it 👍

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: So I made a validator, but I'm not entirely sure if validators are supposed to be made in that way. Anyways, if that's not a valid way to do it let me know 👌

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: The request contains the path, to the user requested page and if we don't want to redirect endlessly we'll need to make sure the redirection is not done on the `/premium` page. Although, now I'm, not entirely sure if next could handle this fine after all.

I'll check and test this out, when I get the local env fully setup.

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+        } else {
+          // Token value was empty, redirect to "Go Premium"
+          return redirectToPremium(request, response);
+        }
+      } catch (e: any) {
+        if (e.code === "ERR_JWT_EXPIRED") {
+          // Expired so figure and validate django session and generate a new JWT token
+          return handleJWTGenerationAndReturnResponseOrRedirect(req, response);
+        }
+        throw e;
```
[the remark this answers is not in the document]
Oksamies: Hmm, I'm not sure actually. The dev env throws the error of course and displays it, but no clue if it displays the customized "error 500" page we have. I'll ask around.

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: Checked this out, worked nicely with this and without it just kept redirecting endlessly

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // For the sake of keeping dev configuration easy and not confusing
+  if (process.env.CYBERSTORM_AUTH_ENABLED) {
+    return response;
+  }
+
+  // JWT exists?
+  if (request.cookies.has("CyberstormAuth")) {
+    // Validate JWT
+
```
anttimaki, line 34: Nitpick but I'm irked about the empty line after this comment. Also I'm not 100% sure that the comments in this if-block add that much value, they mostly just say the same thing the code does?

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,145 @@
...
+
+interface UserStatusType {
+  username: string;
+  capabilities: string[];
+  subscription: { expires: string | null };
+  rated_packages: string[];
+  teams: string[];
+}
+
+function isUserStatusType(data: any): boolean {
```
anttimaki, line 15: - Nitpick: not a fan of `data` as a variable name even in this context
- Type `unknown` would be nice here since the whole point of the function is that we don't know the type of the argument and want to find out by duck typing it. If I'm not mistaken, changing the type would also reveal that this function breaks if the argument is undefined
- The whole point of a type guard is to give TS compiler more information about the typing. Now the function returns boolean, which tells TS squat. Consequently you need to manually type cast the value even after this function is called. Check the documentation I linked to see how this should actually be done 😉
- The type checking in the function body could be improved by e.g. using Array.isArray

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,145 @@
...
+      .setJti(uuid())
+      .encrypt(encryptionSecret);
+  } else {
+    throw new Error("CYBERSTORM_AUTH_SECRET is missing");
+  }
+}
+
+export async function decode(
+  token: string
+): Promise<Record<string, unknown> | null> {
```
anttimaki, line 59: Is there a reason to manually defining the return type (sometimes there is)? Mentioning it because it doesn't match what the current implementation can actually return.

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,145 @@
...
+        headers: {
+          authorization: `Session ${sessionid}`,
+          "Content-Type": "application/json",
+        },
+      }
+    );
+    if (res.status === 200) {
+      const data = await res.json();
+      if (isUserStatusType(data)) {
+        const currentUserStatus = data as UserStatusType;
```
anttimaki, line 91: With proper type guard this isn't needed.

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (currentUserStatus.subscription.expires) {
+          const expiryTimeEpoch = Date.parse(
```
anttimaki, line 93: Date.parse might return NaN I think. What happens if it's compared against Date.now?

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    try {
+      await decode(JWTToken.value);
+    } catch (e) {
+      if (e instanceof errors.JWTExpired) {
+        // Expired so figure and validate django session and generate a new JWT token
+        return handleJWTGenerationAndReturnResponseOrRedirect(req, response);
+      }
+      throw e;
```
anttimaki, line 48: If I'm not mistaken the discussion on Discord agreed on redirecting to error page instead of throwing?

anttimaki (changes requested):
Starting to look pretty good, but I'm returning this for another round to fix the type guard stuff.

apps/cyberstorm-nextjs/middleware.ts
Oksamies: Now that I think about; true. I'll remove all of those that, give no value 👍

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: I added a `!isNan` check

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: In this case I think, I again forgot to remove it. But sometimes I've also wanted to add those there before hand, so that I get type errors when developing and fucking around with the code 🤷 Bad practices maybe? I guess

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: This part of the code has changed few times because of discussions, but anyways I've fixed the stuff you mentioned there 👌 And thanks for the more in depth discussion on discord also 💯

apps/cyberstorm-nextjs/middleware.ts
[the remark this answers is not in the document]
anttimaki: 400 is "bad request", so did you mean 403 or did you mean "forbidden"?

Generally speaking I'd say that it depends on the situations that actually cause `decode()` to throw. For "hey, this token isn't issued by us", 403 would probably be the right choice. If the token is unreadable/malformed, then it's bad request. And for "hey, the server doesn't have the private key set up", 500 would be best.

If it's impossible to to tell the cases apart, i.e. there's no more stuff like `errors.JWTExpired`, then perhaps go with 403. Otherwise I would say 500 (since it's basically us saying we don't know what went wrong), but 403 gives a hint that the problem is somehow related to the authentication, and may user/us help resolve the issue.

apps/cyberstorm-nextjs/utils/auth.ts
anttimaki: I haven't really formed an educated opinion on the matter. My gut feeling is that it's not inherently a bad practice, it can be a nice tool for some situations. If it's obvious that a function returns a string, it might be a bit of clutter to spell it out. If it's not obvious and you want to be 100% that some subsequent changes won't "widen" the return type accidentally, then spelling it out makes sense. But as seen here, spelling is out means you must also ensure it stays up to date if the return type is "narrowed". Having the spelled out type not match the actual return value is obviously bad, unless it's to make the function to implement some interface defined elsewhere.

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,143 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import hkdf from "@panva/hkdf";
+import { EncryptJWT, JWTPayload, jwtDecrypt } from "jose";
```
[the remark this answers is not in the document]
anttimaki: :reee:

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,143 @@
...
+
+interface UserStatusType {
+  username: string;
+  capabilities: string[];
+  subscription: { expires: string | null };
+  rated_packages: string[];
+  teams: string[];
+}
+
+function isUserStatusType(userStatus: unknown): userStatus is UserStatusType {
```
anttimaki, line 15: - Based on the discord discussion I expected we'd only check some fields here, but being thorough ain't bad (unless changes to e.g. `rated_packages` in the backend breaks this type guard and thus auth check.
- I had a kind of allergic reaction to to the repetition and have a suggestion (see below). It should do the same thing, so the change is only for readability and thus not absolutely necessary, but you could consider it while fixing the unused import error above
- This could be a nice candidate for unit tests, since lots of repetition makes it's easy have a bug in the implementation, and writing the test cases is rather straightforward

```
const isRecord = (obj: unknown): obj is Record<string, unknown> =>
  obj instanceof Object;

const isStringArray = (arr: unknown): arr is string[] =>
  Array.isArray(arr) && arr.every((s) => typeof s === "string");

function isUserStatusType(userStatus: unknown): userStatus is UserStatusType {
  return (
    isRecord(userStatus) &&
    typeof userStatus.username === "string" &&
    isRecord(userStatus.subscription) &&
    (typeof userStatus.subscription.expires === "string" ||
      typeof userStatus.subscription.expires === null) &&
    isStringArray(userStatus.capabilities) &&
    isStringArray(userStatus.rated_packages) &&
    isStringArray(userStatus.teams)
  );
}
```

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+async function fetchSubStatus(sessionid: string | undefined): Promise<{
+  username: string | null;
+  subActive: boolean;
+  expires: number;
+}> {
+  if (sessionid) {
+    const res = await fetch(
+      `https://thunderstore.io/api/experimental/current-user/`,
```
anttimaki, line 88: Nitpick: no need for ticks, quotes would work fine.

anttimaki (changes requested):
Good work but still a few new things or things I hadn't noticed earlier.

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function middleware(request: NextRequest) {
+  const req = request;
+  const response = NextResponse.next();
+
+  // For the sake of keeping dev configuration easy and not confusing
+  if (process.env.CYBERSTORM_AUTH_ENABLED) {
+    return response;
+  }
+
+  if (request.cookies.has("CyberstormAuth")) {
```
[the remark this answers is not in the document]
Oksamies: Makes sense 👍

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -0,0 +1,50 @@
...
+    const JWTToken = request.cookies.get("CyberstormAuth");
+    if (!JWTToken) {
+      return redirectToLogin(request, response);
```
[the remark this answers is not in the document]
Oksamies: Yeah, seems like I didn't figure out the if else logic fully, after moving it out of the try catch. Should be fixed in the next commit.

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: > Based on the discord discussion I expected we'd only check some fields here, but being thorough ain't bad (unless changes to e.g. rated_packages in the backend breaks this type guard and thus auth check.

Hmm, yeah I guess that checking those kind of doesn't make sense, since we'll be fetching those in a separate call, whenever needed?

I'll put your suggestion there, but with the unused stuff left out.

apps/cyberstorm-nextjs/middleware.ts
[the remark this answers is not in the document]
Oksamies: Wise words, I've added few more error checks.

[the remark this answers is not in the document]
Oksamies: Also I might revisit this, once I get the Django side done.

apps/cyberstorm-nextjs/.env.development
```diff
@@ -2,3 +2,8 @@
 # the security implications that entails.
 NEXT_PUBLIC_API_DOMAIN=http://thunderstore.localhost
 NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_DOMAIN/api/cyberstorm/
+CYBERSTORM_AUTH_ENABLED=
+CYBERSTORM_AUTH_SECRET=development
+CYBERSTORM_AUTH_DISCORD_URL=https://auth.thunderstore.io/auth/login/discord/?next=https%3A%2F%2Fbeta.thunderstore.io%2F
```
anttimaki, line 7: Should these URL variable names have the `NEXT_PUBLIC` prefix?

(Also I think having separate variable for each provider is 100% fine here since there's only three and we're not planning to add more but generally speaking the repetition could be avoided by having one variable with the URL up to `.../login/` and another with the query parameter.)

apps/cyberstorm-nextjs/app/login/page.tsx
```diff
@@ -1,5 +1,11 @@
 import { BetaLoginLayout } from "@thunderstore/cyberstorm";
 
 export default function Page() {
-  return <BetaLoginLayout />;
+  return (
+    <BetaLoginLayout
+      discordAuthUrl={process.env.CYBERSTORM_AUTH_DISCORD_URL}
```
anttimaki, line 6: Doesn't a change like this also affect the related Storybook story?

apps/cyberstorm-nextjs/utils/auth.ts
```diff
@@ -0,0 +1,132 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import hkdf from "@panva/hkdf";
+import { EncryptJWT, jwtDecrypt } from "jose";
+import { v4 as uuid } from "uuid";
+import { isRecord } from "./typeChecks";
+
+interface UserStatusType {
```
anttimaki, line 8: As mentioned in some earlier comment or discord discussion, when taking this approach with the type guard, this could be named `PartialUserStatus` oslt, to make it obvious for other developers that not all properties are type checked and they shouldn't assume so or their foot might get new ventilation holes.

Obviously also remove the props that are not checked from the interface.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
```diff
@@ -1,14 +1,20 @@
 import styles from "./BetaLoginLayout.module.css";
-import { Button } from "../../Button/Button";
+import { Button, PlainButton } from "../../Button/Button";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faDiscord, faGithub } from "@fortawesome/free-brands-svg-icons";
 import { OverwolfLogo, ThunderstoreLogo } from "../../../svg/svg";
 import { PrivacyPolicyLink, TermsOfServiceLink } from "../../Links/Links";
 
+export interface BetaLoginLayoutProps {
```
anttimaki, line 8: I'm not sure this needs to be exported. Also for brevity it could be just `Props`, since the context makes it clear what it's used for.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
```diff
@@ -19,30 +25,42 @@ export function BetaLoginLayout() {
...
-                colorScheme="red"
-                label="Ovewolf"
-                leftIcon={<OverwolfLogo />}
-                paddingSize="large"
-              />
-            </div>
+            {props.discordAuthUrl ? (
+              <div className={styles.loginButton}>
+                <a href={props.discordAuthUrl}>
+                  <PlainButton
```
anttimaki, line 31: You can't place buttons inside anchor elements. (Well I guess you can, but it's illegal. You wouldn't place a button inside a car.)

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
```diff
@@ -19,30 +25,42 @@ export function BetaLoginLayout() {
...
+                  />
+                </a>
+              </div>
+            ) : null}
+            {props.overwolfAuthUrl ? (
+              <div className={styles.loginButton}>
+                <a href={props.overwolfAuthUrl}>
+                  <PlainButton
+                    colorScheme="red"
+                    label="Ovewolf"
```
anttimaki, line 57: Please fix the typo while at it. Or is this some Swedish wolf?

anttimaki (changes requested):
The code in the initial commit is almost there, good job.

But IDK why you went and added the second commit to this PR instead of creating another PR for those changes. I'm not 100% sure but probably the PRs wouldn't even need to be chained, since the changes are in different files. Nor are they logically that related. Now change requests to the new commit are blocking the old commit.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: The `PlainButton` is not a button, but just a div that looks like a button. But IIRC the current button/link component implementation will be refactored in the near future™️ Consult designer on discord about the timeline/importance.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: It does not need to be exported 👍 Though I'm not sure about the naming thing, I it when those are "specifically named", but I don't mind going by simpler names 🤷

apps/cyberstorm-nextjs/utils/auth.ts
Oksamies: Ah, true true, seems like I missed or forgot that part of the comment/discussions earlier. I'll change these up 💯

apps/cyberstorm-nextjs/.env.development
Oksamies: 1. I left the `NEXT_` prefix out of it, so that it wouldn't be confused with `next-auth` stuff 🤔 I'd say the `CYBERSTORM_AUTH_` prefix is fitting here 🤷 

2. Fixd

Oksamies:
> The code in the initial commit is almost there, good job.
> 
> But IDK why you went and added the second commit to this PR instead of creating another PR for those changes. I'm not 100% sure but probably the PRs wouldn't even need to be chained, since the changes are in different files. Nor are they logically that related. Now change requests to the new commit are blocking the old commit.

The latter commit had to be done eventually, so I just decided to include it in this, as per the name of the PR. Can't be auth without login.

Oksamies (commented):
checked myself

apps/cyberstorm-nextjs/.env.development
anttimaki: 1. I'm referring to this: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser . (If we'd be talking about just good naming without the aforementioned functional perspective, I'd argue that prefixing everything with Cyberstorm is a bit pointless since the whole app is cyberstorm at this point, and misleading since cyberstorm initially referred only to the component library.)
2. You didn't fix the query parameter part.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
anttimaki: I'm not sure about best practices either, but ideally we have one component per file, so "Props" IMO says enough without causing confusion. Things change a bit if the props are exported since then multiple "Props" can be imported in a single file (although this can be resolved with "import x as y").

This comment can be marked as resolved.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
anttimaki: Alright. Having `import { PlainButton } from "./Button";` seems like a footgun then but that's outside the scope of this PR.

This comment can be marked as resolved.

anttimaki:
> The latter commit had to be done eventually, so I just decided to include it in this, as per the name of the PR. Can't be auth without login.

This argument fails to move me at any level.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
```diff
@@ -188,6 +188,11 @@
   line-height: var(--line-height--l);
 }
 
+.loginButton > a > div > svg {
+  width: 1rem;
```
anttimaki, line 192: I don't think this does what you assume it does. Change the values to 10rem and see what happens. Not 100% sure (grav might be the person to ask) but you might want to consider using the size prop of `FontAwesomeIcon`

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
```diff
@@ -19,30 +25,42 @@ export function BetaLoginLayout() {
...
-              <Button
-                colorScheme="overwolf"
-                label="Overwolf"
-                leftIcon={<OverwolfLogo />}
-                paddingSize="large"
-              />
-            </div>
+            {props.discordAuthUrl ? (
+              <div className={styles.loginButton}>
+                <a href={props.discordAuthUrl}>
```
anttimaki, line 30: You could add `onClick((e) => e.preventDefault))` to prevent the user from getting directed to external site. That would be in line with recent changes done by Vilppe.

anttimaki (changes requested):
Note that I consider one of my comments from an earlier review to still be unresolved.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
Oksamies: Sadly this does what I want. Tried to figure this out with the size prop, but that does nothing. The problems we've had with `FontAwesomeIcon` component should be figured out in a separate task. I think I made a task about it somewhere 🤔

Oksamies:
> > The latter commit had to be done eventually, so I just decided to include it in this, as per the name of the PR. Can't be auth without login.
> 
> This argument fails to move me at any level. Also I feel that we (the team) have had this discussion multiple times, always ending in the conclusion that smaller PRs are better.

I'm not removing the second commit from this PR, since it's part of the auth.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: True. We should have a refactor task for the button and link components somewhere. If not that needs to be figured out with grav

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: Agreed 💯

apps/cyberstorm-nextjs/.env.development
Oksamies: 1. Oh, neat, didn't know about that. Yes those should.
2. Fixd, for real now. Also made it so that the query param is added separately, since we probably at some point want to change it dynamically 👍

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: At first I got bit confused, but I'm guessing you meant that; it should be a disabled link instead of the button not being there?

Oksamies: Oh, you probably meant at the storybook end, right.

Oksamies: I am confuse. We want people to go and visit the link? If you are talking about the storybooks end; I'd have to make a new component for the "fakeLinks" and then have that have a preventDefault option as a prop, which is out of scope for now.

apps/cyberstorm-nextjs/.env.development
```diff
@@ -2,3 +2,10 @@
 # the security implications that entails.
 NEXT_PUBLIC_API_DOMAIN=http://thunderstore.localhost
 NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_DOMAIN/api/cyberstorm/
+AUTH_ENABLED=
+AUTH_SECRET=development
+NEXT_PUBLIC_AUTH_BASE_URL=https://auth.thunderstore.io/auth/login
+NEXT_PUBLIC_AUTH_QP=?next=https%3A%2F%2Fbeta.thunderstore.io%2F
```
anttimaki, line 8: `NEXT_PUBLIC_AUTH_RETURN_URL` would be clearer? Not a blocker.

apps/cyberstorm-nextjs/.env.development
```diff
@@ -2,3 +2,10 @@
...
+NEXT_PUBLIC_AUTH_DISCORD_URL=$NEXT_PUBLIC_AUTH_BASE_URL/discord/
```
anttimaki, line 9: Kinda odd to do this part of parsing in the env file but then do the return URL parsing in the component. Why not do both in the same place? Maybe here, since the URLs are defined once but possibly used in multiple places?

Not a blocker.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
anttimaki: To reduce a change there's some miscommunication...

1. Care to elaborate what it is that you want? Based on the selector I'd assume this is to change the size of the SVGs in the link buttons. But when tested (Storybook/Chrome), having this thing or removing it altogether had zero effect on the icons or the buttons (I took screenshots before and after and they look identical).
2. Changing the to e.g. 5rem as I suggested increases the size of the OW icon a bit, but nowhere near the expected 5x. GitHub/Discord icon size doesn't change at all. However, height of all the buttons do change dramatically.
3. Getting this far in writing this question I now assume you're trying to target the OW logo and not the others. And since OW logo isn't a FontAwesomeIcon like the rest so the size prop doesn't work. Is this correct? Your answer was a bit ambiguos.
4. If the above is correct it's a bit problematic since selector chooses all three icons while you want to change just one, and the height property works in unexpected way. Adding a class to e.g. OW `<a>` to target only it would resolve the first issue, and maybe the height property could just be dropped since it doesn't affect the visible size of the icon? I don't consider this a blocker though.
5. Care to link to that task you mentioned?

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
anttimaki: Alright, my bad.

Using `/` as the url prop in the story would make the link to lead to Storybooks frontpage, which while not perfect, would IMO be an improvement. Couldn't find a clever way to make the user stay on the same page (due to use of iframes) 😞. I don't consider this a blocker though.

anttimaki:
> I'm not removing the second commit from this PR, since it's part of the auth.

I didn't suggest you'd do that. (After a round of reviews that would be counterproductive.) I was commenting, that the argument which you used to brush of my earlier crituque, was IMO weak.

As is this second one. Just because they're part of the same feature doesn't in any way mean they should be one PR. What kind of feedback would I receive if I submitted a single PR for the whole "implement dapper-ts" milestone?

anttimaki (approved):
I'm not making a blocker of these but there would still IMO be room for improvement.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
Oksamies: I'm not using storybook for development right now. I've only used the dev env and in there in the login page, the two FA icons are not shown at all without this, because the parent div doesn't have a size. I could have added a wrapper element for the FA icons and used a component css class in it, but seemed a bit overkill. Anyways I guess there are some differences in the reset css of dev env and storybook.

I'll link you the task in discord.

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.tsx
Oksamies: Yeah, that makes sense, I'll change that up

packages/cyberstorm/src/components/Layout/LoginLayout/BetaLoginLayout.module.css
anttimaki: > I'm not using storybook for development right now.

IMO, since the page has a Story, you should use Storybook too. A bad analogy would be "I didn't update the unit tests because I didn't use them for development" I guess?

> the two FA icons are not shown at all without this, because the parent div doesn't have a size

Tested master branch locally, this doesn't seem to be the case. Instead, the icons consume all the width available to them, and thus result in different heights, affecting the button dimensions. Bringing this out since it might help writing a task for this and coming up with a solution.

> I guess there are some differences in the reset css of dev env and storybook

Is there? Shouldn't they import the same stuff from `cyberstorm-styles`? If there's indeed a difference, create a task about seeing if there's a real reason for it, and remove it otherwise. I'd assume having such a difference will cause similar confusion in the future too.

apps/cyberstorm-nextjs/.env.development
anttimaki: My recollections of OAuth are a bit hazy but I'm not sure that would work. Doesn't the userneed to return to a login page that does some final validations? Any "redirect to original page" would have to be handled by that page? And the return URLs need to be registered with the OAuth provider anyway, and we don't want to register all the pages in the site. (Unless some wild card approach is possible.)

PR #794, opened 2023-08-31
Linking refactor

VilppeRiskidev:


apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -3,6 +3,7 @@ import React from "react";
 
 interface CyberstormLinkProps {
```
anttimaki, line 4: Not really related to the task at hand and therefore not a blocker, but I find it somewhat confusing that we have this intermediary `CyberstormLinkProps` that's used only to define the `LinkProps`. My first thought was that some specific meaning is being conveyed with the naming but I can't figure out what it might be. Furthermore this differs from the way the `LinkLibrary` in `cyberstorm-nextjs` defines the interface, for no particular reason. I'd prefer if they'd use the same way just for consistency.

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
```diff
@@ -3,41 +3,133 @@ import React, { PropsWithChildren } from "react";
...
+type PageTitleCrumbProps = PropsWithChildren<{
+  pageTitle: string;
+}>;
 
 type BreadCrumbsProps = PropsWithChildren<{
-  excludeHome?: boolean;
+  pageTitle: string;
 }>;
 
-export function BreadCrumbs(props: BreadCrumbsProps) {
```
anttimaki, line 12: I don't like dropping the generic `BreadCrumbs` component since we really don't want to have to define a separate component whenever we add new views. Why was this change done? What problem is it solving/bypassing? (Side note: the commit messages are quite unhelpful.)

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
```diff
@@ -3,41 +3,133 @@ import React, { PropsWithChildren } from "react";
...
-        </div>
-      </div>
-    );
-  });
+interface PackageBreadCrumbsProps extends BreadCrumbsProps {
+  community: string;
+}
 
-  return <div className={styles.root}>{nodes}</div>;
+interface PackageDependantsBreadCrumbsProps {
```
anttimaki, line 26: If we need these interfaces, I think we can do with some reusability here. E.g. this could be just `PackageBreadCrumbsProps` since the three props will be shared by many package related views.

But I hope we don't need to drop the BreadCrumbs component and these interfaces won't be needed.

packages/cyberstorm/src/components/BreadCrumbs/BreadCrumbs.tsx
```diff
@@ -3,41 +3,133 @@ import React, { PropsWithChildren } from "react";
...
+export function PageTitleCrumb(props: PageTitleCrumbProps) {
+  const { pageTitle } = props;
+  return (
+    <div className={`${styles.outer} ${styles.outer__end}`}>
+      <div className={`${styles.inner} ${styles.inner__end}`}>{pageTitle}</div>
+    </div>
+  );
+}
+
+export function HomeOnlyBreadCrumbs() {
```
anttimaki, line 51: `HomeOnlyBreadCrumbs` and `TitleOnlyBreadCrumbs` *might* be worth keeping even if we don't end up creating separate BreadCrumbs components for most views.

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageDetailLayout.tsx
```diff
@@ -90,26 +85,28 @@ export function PackageDetailLayout(props: PackageDetailLayoutProps) {
   }
   if (packageData.gitHubLink) {
     packageDetailsMeta.push(
-      <Link
-        key="2"
-        externalUrl={packageData.gitHubLink}
-        label="GitHub"
-        leftIcon={<FontAwesomeIcon icon={faGithub} fixedWidth />}
-      />
+      <a key="2" href={packageData.gitHubLink}>
```
anttimaki, line 88: Buttons aren't supposed to be nested inside anchors. We probably want to keep some sort of `Link` element for links leading outside the website.

packages/cyberstorm/src/components/Layout/PackageListLayout/PackageListLayout.tsx
```diff
@@ -89,13 +84,17 @@ export function PackageListLayout(props: PackageListLayoutProps) {
...
               size="bold_large"
             />,
-            <Link
-              key="meta-link"
-              leftIcon={<FontAwesomeIcon icon={faDiscord} fixedWidth />}
-              label="Join our community"
-              externalUrl="https://discord.thunderstore.io/"
-              size="bold"
-            />,
+            <a key="meta-link" href="https://discord.thunderstore.io/">
```
anttimaki, line 87: Buttons aren't supposed to be nested inside anchors. We probably want to keep some sort of `Link` element for links leading outside the website.

packages/cyberstorm/src/components/Layout/Teams/TeamsLayout.tsx
```diff
@@ -13,7 +12,7 @@ import { PageHeader } from "../BaseLayout/PageHeader/PageHeader";
 import { useDapper } from "@thunderstore/dapper";
 import usePromise from "react-promise-suspense";
 
-//TODO: Use Alert component
```
anttimaki, line 16: The component doesn't seem to use Alert component yet, was this comment intentionally removed?

packages/cyberstorm/src/components/Link/Link.tsx
anttimaki, line 1: As mentioned in other comments, I think we'll still need a component like this one.

packages/cyberstorm/src/components/Links/LinkLibrary.tsx
anttimaki, line 1: What is this file used for? (Is it used at all?) Generally speaking the apps should define the LinkLibrary since they know how they want to render their links, so I can't remember if there's a reason for the component package to do so.

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageDetailLayout.tsx
VilppeRiskidev: PlainButton is not a <button>, but a <div>🤔  There _was_ a Link component, but that was only used in 2 places, both of which looked exactly like buttons

packages/cyberstorm/src/components/Layout/Teams/TeamsLayout.tsx
VilppeRiskidev: Alert has been added to the subcomponent, whoever did that probably forgot to remove the TODO comment here.

anttimaki:
It was decided that a different approach is needed. See the ticket TS-1803 for more information.

PR #805, opened 2023-09-12
Fix disable prop and clean up components story

Cyanideankst:
refs [TS-1816](https://app.clickup.com/t/20413243/TS-1816)

### DIFF
--- a/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
@@ -1,42 +1,27 @@
 import { StoryFn, Meta } from "@storybook/react";
 import { Switch } from "@thunderstore/cyberstorm";
-import React, { CSSProperties, useState } from "react";
+import React, { useState } from "react";
 
 export default {
   title: "Cyberstorm/Components/Switch",
   component: Switch,
 } as Meta<typeof Switch>;
 
-const backgroundStyle: CSSProperties = {
-  display: "flex",
-  flexDirection: "column",
-  gap: "var(--gap--16)",
-  backgroundColor: "var(--color-purple--4)",
-  padding: "var(--gap--16)",
+const defaultArgs = {
+  disabled: false,
 };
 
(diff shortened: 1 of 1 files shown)


apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -1,42 +1,27 @@
...
 import { Switch } from "@thunderstore/cyberstorm";
-import React, { CSSProperties, useState } from "react";
+import React, { useState } from "react";
 
 export default {
   title: "Cyberstorm/Components/Switch",
   component: Switch,
 } as Meta<typeof Switch>;
 
-const backgroundStyle: CSSProperties = {
```
anttimaki, line 10: Ideally the cleanup would've been done in a separate commit to make reviewing the "meat" of the PR easier. This is just for future reference, doesn't require action now.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -1,42 +1,27 @@
...
-  display: "flex",
-  flexDirection: "column",
-  gap: "var(--gap--16)",
-  backgroundColor: "var(--color-purple--4)",
-  padding: "var(--gap--16)",
+const defaultArgs = {
+  disabled: false,
```
anttimaki, line 11: `false` is the default value for this prop, so strictly speaking this isn't required. I guess it doesn't hurt either. It's just that not including it would make the change smaller since now this is passed to components for no reason.

anttimaki (approved):
A minor comment but can be merged as is. Also note that the commit message is missing the "refs" part.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
Cyanideankst: My first intention was to do it without passing values here, but couldn't find way to do this without changing it to use defaultArgs and because of that couldn't just pass empty variable either. 
More than happy to hear out what would have been correct way to do it!

PR #806, opened 2023-09-13
Prevent link default actions in cyberstorm storybook

VilppeRiskidev:
Ref: TS-1820

### DIFF
--- a/apps/cyberstorm-storybook/LinkLibrary.tsx
+++ b/apps/cyberstorm-storybook/LinkLibrary.tsx
@@ -14,8 +14,17 @@ const Link = (props: LinkProps): React.ReactElement => {
   const { children, queryParams, url } = props;
   const q = queryParams ? `?${queryParams}` : "";
 
-  return <a href={`${url}${q}`}>{children}</a>;
+  return (
+    <a onClick={(event) => linkOnClick(event)} href={`${url}${q}`}>
+      {children}
+    </a>
+  );
 };
+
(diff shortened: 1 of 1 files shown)


apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -14,9 +14,18 @@ const Link = (props: LinkProps): React.ReactElement => {
   const { children, queryParams, url } = props;
   const q = queryParams ? `?${queryParams}` : "";
 
-  return <a href={`${url}${q}`}>{children}</a>;
+  return (
+    <a onClick={(event) => linkOnClick(event)} href={`${url}${q}`}>
```
anttimaki, line 18: `onClick={linkOnClick}` would amount to same behaviour without the wrapping function.

apps/cyberstorm-storybook/LinkLibrary.tsx
```diff
@@ -14,9 +14,18 @@ const Link = (props: LinkProps): React.ReactElement => {
...
+      {children}
+    </a>
+  );
 };
 
+function linkOnClick(event: React.MouseEvent<HTMLAnchorElement>) {
+  event.preventDefault();
+  console.log("Links are disabled in Storybook");
```
anttimaki, line 26: IDK if this is more clutter than helpful?

anttimaki (approved):
Couple of opinion comments but I'm okay with this being merged as is.

PR #808, opened 2023-09-14
@thunderstore/cyberstorm-storybook: clean up Switch story file

anttimaki:
Passing around defaultArgs was pointless since it only contained one optional prop and the value was set to component's default value.

Refs TS-1815

### DIFF
--- a/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
@@ -7,21 +7,17 @@ export default {
   component: Switch,
 } as Meta<typeof Switch>;
 
-const defaultArgs = {
-  disabled: false,
-};
-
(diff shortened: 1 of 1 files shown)


anttimaki:
PR to address this https://github.com/thunderstore-io/thunderstore-ui/pull/805#discussion_r1325044141 , i.e. these changes are more about learning than actual need for changes.

@Cyanideankst :

- `DisabledSwitch.args` doesn't need `defaultArgs`, since we end up overwriting the only arg it provides with `disabled: true`
- `SameStateSwitch.args` doesn't need it either, since `disabled` is optional prop for `Switch` and the default value is `false`
- `Since `SameStateSwitch.args` is now empty, there's no need to pass args to `SameStateSwitchTemplate` function

MythicManiac (approved):
LGTM

anttimaki:
@Cyanideankst have you reviewed this already?

Cyanideankst (approved):
Yup, got your point! 👍

PR #816, opened 2023-09-19
Fix storybook imports

Cyanideankst:
[refs TS-1817](https://app.clickup.com/t/20413243/TS-1817)

### DIFF
--- a/apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
@@ -2,10 +2,10 @@ import { StoryFn, Meta } from "@storybook/react";
 import { Pagination } from "@thunderstore/cyberstorm";
 import React, { useState } from "react";
 
-export default {
+const meta = {
   title: "Cyberstorm/Components/Pagination",
   component: Pagination,
-} as Meta;
+} as Meta<typeof Pagination>;
 
 const defaultArgs = {
   currentPage: 1,
@@ -54,6 +54,7 @@ EmptyPagination.args = {
 };
 
 export {
+  meta as default,
   ReferencePagination,
   DefaultPagePagination,
   DisabledPagination,
--- a/apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
@@ -1,11 +1,11 @@
-import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { StoryFn, Meta } from "@storybook/react";
 import { Pagination } from "@thunderstore/cyberstorm";
 import React, { useState } from "react";
 
-const meta = {
+export default {
   title: "Cyberstorm/Components/Pagination",
   component: Pagination,
-} as ComponentMeta<typeof Pagination>;
+} as Meta;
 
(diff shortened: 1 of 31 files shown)


apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
anttimaki, line 1: If further changes are made, the comments made here naturally apply to all the other files too.

apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
```diff
@@ -1,11 +1,11 @@
-import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { StoryFn, Meta } from "@storybook/react";
 import { Pagination } from "@thunderstore/cyberstorm";
 import React, { useState } from "react";
 
-const meta = {
+export default {
```
anttimaki, line 5: Is there any specific reason to change the default export to happen here? For me it would be clearer if all the exports happened in one place, at the end of the file.

apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
```diff
@@ -1,11 +1,11 @@
...
   title: "Cyberstorm/Components/Pagination",
   component: Pagination,
-} as ComponentMeta<typeof Pagination>;
+} as Meta;
```
anttimaki, line 8: Why are we dropping the `<typeof Pagination>` part here?

anttimaki (commented):
Were you following some guide or example while doing the changes? I'm trying to understand the reasons behind the changes I commented on.

The PR works and fulfills the DoD of the task, so no changes are absolutely necessary, but I'd like to get some answers to the questions before approving.

Also I checked the Storybook documentation and it seems there's a new(?) recommended way of writing the stories that differs a bit from what we're using. But I guess adjusting to use that is outside the scope of this task.

apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
Cyanideankst: As you questioned above, I did use working example as source of how it could be done and after checking some storybooks change logs made assumption that it wasn't needed there anymore. I admit that it might have been hasty decision since CSFv3 introduced new methods to use objects instead of functions to represent stories.
Should I add them back since it wouldn't be that huge work and might save time on future migration?

apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
Cyanideankst: Not that well founded reasoning. Did use component that was up to date on how imports & exports should be dealt with and also storybook forum post used example done like this so thought it would be best practices.
After further reading about how CSFv3 would work and readability I would also second to revert this change.

apps/cyberstorm-storybook/stories/components/Pagination.stories.tsx
anttimaki: Well if you're going to revert the `export` changes, might do these as well 👍🏻

anttimaki (approved):
LGTM 👍🏻

Oksamies (approved):

PR #818, opened 2023-09-20
@thunderstore/thunderstore-api: flesh out the implementation

anttimaki:
Improve the helpers used in all future fetch methods and use the
community list fetcher as an model implementation.


packages/thunderstore-api/src/__tests__/typeguards.ts
```diff
@@ -0,0 +1,30 @@
+import { isNumberArray, isStringArray } from "../typeguards";
+
+describe("isNumberArray", () => {
+  it.each([
+    { value: [], expected: true },
+    { value: [-1, 0, 1], expected: true },
+    { value: [-1, "0", 1], expected: false },
+    { value: [Math.PI], expected: true },
+    { value: [NaN], expected: false },
+    { value: 1, expected: false },
```
MythicManiac, line 10: Seems to miss a test case for infinity, which IIRC is considered a number and has to be checked for explicitly (if we need to do that).

Not a blocker for this PR, address separately if you feel it's critical

MythicManiac (approved):
One minor comment, looks good overall

packages/thunderstore-api/src/__tests__/typeguards.ts
anttimaki: @MythicManiac care to elaborate what you mean by "has to be checked for explicitly"? The current implementation seems to pass fine when tested with `Infinity`, classifying it as a `number`.

PR #819, opened 2023-09-20
Fix the current-user fetchs' header

Oksamies:
refs TS-1819


Oksamies:
These changes are not needed, they were a mistake.

PR #822, opened 2023-09-21
Change /communities to be the landing page

Oksamies:
Change links to landing page to be to community list view
Redirect /communities to / when enabled

 refs TS-1828

### DIFF
--- a/apps/cyberstorm-nextjs/middleware.ts
+++ b/apps/cyberstorm-nextjs/middleware.ts
@@ -24,6 +24,10 @@ export async function middleware(request: NextRequest) {
   const req = request;
   const response = NextResponse.next();
 
+  if (request.nextUrl.pathname === "/communities") {
+    return NextResponse.redirect(new URL("/", request.url));
+  }
+
   // For the sake of keeping dev configuration easy and not confusing
   if (!process.env.AUTH_ENABLED) {
     return response;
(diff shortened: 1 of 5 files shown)


apps/cyberstorm-nextjs/.env.development
```diff
@@ -1,5 +1,6 @@
 # Values prefixed with NEXT_PUBLIC_ are available in browser, with all
 # the security implications that entails.
+NEXT_PUBLIC_PRODUCTION_PAGE_STRUCTURE=
```
anttimaki, line 3: I second what Mythic mentioned in Discord, I think this is overkill.

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -24,6 +24,12 @@ export async function middleware(request: NextRequest) {
   const req = request;
   const response = NextResponse.next();
 
+  if (process.env.NEXT_PUBLIC_PRODUCTION_PAGE_STRUCTURE) {
+    if (request.nextUrl.pathname.startsWith("/communities")) {
```
anttimaki, line 28: This blocks the access to all subpages under `/communities`. Sure, we don't have them yet, but wouldn't hurt to use exact matching here either.

anttimaki (changes requested):
Looks thorough, e.g. checking the footer and links which I forgot to mention in the task 👍🏻

anttimaki (approved):
The changes are good but the commit message is now blatantly wrong. Feel free to merge without another review once that's fixed.

Oksamies:
Ah, nice catch, didn't even realize that

PR #830, opened 2023-09-26
Add @thunderstore/use-promise

anttimaki:


anttimaki:
> I'll leave you with a design consideration:

Perhaps I'll ponder on this later, but for now I want this thing merged to fix the CI. The process is quite long as it is and have it fail on the last step as a false positive is highly annoying. Not mention reruns wasting our resources that have quota limits.

PR #833, opened 2023-09-27
Improve GitHub Actions Workflows

MythicManiac:
- Move the chromatic build to happen only after tests pass to avoid spamming the API (and thus consuming our quota)
- Fix yarn cache


anttimaki (approved):
LGTM

PR #839, opened 2023-10-03
@thunderstore/dapper-ts: implement getCurrentUser

anttimaki:


anttimaki:
Related backend PR: https://github.com/thunderstore-io/Thunderstore/pull/884. Test run fails because those changes haven't been deployed to test environment yet.

packages/thunderstore-api/src/fetch/__tests__/currentUser.ts
```diff
@@ -0,0 +1,13 @@
+import { fetchCurrentUser } from "../currentUser";
+
+it("receives emptyish object when querying without session id", async () => {
+  const config = { apiHost: "https://thunderstore.dev" };
```
MythicManiac, line 4: Out of scope for this PR but it'd probably be smart to import the domain used for testing from a common location / make it configurable somehow. Can then later on be replaced with a docker setup that pulls the latest TS image & tests locally.

MythicManiac (approved):
LGTM

PR #842, opened 2023-10-05
Switch component touchable label element

Cyanideankst:
[Refs TS-1840](https://app.clickup.com/t/20413243/TS-1840)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,6 +28,15 @@
   "rules": {
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
+    "jsx-a11y/label-has-associated-control": [
+      2,
+      {
+        "labelComponents": ["CustomInputLabel"],
+        "labelAttributes": ["label"],
+        "controlComponents": ["CustomInput"],
+        "depth": 3
+      }
+    ],
     "@typescript-eslint/no-unused-vars": [
       "error",
       { "ignoreRestSiblings": true }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,6 +31,7 @@
     "jsx-a11y/label-has-associated-control": [
       2,
       {
+        "controlComponents": ["Switch"],
         "depth": 3
       }
     ],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -31,9 +31,6 @@
     "jsx-a11y/label-has-associated-control": [
       2,
       {
-        "labelComponents": ["CustomInputLabel"],
-        "labelAttributes": ["label"],
-        "controlComponents": ["CustomInput"],
         "depth": 3
       }
     ],
--- a/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
@@ -22,7 +22,7 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
     </div>
   );
 };
-const SameStateSwitchTemplate2: StoryFn<typeof Switch> = () => {
+const SwitchLabelIdTemplate: StoryFn<typeof Switch> = () => {
   const [state, setState] = useState(false);
   return (
     <div>
@@ -30,18 +30,15 @@ const SameStateSwitchTemplate2: StoryFn<typeof Switch> = () => {
       <div>
         <Switch state={state} onChange={setState} labelId="my-switch" />
       </div>
-      <div>
-        <Switch state={state} onChange={setState} labelId="my-switch" />
-      </div>
     </div>
   );
 };
-const SameStateSwitchTemplate1: StoryFn<typeof Switch> = () => {
+const SwitchLabelWithoutNestingTemplate: StoryFn<typeof Switch> = () => {
   const [state, setState] = useState(false);
   return (
     <div>
       <label htmlFor="my-switch">
-        My itch
+        My switch
         <Switch state={state} onChange={setState} />
       </label>
     </div>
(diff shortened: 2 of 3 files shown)


.eslintrc.json
```diff
@@ -28,6 +28,15 @@
   "rules": {
     "react/react-in-jsx-scope": "off",
     "react/prop-types": "off",
+    "jsx-a11y/label-has-associated-control": [
+      2,
+      {
+        "labelComponents": ["CustomInputLabel"],
```
anttimaki, line 34: Per the [documentation](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md): _"`labelComponents` is a list of custom React Component names that should be checked for an associated control."_

We don't have any custom label components currently, especially any named "CustomInputLabel", so this row is not needed.

.eslintrc.json
```diff
@@ -28,6 +28,15 @@
...
+        "labelAttributes": ["label"],
```
anttimaki, line 35: _"`labelAttributes` is a list of attributes to check on the label component and its children for a label."_

Since we don't have those custom label components, this row isn't needed either.

.eslintrc.json
```diff
@@ -28,6 +28,15 @@
...
+        "controlComponents": ["CustomInput"],
```
anttimaki, line 36: _"`controlComponents` is a list of custom React Components names that will output an input element."_

We don't have a component named "CustomInput". We do have a "Switch" though. And maybe other components like "TextInput" as well, but feel free to consider them outside the scope of this PR.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -22,10 +22,41 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
     </div>
   );
 };
+const SwitchLabelIdTemplate: StoryFn<typeof Switch> = () => {
+  const [state, setState] = useState(false);
+  return (
+    <div>
+      <label htmlFor="my-switch">My switch</label>
+      <div>
```
anttimaki, line 30: This `div` doesn't really do anything so it's not needed.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -22,10 +22,41 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
...
+        <Switch state={state} onChange={setState} labelId="my-switch" />
+      </div>
+    </div>
+  );
+};
+const SwitchLabelWithoutNestingTemplate: StoryFn<typeof Switch> = () => {
```
anttimaki, line 36: Why is this named "SwitchLabel**Without**NestingTemplate" when it indeed does use nesting?

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -22,10 +22,41 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
...
+  const [state, setState] = useState(false);
+  return (
+    <div>
```
anttimaki, line 39: Not needed.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -22,10 +22,41 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
...
+      <label htmlFor="my-switch">
```
anttimaki, line 40: The `htmlFor` shouldn't be used if there's no component with the matching `id` present. This is why this example doesn't work, browsers probably assumes it should look for an input with id "my-switch", not the one nested inside the label.

So the `htmlFor` attribute should be removed. Removing it would also reveal that what you attempted with ESLint config did not work.

packages/cyberstorm/src/components/Switch/Switch.tsx
```diff
@@ -8,16 +8,18 @@ export interface SwitchProps {
   state: boolean;
   onChange?: Dispatch<SetStateAction<boolean>>;
   disabled?: boolean;
+  labelId?: string;
```
anttimaki, line 11: This should be named just `id`. It has other uses besides just coupling a label element. Also it's what developers would expect to use when wanting to add in id to an element.

anttimaki (changes requested):
Overall the direction is right. The comments should be rather quick to fix now, I think.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
Cyanideankst: Just brainfart while thinking describing name. Will change name to have opposite meaning 👍

packages/cyberstorm/src/components/Switch/Switch.tsx
Cyanideankst: I was pondering if here should be specified what id it refers to. Thanks for clarifying it 👌

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
```diff
@@ -22,10 +22,41 @@ const SameStateSwitchTemplate: StoryFn<typeof Switch> = () => {
...
+      <div>
+        <Switch state={state} onChange={setState} Id="my-switch" />
+      </div>
+    </>
+  );
+};
+const SwitchNestedLabelTemplate: StoryFn<typeof Switch> = () => {
+  const [state, setState] = useState(false);
+  return (
+    <>
```
anttimaki, line 39: This `<>` shouldn't be needed either, since label is the only child.

packages/cyberstorm/src/components/Switch/Switch.tsx
```diff
@@ -8,16 +8,18 @@ export interface SwitchProps {
   state: boolean;
   onChange?: Dispatch<SetStateAction<boolean>>;
   disabled?: boolean;
+  Id?: string;
```
anttimaki, line 11: The props shouldn't be capitalized, at least if their just strings. Use `id` instead.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
anttimaki: The earlier comment hasn't been addressed?

anttimaki (changes requested):
Looks good but a hew stylistic things to do still.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
Cyanideankst: 👍 Yup fixed it. Didn't remember to question my choice here after I got this story working as intended.

apps/cyberstorm-storybook/stories/components/Switch.stories.tsx
Cyanideankst: No it's!

anttimaki (approved):
Nice! Use ___Squash___ merge when merging this.

PR #858, opened 2023-10-12
Fix FilterItem styles

Oksamies:
refs TS-1841

### DIFF
--- a/packages/cyberstorm/src/components/FilterItemList/FilterItemList.tsx
+++ b/packages/cyberstorm/src/components/FilterItemList/FilterItemList.tsx
@@ -62,10 +62,10 @@ export function FilterItemList(props: FilterItemListProps) {
   });
 
   return (
-    <>
+    <div className={styles.root}>
       <div className={styles.filterListHeader}>Categories</div>
-      {filters}
-    </>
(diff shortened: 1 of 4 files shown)


packages/cyberstorm/src/components/FilterItemList/FilterItemList.tsx
```diff
@@ -62,10 +62,10 @@ export function FilterItemList(props: FilterItemListProps) {
   });
 
   return (
-    <>
+    <div className={styles.root}>
       <div className={styles.filterListHeader}>Categories</div>
-      {filters}
-    </>
+      <div>{filters}</div>
```
gkurck, line 67: I'm generally not a fan of empty divs. Is there a function for having this?

gkurck (commented):
Added some suggestions, use your discretion on what to address now and what to leave on the refactoring round.

packages/cyberstorm/src/components/FilterItemList/FilterItemList.tsx
Oksamies: Structure for the gap between "Categories" word and the list of filters.

PR #868, opened 2023-10-17
add tooltip to button component

Cyanideankst:
[Refs TS-1788](https://app.clickup.com/t/20413243/TS-1788)

### DIFF
--- a/packages/cyberstorm/src/components/Button/Button.tsx
+++ b/packages/cyberstorm/src/components/Button/Button.tsx
@@ -43,9 +43,22 @@ export interface ButtonProps {
   tooltipText?: string;
 }
 
+const TooltipWrapper = (props: TooltipWrapperProps) =>
+  props.tooltipText ? (
+    <Tooltip content={props.tooltipText} side="bottom">
+      {props.children}
+    </Tooltip>
+  ) : (
+    <>{props.children}</>
+  );
+interface TooltipWrapperProps extends PropsWithChildren {
+  tooltipText?: string;
+}
+
 /**
  * Cyberstorm Button component
  */
+
 const Button = React.forwardRef<
   HTMLButtonElement | HTMLDivElement,
   ButtonProps
@@ -62,18 +75,6 @@ const Button = React.forwardRef<
     ...forwardedProps
   } = props;
 
-  interface TooltipWrapperProps extends PropsWithChildren {
-    tooltipText?: string;
-  }
-
-  const TooltipWrapper = (props: TooltipWrapperProps) =>
-    props.tooltipText ? (
-      <Tooltip content={props.tooltipText} side="bottom">
-        {props.children}
-      </Tooltip>
-    ) : (
-      <>{props.children}</>
-    );
   const fallbackRef = useRef(null);
 
   if (plain) {
--- a/packages/cyberstorm/src/components/Button/Button.tsx
+++ b/packages/cyberstorm/src/components/Button/Button.tsx
@@ -6,7 +6,6 @@ import React, {
   useRef,
 } from "react";
 import styles from "./Button.module.css";
-import { TooltipProvider } from "@radix-ui/react-tooltip";
 import { Tooltip } from "../Tooltip/Tooltip";
 
 export interface ButtonProps {
@@ -54,7 +53,7 @@ const Button = React.forwardRef<
   const {
     children,
     plain = false,
-    type = "button",
+    type,
     colorScheme = "default",
     onClick,
     paddingSize = "medium",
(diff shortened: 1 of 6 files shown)


packages/cyberstorm/src/components/Button/Button.tsx
```diff
@@ -79,17 +83,21 @@ const Button = React.forwardRef<
...
-        ref={ref}
-        type={type}
-        className={`${styles.root} ${getIconAlignment(
-          iconAlignment
-        )} ${getStyle(colorScheme)} ${getPaddingSize(paddingSize)}`}
-        onClick={onClick}
-      >
-        {children}
-      </button>
+      <TooltipProvider>
```
Oksamies, line 86: `<TooltipProvider>` is defined in `packages\cyberstorm\src\components\Providers.tsx`, so this wrapper can be removed.

packages/dapper/src/context.tsx
Oksamies, line 1: Don't commit these changes. These were just temporary for your local dev env, so that dapper fake is used.

packages/cyberstorm/src/components/Button/Button.tsx
```diff
@@ -79,17 +83,21 @@ const Button = React.forwardRef<
     const fRef = forwardedRef as React.ForwardedRef<HTMLButtonElement>;
```
Oksamies, line 83: There should be 4 different outcomes. 1. Plain button without tooltip 2. Button (non-plain) without tooltip 3. Plain button with tooltip 4. Button (non-plain) with tooltip. You can check if the `tooltipText` is `null` `if (tooltipText) {` and then put the `if (plain)` checks inside of the `tooltipText` if check and the following `else` block.

anttimaki: To take this idea further, the code duplication in if-else-blocks can be reduced with something like this: 

```jsx
interface TooltipWrapperProps extends PropsWithChildren {
  tooltipText?: string;
}

const TooltipWrapper = (props: TooltipWrapperProps) =>
  props.tooltipText ? (
    <Tooltip content={props.tooltipText} side="bottom">
      {props.children}
    </Tooltip>
  ) : (
    <>{props.children}</>
  );
```

and then using it:

```jsx
  if (plain) {
    const fRef = forwardedRef as React.ForwardedRef<HTMLDivElement>;
    const ref = fRef || fallbackRef;
    return (
      <TooltipWrapper tooltipText={type}>
        <div
          {...forwardedProps}
          ref={ref}
          className={`${styles.root} ${getIconAlignment(
            iconAlignment
          )} ${getStyle(colorScheme)} ${getPaddingSize(paddingSize)}`}
        >
          {children}
        </div>
      </TooltipWrapper>
    );
  } else {
    const fRef = forwardedRef as React.ForwardedRef<HTMLButtonElement>;
    const ref = fRef || fallbackRef;
    return (
      <TooltipWrapper tooltipText={type}>
        <button
          {...forwardedProps}
          ref={ref}
          type={type}
          className={`${styles.root} ${getIconAlignment(
            iconAlignment
          )} ${getStyle(colorScheme)} ${getPaddingSize(paddingSize)}`}
          onClick={onClick}
        >
          {children}
        </button>
      </TooltipWrapper>
    );
  }
```

As I'm just yelling from the sidelines, use your own judgement if this approach makes sense here or if it's overengineering or whatever. It would probably be too daring (and outside the scope) to use the same conditional rendering approach to regular/plain button parts?

packages/dapper/src/context.tsx
Cyanideankst: Ok, thought so too but wouldn't it be ok to push them on temp commit while I need to change to different PC?

packages/cyberstorm/src/components/Button/Button.tsx
Cyanideankst: Thanks! If-else stacking was most likely way I would have approached this and even it felt uncomfortable way to do so finding this way to do it would have taken bit more time.. 
Sadly overengineering is like second habit of mine, so I'll let @Oksamies make executive decision on this one!

packages/cyberstorm/src/components/Button/Button.tsx
Cyanideankst: Oh, ok. It exports delay when we call tooltip or do I still need to trigger it somehow?

packages/dapper/src/context.tsx
Oksamies: It's fine to push them in a temp commit, while the PR is a draft. It was a draft when I commented this, just wanted to make sure these don't slip by 👍

packages/cyberstorm/src/components/Button/Button.tsx
Oksamies: Nope, it's fine with just the Tooltip in use in the Button component. Like I said before, the provider is defined in the providers file and has to be initiated only once in the project (project either NextJS or storybook). If you feel a bit unclear after this comment; reading up on how providers work will clarify things up 👍

PR #896, opened 2023-11-08
@thunderstore/dapper: augment GetPackageListings function signature

anttimaki:
@thunderstore/cyberstorm: remove unsupported props from PackageList

For the time being PackageList, and thus also PackageSearch, will
support only showing community or community+namespace scoped results.
Remove props that are meant for non-supported listings since they're
just adding noice currently. They can be readded if/when we add
related support to @thunderstore/dapper.

Refs TS-1875

@thunderstore/dapper: strongly type the scope of GetPackageListings

THe scope refers to the factors limiting the base result set when no
filters etc. are in play. E.g. the listing can be scoped to packages
listed in a specific community.

In attempt to be reusable, the earlier interface included a bunch of id
strings, e.g. community identifier and namespace name. All of these
were optional and there was no plan for which part of the code base
should be responsible to parsing the combination of these values to the
actual scope.

My initial idea was to do that parsing on the PackageList component,
but in the end I chose to do this at the top the chain in Layout
components. The rationale being that doing the parsing in middle of the
chain in unnecessary and a possible source of errors. It's better to
bypass it by requiring the Layouts to be explicit about the listing
from the get go.

It might seem that this couples Cyberstorm components and Dapper more
tightly together, but I'd argue the coupling was there already with the
old implementation. The changes in this commit only make that coupling
explicit.

Refs TS-1875

@thunderstore/dapper: add filter support for GetPackageListings

The method now accepts parameters used for filtering, ordering and
paginating the results.

PackageList needs to use deferred values to separate the light "update
UI" operation from the heavy "fetch results" operation. In practice
searchQuery is already debounced and wouldn't require deferring, but
for clarity's sake it's also deferred.

Refs TS-1875

@thunderstore/cyberstorm: reset page when PackageList filters change

Also fix the hard coded page number on PackageCount display.

Refs TS-1875

### DIFF
--- a/packages/cyberstorm/src/components/Layout/CommunityProfileLayout/CommunityProfileLayout.tsx
+++ b/packages/cyberstorm/src/components/Layout/CommunityProfileLayout/CommunityProfileLayout.tsx
@@ -29,6 +29,8 @@ export function CommunityProfileLayout(props: Props) {
   const community = usePromise(dapper.getCommunity, [communityId]);
   const filters = usePromise(dapper.getCommunityFilters, [communityId]);
 
+  const listingType = { kind: "community" as const, communityId };
+
   return (
     <BaseLayout
       backGroundImageSource={
@@ -84,7 +86,7 @@ export function CommunityProfileLayout(props: Props) {
       }
       mainContent={
         <PackageSearch
-          communityId={communityId}
+          listingType={listingType}
           packageCategories={filters.package_categories}
           sections={filters.sections}
         />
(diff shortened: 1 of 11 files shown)


anttimaki:
* master
  * **PR #896** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/896" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/896?utm_source=stack-comment).

Oksamies (approved):

PR #906, opened 2023-11-14
Add Toast

Oksamies:
refs TS-1939


packages/cyberstorm/src/components/Providers.tsx
```diff
@@ -1,15 +1,18 @@
...
 interface CyberstormProvidersProps {
   children: ReactNode | ReactNode[];
 }
 
 export function CyberstormProviders(props: CyberstormProvidersProps) {
   return (
-    <RadixTooltip.Provider delayDuration={80}>
-      {props.children}
-    </RadixTooltip.Provider>
+    <ToastProvider>
```
anttimaki, line 12: Could be handy if `CyberstormProviders` could accept toast's duration and tooltip's delayDuration as props. Now we are hardcoded to use one values in all our projects.

Not a blocker.

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  display: flex;
+  flex-direction: column;
+  box-shadow: 0 12px 24px 5px rgb(0 0 0 / 0.65);
+}
+
+.contentWrapper {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  border-left: 3px solid var(--feature-color);
```
anttimaki, line 12: I'm not a/the designer but I wonder if the 1px difference is significant enough to warrant not using `--border-radius--2` or `--border-radius--4`?

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  border-radius: var(--border-radius--4);
+  background-color: var(--background-color);
+
+  --background-color: var(--alert-info-bg-color);
```
anttimaki, line 16: Does the naming `--alert-` cause possible problems, i.e. someone editing the values assuming it affects only the alert component? Dunno what would be a better name, just pointing out that I think it's usually frowned upon to bind the names to specific components and then reusing them in other components.

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  --text-color: var(--alert-info-text-color);
+}
+
+.content {
+  display: flex;
+  gap: var(--space--16);
+  align-items: flex-start;
+  align-self: stretch;
+  width: 100%;
+  padding: 1rem 3.25rem calc(1rem - 0.125rem) 1.5rem;
```
anttimaki, line 27: Why use `calc` here? It would be easier to read if it would just have the precalculated value.

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+.closeIconWrapper {
+  position: absolute;
+  top: 0.5rem;
+  right: 0.5rem;
+  background-color: transparent;
+}
+
+.closeIcon {
+  display: flex;
+  align-items: center;
```
anttimaki, line 39: I thought the Icon component was supposed to handle the centering and width/height?

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  background-color: var(--feature-color);
+  animation: progress-bar var(--bar-timer) linear forwards;
+}
+
+/* Styles for Toast's text */
+.message {
+  display: flex;
+  flex: 1 0 0;
+  min-height: calc(1rem * var(--line-height--m));
+  color: var(--text-color);
```
anttimaki, line 63: Not the expert on this either, but the color and the three rows below it seem to use what I would assume are the standard values? Do we need to explicitly define them here? Or is it our "best practice" to always define them, even if they aren't actually needed?

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  100% {
+    width: 0%;
+  }
+}
+
+/* CSS for the Toast container */
+.toastContainer {
+  --viewport-padding: 25px;
```
anttimaki, line 113: Is using pixels here (as well as in `gap` and `width`) a good idea? Wouldn't rem be safer with no downsides?

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  z-index: 2147483647;
```
anttimaki, line 118: 🙄 Wouldn't probably hurt to at least to try to have a predefined system for z-indexes, although it will quite easily erode if people don't know about it and throw in whatever values in their components. Fixing these issues is annoying if the code base uses a bunch of different values.

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+.root[data-state="open"] {
+  animation: slide-in 600ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.root[data-state="closed"] {
+  animation: hide 200ms ease-in;
+}
+
+.root[data-swipe="move"] {
+  transform: translateX(var(--radix-toast-swipe-move-x));
```
anttimaki, line 141: Where is this defined?

packages/cyberstorm/src/components/Toast/Toast.module.css
```diff
@@ -0,0 +1,181 @@
...
+  transition: transform 200ms ease-out;
+}
+
+.root[data-swipe="end"] {
+  animation: swipe-out 600ms ease-out;
+}
+
+@keyframes hide {
+  from {
+    opacity: 1;
```
anttimaki, line 155: Not pretending to know what the `@keyframes` do here, but is changing opacity enough?, or should it be paired with `display: block`/`display: none`? IIRC opacity just hides the element, but doesn't remove it from the document flow. That would mean that if we have multiple active toasts and one is hidden, it would still occupy the space where it's, preventing other toasts from moving into it's place.

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
+"use client";
```
anttimaki, line 1: Is this required?

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+import styles from "./Toast.module.css";
+import { Icon } from "../Icon/Icon";
+import { classnames } from "../../utils/utils";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faXmarkLarge } from "@fortawesome/pro-solid-svg-icons";
+import * as RadixToast from "@radix-ui/react-toast";
+
+type _ToastProps = {
+  id: string;
+  content: ReactNode;
```
anttimaki, line 12: This could be `message` so it would match the terms used by Radix. Also are we expecting to ever pass anything but strings here?

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+  icon?: JSX.Element;
```
anttimaki, line 13: Looking at the designs, I would remove this prop and infer the icon internally from the `variant` prop. It would make the usage of the component easier/cleaner. And while it would limit us to predefined icons, that could also be seen as plus since it keeps the UI consistent.

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+  variant?: "info" | "danger" | "warning" | "success";
+  timer?: number;
```
anttimaki, line 15: This could be `duration` so it would match the terms used by Radix (also IMO it's a better name anyway, I wasn't sure what timer referred to without checking the code).

Also does override the `duration` prop in the provider? The one which is currently hardcoded to 10000. What happens if I pass a conflicting value here?

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+export type ToastProps = _ToastProps &
```
anttimaki, line 17: Is there a need to export these?

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+  Omit<React.HTMLProps<HTMLDivElement>, keyof _ToastProps>;
```
anttimaki, line 18: This comment is outside the scope of the PR since the approach used here is copy-pasted all over the code base already. I was planning to address this myself but that was a month ago and it seems I can't find the time for it. Maybe you can add this to your TODO list?

Anyways, found this article, which in the collapsible "Why not..." box describes the shortcomings of this approach (and the main article offers a preferred alternative): https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#maybe-reactjsxintrinsicelements-or-elementhtmlattributes

packages/cyberstorm/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <RadixToast.Description asChild>
+            <div className={styles.content}>
+              <Icon wrapperClasses={styles.icon}>{icon}</Icon>
+              <div className={styles.message}>{content}</div>
+            </div>
+          </RadixToast.Description>
+          <div className={styles.toastProgress}>
+            <div className={styles.toastProgressBar} style={timerCSS} />
+          </div>
+          <RadixToast.Close className={styles.closeIconWrapper}>
```
anttimaki, line 39: How about [them aria-labels](https://www.radix-ui.com/primitives/docs/components/toast#close-icon-button)?

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
```
anttimaki, line 1: Is this required?

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Toast } from "./Toast";
+import styles from "./Toast.module.css";
+import * as RadixToast from "@radix-ui/react-toast";
+
+export function ToastContainer(props: {
```
anttimaki, line 6: I would argue argue we could as well call this ToastViewport to match Radix terminology. Why use different terms for the same thing? Applies as well to `styles.viewport` instead of `styles.toastContainer`,

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+  toasts: {
+    id: string;
```
anttimaki, line 8: The comments made in `Toast.tsx` regarding the props and their naming obviously apply in here and in `ToastComplex`. In fact defining the interface once and importing it to where it's actually needed would be a valid reason to export it in the first place (currently it's exported but never used anywhere). Note that we can export it internally within `@thunderstore/cyberstorm`, without exporting it outside of the package.

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
```diff
@@ -0,0 +1,35 @@
...
+    timer?: number;
+  }[];
+}) {
+  const { toasts } = props;
+
+  return (
+    <RadixToast.Viewport asChild>
+      <div className={styles.toastContainer}>
+        {toasts.map((toast) => (
+          <Toast
```
anttimaki, line 21: If the prop interfaces would match, this could be reduced to `<Toast key={toast.id} {...toast} />`

packages/cyberstorm/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,98 @@
+import { createContext, PropsWithChildren, useReducer } from "react";
+import { ToastContainer } from "./ToastContainer";
+import { v4 as uuid } from "uuid";
+import * as RadixToast from "@radix-ui/react-toast";
+
+const initState: {
+  toasts: {
```
anttimaki, line 7: Don't copy paste this same interface one million times. Define it once.

packages/cyberstorm/src/components/Toast/ToastContext.tsx
```diff
@@ -0,0 +1,98 @@
...
+    id: string;
+    variant?: "info" | "danger" | "warning" | "success";
+    icon?: JSX.Element;
+    message?: string;
```
anttimaki, line 11: Why does the message change here from `ReactNode` to `string` and from required to optional?

packages/cyberstorm/src/index.ts
```diff
@@ -84,6 +84,7 @@ export {
 export { Title, type TitleProps } from "./components/Title/Title";
 export { Tooltip, type TooltipProps } from "./components/Tooltip/Tooltip";
 export { Alert, type AlertProps } from "./components/Alert/Alert";
+export { Toast, type ToastProps } from "./components/Toast/Toast";
```
anttimaki, line 87: As mentioned in the other comment's I would refrain from exporting the interface outside the package if it isn't absolutely needed.

anttimaki (approved):
No hard blockers **BUT I'd strongly advice to at least fix the interface issues** since changing them is more work once the component is in use.

I would also like to see Storybook stories for these components.

packages/cyberstorm/src/components/Toast/ToastContext.tsx
Oksamies: Mishap

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
Oksamies: Changed all imports to use one type definition

packages/cyberstorm/src/components/Toast/ToastContainer.tsx
Oksamies: Not 100% sure, but shouldn't be

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: Ah, yep. Added it

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: Yep https://www.radix-ui.com/primitives/docs/components/toast#custom-duration

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: I think so, because of the usage of Radix. I'll try without it, while fixing #908

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: IIRC, Radix removes the element after a short delay. But to make sure I'll test it while fixing #908

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: That's a radix defined variable https://www.radix-ui.com/primitives/docs/components/toast#root

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: I don't want to block this PR with the "z-index system" issue, since it can be handled a bit later after the MVP. I'll add this issue as a task to the after MVP task list in clickup.

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: AFAIK we don't have any agreed best practices. I've planned these to be established after the MVP, while we work on the style library.

BUT, I've tried to define anything relating to fonts/text on the nearest parent element to the text, just as a precautions, to keep everything in check if some global variables, etc, go missing.

I'd say let's go forward with this for now on and go over this, when the style library is being worked on.

Oksamies: TS-1957

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: Old habits die hard

packages/cyberstorm/src/components/Toast/Toast.module.css
Oksamies: Radix removes the Toasts when they expire

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: Yes, tested.

apps/cyberstorm-storybook/stories/components/Toast.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+} as Meta<typeof Toast>;
+
+const defaultArgs: ToastProps = {
+  id: uuid(),
+  message: "-",
+  variant: "info",
+};
+
+const Template: StoryFn<typeof Toast> = (args) => (
+  <div style={{ width: "500px" }}>
```
anttimaki, line 18: Just a thought but would it make sense to use the Viewport component here?

packages/cyberstorm/src/components/Providers.tsx
```diff
@@ -1,15 +1,24 @@
...
+  toastDuration?: number;
 }
 
 export function CyberstormProviders(props: CyberstormProvidersProps) {
   return (
-    <RadixTooltip.Provider delayDuration={80}>
-      {props.children}
-    </RadixTooltip.Provider>
+    <Toast.Provider
+      toastDuration={props.toastDuration ? props.toastDuration : 10000}
```
anttimaki, line 15: `toastDuration={props.toastDuration ?? 10000}` would be shorter.

packages/cyberstorm/src/components/Toast/Provider.tsx
```diff
@@ -0,0 +1,71 @@
+import { createContext, PropsWithChildren, useReducer } from "react";
+import { v4 as uuid } from "uuid";
+import * as RadixToast from "@radix-ui/react-toast";
+import Toast from ".";
+import { ToastProps } from "./Toast";
```
anttimaki, line 5: A bit confusing to have to separate imports that have different `from` value but actually import from the same file, I'd combine the imports on L4 & L5.

packages/cyberstorm/src/components/Toast/Provider.tsx
```diff
@@ -0,0 +1,71 @@
...
+        toasts: updatedToasts,
+      };
+    }
+    default:
+      throw new Error(`Unhandled action: ${action.type}`);
+  }
+};
+
+interface ContextInterface {
+  addToast: (props: Omit<ToastProps, "id">) => void;
```
anttimaki, line 40: Strictly speaking I wouldn't consider these props since they're not passed to JSX component 🤓

anttimaki (approved):
Good changes 👍🏻 . Added some new minor comments. Also note that I've replied to some of the comments from the earlier review.

packages/cyberstorm/src/components/Toast/Toast.module.css
anttimaki: And does the expiration always happen very soon after the opacity is set to 0? I.e. the box doesn't first turn transparent and then e.g. 5s seconds later it's removed from the DOM? I assume it does if you tested it, just wanted to clarify what I meant.

packages/cyberstorm/src/components/Toast/Toast.tsx
anttimaki: Just pointing out that the docs also use `aria-hidden` for the child of `Toast.Close`;

packages/cyberstorm/src/index.ts
anttimaki: You've reacted with 👍🏻 but `ToastProps` is still being exported. Pointing this out to make sure the decision is intentional

apps/cyberstorm-storybook/stories/components/Toast.stories.tsx
Oksamies: It is in use, because of Cyberstorm providers being used in preview.js.

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: As I'm not sure how `aria-hidden` is best applied here or with other similar places (places that use `Icon`), I'll add the figuring of aria labels as a separate task to the Cyberstorm Website Release folders lists.

Oksamies: TS-1964

packages/cyberstorm/src/components/Toast/Toast.tsx
Oksamies: I'll add this to the list of stuff to do 👍 TS-1965

packages/cyberstorm/src/index.ts
Oksamies: Yep was intentional, since it's needed in the story

anttimaki: It's not needed in story. Do `variant: "info" as const` which tells TS that the type is "info" and not a string and you're good to go without imports.

packages/cyberstorm/src/components/Providers.tsx
anttimaki: The same tip would apply to `RadixTooltip` on the next line too.

packages/cyberstorm/src/index.ts
Oksamies: Huh, didn't think of that. Nice.

PR #907, opened 2023-11-14
Update Dialog to use children

Oksamies:
refs TS-1955


apps/cyberstorm-storybook/stories/components/Dialog.stories.tsx
```diff
@@ -1,20 +1,24 @@
 import { StoryFn, Meta } from "@storybook/react";
-import { Button, Dialog } from "@thunderstore/cyberstorm";
+import { Button } from "@thunderstore/cyberstorm";
+import { Dialog } from "@thunderstore/cyberstorm";
```
anttimaki, line 3: I see no reason to import these on separate lines.

packages/cyberstorm/src/components/Dialog/Dialog.module.css
```diff
@@ -32,7 +32,7 @@
 }
 
 .bodyPadding {
-  padding: var(--space--16) var(--space--32) var(--space--32) var(--space--32);
```
anttimaki, line 35: Clean up the CSS no longer in use, e.g. the stuff related to footer.

packages/cyberstorm/src/components/Dialog/Dialog.tsx
```diff
@@ -8,64 +8,32 @@ import { faXmarkLarge } from "@fortawesome/pro-solid-svg-icons";
 import { Tooltip } from "../..";
 import { classnames } from "../../utils/utils";
 
-type DialogProps = {
-  content?: ReactNode;
+export type DialogProps = {
```
anttimaki, line 11: Why export this?

packages/cyberstorm/src/components/Dialog/Dialog.tsx
```diff
@@ -8,64 +8,32 @@ import { faXmarkLarge } from "@fortawesome/pro-solid-svg-icons";
...
+  children?: ReactNode;
```
anttimaki, line 12: Could use `interface DialogProps extends PropsWithChildren`

packages/cyberstorm/src/components/Dialog/Dialog.tsx
```diff
@@ -111,25 +79,8 @@ export function Dialog(props: DialogProps) {
                   noPadding ? null : styles.bodyPadding
                 )}
               >
-                {content}
+                {children}
               </div>
-
-              {hideFooter ? null : (
```
anttimaki, line 117: I'm not sure if this is intentional, but as a headsup, if a footer is now passed as a child, it will be rendered inside the `div` above, meaning e.g. `styles.bodyPadding` will affect it as well.

packages/cyberstorm/src/components/Dialog/DialogClose.tsx
```diff
@@ -0,0 +1,18 @@
+"use client";
```
anttimaki, line 1: Is this required?

packages/cyberstorm/src/components/Dialog/DialogClose.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ReactNode } from "react";
+import * as RadixDialog from "@radix-ui/react-dialog";
+
+export interface DialogCloseProps extends RadixDialog.DialogCloseProps {
```
anttimaki, line 5: No need to export this?

Also the PropsWithChildren could be used here as well.

packages/cyberstorm/src/components/Dialog/DialogClose.tsx
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Cyberstorm Dialog Close Sub Component
+ */
+export function DialogClose(props: DialogCloseProps) {
+  const { children, ...restOfProps } = props;
+
+  return <RadixDialog.Close {...restOfProps}>{children}</RadixDialog.Close>;
+}
+
+DialogClose.displayName = "DialogClose";
```
anttimaki, line 18: Not sure if this was discussed and decided we want to have the prefix (e.g. Dialog) in the names of all the subcomponents, but this approach might make the naming cleaner: https://react-typescript-cheatsheet.netlify.app/docs/advanced/misc_concerns/#namespaced-components

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageManagementForm/PackageManagementForm.tsx
```diff
@@ -4,13 +4,16 @@ import { TextInput } from "../../../TextInput/TextInput";
 import { Alert } from "../../../Alert/Alert";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCircleExclamation } from "@fortawesome/pro-solid-svg-icons";
+import { Button } from "../../../..";
```
anttimaki, line 7: I dislike these kinds imports, would add `index` at the end for clarity.

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageManagementForm/PackageManagementForm.tsx
```diff
@@ -4,13 +4,16 @@ import { TextInput } from "../../../TextInput/TextInput";
...
+import { Package } from "@thunderstore/dapper/types";
 
 export interface PackageManagementFormProps {
   packageStatus?: string;
+  packageData: Package;
```
anttimaki, line 12: Dunno if you're noticed while reviewing my changes, but I've actively tried to avoid the `packageData` as a prop name since it's clunky and redundant. Anyway, in this case it would seem that simply passing `isDeprecated` as a prop would suffice.

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamMembers/TeamMemberList.tsx
```diff
@@ -77,16 +79,13 @@ export function TeamMemberList(props: Props) {
...
-          acceptButton={
-            <Button.Root colorScheme="danger" paddingSize="large">
-              <Button.ButtonLabel>Kick member</Button.ButtonLabel>
-            </Button.Root>
-          }
-          cancelButton="default"
-          showFooterBorder
-          title="Confirm member removal"
-        />
+            <div className={styles.kickMemberDialogFooter}>
```
anttimaki, line 82: Are the related styles implemented? Is the `div` in fact unnecessary?

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamMembers/TeamMembers.module.css
```diff
@@ -50,3 +50,8 @@
 .kickDescriptionUserName {
   color: var(--color-highlight);
 }
+
+.createAddMemberDialogFooter {
```
anttimaki, line 54: Is this class actually used anywhere?

anttimaki (approved):
No hard blockers but I think at least some of the comments, like cleaning up the dead CSS, should be done within this PR.

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamMembers/TeamMembers.module.css
Oksamies: I miss wrote the class name, should have been `addMemberDialogFooter`

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: No reason 👍

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: I'm aware. I'll be going over the Dialogs again on the second QA round and fixing any problems, that might have arisen.

packages/cyberstorm/src/components/Dialog/DialogClose.tsx
Oksamies: Yep

packages/cyberstorm/src/components/Dialog/DialogClose.tsx
Oksamies: I removed the "Dialog" prefix, but I couldn't get that thing working that was suggested in the link.

Problem came when I had to export the component through packages/cyberstorm/src/index.

I've added a task to sort out the imports and export as a whole after the MVP. Seems like there's a lot of inconsistencies, which might be more efficient to fix all at ones.

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamMembers/TeamMembers.tsx
```diff
@@ -51,13 +50,13 @@ export function TeamMembers(props: Props) {
...
           </div>
         </div>
-      }
-      acceptButton={
-        <Button.Root colorScheme="success">
-          <Button.ButtonLabel>Add Member</Button.ButtonLabel>
-        </Button.Root>
-      }
-    />
+        <div className={styles.addMemberDialogFooter}>
```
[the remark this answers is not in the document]
Oksamies: Mishap with class names

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamServiceAccounts/ServiceAccountList/ServiceAccountList.tsx
```diff
@@ -52,16 +54,13 @@ export function ServiceAccountList(props: Props) {
...
-          acceptButton={
-            <Button.Root colorScheme="danger" paddingSize="large">
-              <Button.ButtonLabel>Remove service account</Button.ButtonLabel>
-            </Button.Root>
-          }
-          cancelButton="default"
-          showFooterBorder
-          title="Confirm service account removal"
-        />
+            <div className={styles.removeServiceAccountFooter}>
```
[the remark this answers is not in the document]
Oksamies: It's in use. Because the dialog doesn't have a footer anymore.

apps/cyberstorm-storybook/stories/components/Dialog.stories.tsx
```diff
@@ -4,17 +4,20 @@
...
   trigger: (
     <Button.Root>
       <Button.ButtonLabel>trigger</Button.ButtonLabel>
     </Button.Root>
   ),
-  title: "This is a dialog",
-  content: (
+};
+
+const Template: StoryFn<typeof Dialog.Root> = (args: any) => (
```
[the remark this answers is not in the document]
anttimaki: Don't use `any`

packages/cyberstorm/src/components/Dialog/Dialog.module.css
```diff
@@ -31,8 +31,26 @@
...
+
+.bodyPadding > :first-child {
+  padding-top: var(--space--16);
+}
+
+.bodyPadding > :last-child {
+  padding-bottom: var(--space--16);
+}
+
+.bodyPadding:where(.footerBorder) > :nth-child(n - 2) {
```
anttimaki, line 47: What is `n - 2` supposed to select, and why do we want to use it here when we don't know what the content of the children will be? I first assumed it would apply to all but first three (or two) children, but testing on a [online test bench](https://nth-child.belter.io/) seems to show it applies to all children. Is this a bug?

packages/cyberstorm/src/components/Dialog/Dialog.tsx
```diff
@@ -108,28 +78,12 @@ export function Dialog(props: DialogProps) {
               <div
                 className={classnames(
                   styles.body,
-                  noPadding ? null : styles.bodyPadding
+                  noPadding ? null : styles.bodyPadding,
+                  showFooterBorder ? null : styles.footerBorder
```
anttimaki, line 82: TBH what you're trying to do with this prop seems really hacky. Can you come up with alternative ways to achieve what you're aiming for?

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageDependencyList/PackageDependencyList.tsx
```diff
@@ -5,11 +5,11 @@ import { WrapperCard } from "../../../WrapperCard/WrapperCard";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faBoxOpen } from "@fortawesome/pro-regular-svg-icons";
 import { faCaretRight } from "@fortawesome/pro-solid-svg-icons";
-import { Dialog } from "../../../Dialog/Dialog";
 import { PackageDependencyDialog } from "./PackageDependencyDialog/PackageDependencyDialog";
 import * as Button from "../../../Button/";
 import { useDapper } from "@thunderstore/dapper";
 import { usePromise } from "@thunderstore/use-promise";
+import { Dialog } from "../../../..";
```
anttimaki, line 12: The latest changes brought multiple of these import back the PR. IDK if the tooling for autoimports could be improved to resolve this. Until then I'd prefer we'd fix these manually. Not a blocker though.

anttimaki (changes requested):
I think the `any` type and CSS `nth-child` bug (unless it's just misunderstanding on my part) should be fixed. Also check the older comments for replies.

packages/cyberstorm/src/components/Dialog/Dialog.tsx
anttimaki: It's still exported?

packages/cyberstorm/src/components/Dialog/Dialog.tsx
anttimaki: No changes for this one either, what does the :+1: reaction signify here? 😅

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: Huh, that's weird. Should not be now. Will have to double check these in future.

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: That I've fixed it, but it seems that I've walked back the type changes at somepoint after that. Idk, I might do the same mistakes again, while fixing a later comment.

But the 👍 always means that I've fixed it, when I react with that.

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: This is now needed on .stories file

packages/cyberstorm/src/components/Dialog/Dialog.module.css
Oksamies: Second last, but seems like it worked by accident on my env. Fixed it now

packages/cyberstorm/src/components/Dialog/Dialog.tsx
Oksamies: Needed in the .stories

anttimaki: It's not needed, Storybook can infer the type of `args` from the information it already has? Also the import is done using the full path, which shouldn't be done, but that doesn't matter if you remove the whole import.

packages/cyberstorm/src/components/Dialog/Dialog.tsx
anttimaki: Now there's some `preStyledFooter` prop that isn't used?

packages/cyberstorm/src/components/Layout/PackageDetailLayout/PackageDependencyList/PackageDependencyList.tsx
anttimaki: Seems they are not all fixed yet.

PR #908, opened 2023-11-14
Add RHF and CreateTeamForm

Oksamies:
refs TS-1942, TS-1829


packages/cyberstorm/package.json
```diff
@@ -38,6 +38,7 @@
     "react": "^18.2.0",
     "react-data-table-component": "^7.5.3",
     "react-dom": "^18.2.0",
+    "react-hook-form": "^7.48.2",
```
anttimaki, line 41: Not really related to this line, but: have we considered whether the forms should be defined in the Cyberstorm package or in the nextjs app? IDK right now what the ramifications of the decision would be (if any). Based on a gut feeling I'd say the forms are unlikely to be very reusable and relate more to the app than the components.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.module.css
```diff
@@ -0,0 +1,37 @@
+.createTeamDialog {
```
anttimaki, line 1: Since we're at least currently using CSS modules, prefixing everything with `createTeam` is unnecessary.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.module.css
```diff
@@ -0,0 +1,37 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: var(--space--24);
+}
+
+.createTeamForm {
```
anttimaki, line 7: Y U NO .root? Also should this be the first thing in the file?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
+"use client";
+import styles from "./CreateTeamForm.module.css";
+import * as Button from "../../Button";
+import { TextInput } from "../../TextInput/TextInput";
+import { useForm, FieldErrors, useController } from "react-hook-form";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faArrowsRotate } from "@fortawesome/free-solid-svg-icons";
+
```
anttimaki, line 8: This empty line is confusing since the imports aren't in any logical order.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useContext } from "react";
+import { ToastContext } from "../../Toast/ToastContext";
+import {
+  faCircleCheck,
+  faCircleExclamation,
+  faOctagonExclamation,
+} from "@fortawesome/pro-solid-svg-icons";
+
+/**
+ * Form for creating a team
```
anttimaki, line 18: Could have figured as much from the component name.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+ */
+export function CreateTeamForm() {
+  const useToast = () => useContext(ToastContext);
```
anttimaki, line 21: Why not just `const toast = useContext(ToastContext);`?

Better yet [do something like this](https://github.com/thunderstore-io/thunderstore-ui/blob/master/apps/cyberstorm-nextjs/utils/SessionContext.tsx#L101-L109) to ensure the toast is not null so you don't have to check for it everywhere. And naturally do it once in the toast "module" and just import it here so it doesn't have to be repeated everywhere.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const toast = useToast();
+
+  async function onValid(data: { teamName: string }) {
```
anttimaki, line 24: Define the interface for the form's data exactly once and reuse it.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    // TODO: Add sending to TS API endpoint
+    toast?.addToast(
+      "info",
+      <FontAwesomeIcon icon={faCircleExclamation} />,
+      "Creating team"
+    );
+    await new Promise((resolve) => {
+      setTimeout(resolve, 1000);
+    });
+    // TODO: Add a toast on response return based on the response
```
anttimaki, line 34: Also remove the "Creating team" toast once the request is complete?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    toast?.addToast(
+      "success",
+      <FontAwesomeIcon icon={faCircleCheck} />,
+      "Team created"
+    );
+    console.log(JSON.stringify(data));
+  }
+
+  async function onInvalid(
```
anttimaki, line 43: I believe this is exactly the boilerplatey approach Mythic asked to avoid. Did you check the Zod option? Did you try to generalize this approach to a reusable helper?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+        "danger",
+        <FontAwesomeIcon icon={faOctagonExclamation} />,
+        errors.teamName.message,
+        30000
+      );
+    } else {
+      toast?.addToast(
+        "danger",
+        <FontAwesomeIcon icon={faOctagonExclamation} />,
+        "Unknown error",
```
anttimaki, line 59: Surely RHF gives us some information about what the error was?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    formState: { isSubmitting },
+  } = useForm<{ teamName: string }>({
+    mode: "onSubmit",
+  });
+
+  const teamName = useController({
+    control: control,
+    name: "teamName",
+    rules: {
+      required: "Team name is required",
```
anttimaki, line 77: Isn't RHF able to generate such a simple error message automatically?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+  });
+
+  return (
+    <form
```
anttimaki, line 82: There's no commit message and the linked tasks don't really tell me anything so it's a bit hard to know which level of feedback would be needed here. So in short, from the top of my mind IDK what would be the best way to handle this, but this approach seems very verbose and manual. Maybe it's what we were aiming for, for now? I'm just a bit worried that since this is the first form, it will get copy-pasted around, creating code debt.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+              ? teamName.fieldState.invalid
+                ? "red"
+                : "green"
+              : undefined
+          }
+        />
+      </div>
+      <div className={styles.footer}>
+        <Button.Root type="submit" paddingSize="large" colorScheme="success">
+          {isSubmitting ? (
```
anttimaki, line 105: If it's not built-in, consider disabling the form when isSubmitting. Note that this might require disabling more than just the button (e.g. enter press on the text field).

packages/cyberstorm/src/components/Layout/Developers/PackageUpload/PackageUploadLayout.tsx
```diff
@@ -34,14 +36,19 @@ export function PackageUploadLayout() {
...
-                  <Button.ButtonLabel>Create Team</Button.ButtonLabel>
-                  <Button.ButtonIcon>
-                    <FontAwesomeIcon icon={faPlus} />
-                  </Button.ButtonIcon>
-                </Button.Root>
-              </div>
+              <Dialog.Root
+                title="Create Team"
+                trigger={
+                  <Button.Root colorScheme="primary" paddingSize="large">
```
anttimaki, line 42: I'm not sure this matches the designs in Figma?

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -1,17 +1,22 @@
 "use client";
-import React from "react";
+import React, { ReactNode } from "react";
 import styles from "./TextInput.module.css";
 import { Icon } from "../Icon/Icon";
 import { classnames } from "../../utils/utils";
 
 export interface TextInputProps {
+  children?: ReactNode;
```
anttimaki, line 8: What?

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -1,17 +1,22 @@
...
   placeHolder?: string;
   leftIcon?: JSX.Element;
   id?: string;
   type?: "text" | "email" | "password" | "tel" | "url";
   rightIcon?: JSX.Element;
   value?: string;
-  setValue?: React.Dispatch<React.SetStateAction<string>>;
+  name?: string;
+  color?: string;
+  ref?: (instance: any) => void;
```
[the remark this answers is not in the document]
anttimaki: Don't use `any`. If there's no sensible way around this, disable the linting for the row(s) and add a comment describing why it was necessary.

Also for the ref, the [article mentioned in another PR discusses refs](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#maybe-reactjsxintrinsicelements-or-elementhtmlattributes) as well.

Also also, I feel the need to define e.g. the name is an antipattern, one that the `ComponentPropsWithRef` from the article might resolve neatly.

anttimaki (commented):
I was in the middle of the review when I noticed this wasn't assigned to me, but had already wrote some comments, so decided to finish the review anyway. Marking just as comment and letting Mythic handle the rest as you intended.

packages/cyberstorm/src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
Oksamies: Nice, didn't know about `ComponentPropsWithRef`. I've added proper types 👍

packages/cyberstorm/package.json
Oksamies: Have not really talked about this yet; I had an idea that we should move the layouts to the nextjs app. So then these forms would move there also. I don't want to do anything towards that yet though, not at least before we've come to an unified conclusion on the matter.

Let's think about this after the MVP 👍

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.module.css
Oksamies: Hihi 👍

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: The idea is that it goes away after it's timer is done, found it a bit better for UX and there's plenty of room for few toasts. We'll tweak the timers and animations after MVP with Gösta 👍

packages/cyberstorm/src/components/Layout/Developers/PackageUpload/PackageUploadLayout.tsx
Oksamies: Did not before, does now 👍

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: I think I've reduced boilerplate a lot in the fix comment. Used zod

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: Does with Zod and response from API

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: No I don't think so. But with Zod it does

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: I think the new version in fix commit, is a lot better 🤷

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: You know, I don't know how to do that and didn't find a way after looking at it for a good while. I mean I know how to do it with a normal form, but with RHF I've no clue 😕

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.module.css
```diff
@@ -0,0 +1,37 @@
+.root {
+  display: flex;
+  flex-direction: column;
+  gap: 2rem;
```
anttimaki, line 4: For `gap`, `.root` uses `2rem`, `.dialog` uses `var(--space--24)`, and `.footer` uses `var(--gap--16);`. Could we manage with one approach?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useController, useForm, Form } from "react-hook-form";
+import * as Button from "../../Button";
+import { TextInput } from "../../TextInput/TextInput";
+import { useToast } from "../../Toast/Provider";
+import styles from "./CreateTeamForm.module.css";
+import { z } from "zod";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { isRecord } from "../../../utils/type_guards";
+
+interface errorResponse {
```
anttimaki, line 13: * By convention it should be ErrorResponse with capital E
* This (and the type guard below) should probably reside in the utils file somewhere?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+    isRecord(response) &&
+    isRecord(response.error) &&
+    typeof response.error.message === "string"
+  );
+}
+
+export function CreateTeamForm() {
+  const toast = useToast();
+
+  interface formFields {
```
anttimaki, line 28: * By convention it should be FormFields with capital F
  * But to stick to what RHF's own documentation is using, it could be called `Schema`
* The interface and the schema below should probably be defined outside the function?
* If you check the [example in the documentations](https://react-hook-form.com/docs/useform#resolver) you'll find that you can infer the interface instead of manually duplicating it. I think this issue is borderline blocker.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+    teamName: string;
+  }
+
+  const schema = z.object({
+    teamName: z.string({ required_error: "Team name is required" }),
```
anttimaki, line 33: [Zod's refine](https://zod.dev/?id=refine) and regexp could perhaps be used to validate the value matches the rules we show on the form? While not a blocker per se, it would be nice to have this kind of "essentials" done right from the get-go.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+  } = useForm<formFields>({
+    mode: "onSubmit",
+    resolver: zodResolver(schema),
+  });
+
+  const teamName = useController({
+    control: control,
+    name: "teamName",
+    rules: {
+      required: "Team name is required",
```
anttimaki, line 48: Do we really need to duplicate the message here? Seems to defeat the point of Zod's validation?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+    },
+  });
+
+  return (
+    <Form
+      control={control}
+      action="https://thunderstore.io/api/team/create"
+      method="post"
+      onSubmit={() => {
```
anttimaki, line 57: Seems that when testing locally, the field value is not validated before submit. I.e. entering a character and then removing it (to activate the submit button) lets me submit an empty string to backend. You probably need to check the documentation for some missing piece? Unless the problem is on my local env, I'd consider this a borderline blocker.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,111 @@
...
+      }}
+      onSuccess={(response) => {
+        console.log(response);
+        toast.addToast({ variant: "success", message: "Team created" });
+      }}
+      onError={(response) => {
+        if (isErrorResponse(response)) {
+          toast.addToast({
+            variant: "danger",
+            message: response.error.message,
```
anttimaki, line 68: We should probably show some error toast to user anyway? Or redirect them to 500 page? Now it might seem to them that nothing happened or perhaps the creation succeeded?

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -4,14 +4,19 @@ import styles from "./TextInput.module.css";
 import { Icon } from "../Icon/Icon";
 import { classnames } from "../../utils/utils";
 
-export interface TextInputProps {
+export interface TextInputProps
+  extends React.ComponentPropsWithRef<"input">,
```
anttimaki, line 8: FYI opening the create form dialog on teams-page now shows this in console:

```
Warning: TextInput: `ref` is not a prop. Trying to access it will result in `undefined` being returned.
If you need to access the same value within the child component, you should pass it as a different
prop. (https://reactjs.org/link/special-props)
...


Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you
mean to use React.forwardRef()?
Check the render method of `CreateTeamForm`.
…
```

Don't know if it's related to this change, but it seems likely?

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -4,14 +4,19 @@ import styles from "./TextInput.module.css";
...
+    React.PropsWithChildren {
```
anttimaki, line 9: To clarify my "what" comment from the earlier review: "what in tarnation?"

Seriously though, why accept children? Doesn't seem they're used anywhere in the context of this PR? Also IDK what the use case for passing children to an input would be? Pls remove or justify the need for this in a review comment (and possibly code comment as well).

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -4,14 +4,19 @@ import styles from "./TextInput.module.css";
...
   placeHolder?: string;
   leftIcon?: JSX.Element;
   id?: string;
```
anttimaki, line 12: Now that we use `ComponentPropsWithRef<"input">`, doesn't that mean we don't need to manually define id, type, name, onBlur, or onChange, unless we want to override the types of whatever values `<input>` accepts normally? (And placeHolder, if we rename it to placeholder.)

anttimaki (changes requested):
I think including Zod is a step into right direction, but currently the PR doesn't utilize it to it's full potential.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: Agreed with Mythic to not do extensive client-side validation right now.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: No

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: Added a Toast for it.

packages/cyberstorm/src/components/TextInput/TextInput.tsx
Oksamies: Fixd

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: just added the disabled there

packages/cyberstorm/src/components/Button/Button.tsx
```diff
@@ -90,6 +92,7 @@ const Button = React.forwardRef<
   if (plain) {
     const fRef = forwardedRef as React.ForwardedRef<HTMLDivElement>;
     const ref = fRef || fallbackRef;
+    // TODO: Add disabled mode
```
anttimaki, line 95: Doesn't this commit already add the disabled mode?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+import styles from "./CreateTeamForm.module.css";
+import { z } from "zod";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { isErrorResponse } from "../../../utils/type_guards";
+import { useEffect } from "react";
+
+export function CreateTeamForm() {
+  const toast = useToast();
+
+  // "Required field" error is different from value being too little
```
anttimaki, line 17: This comment raises more questions than answers. How is it different? What do we need required_error for if we have a separate check for value length?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const teamName = useController({
+    control: control,
+    name: "teamName",
+  });
+
+  // We have to do this because the RHF Form component is in beta and it
+  // doesn't have a callback prop like "onValidation" that could take in
+  // the generalized addToast error informing block.
+  useEffect(() => {
```
anttimaki, line 41: IMO this is bad UX. The form error should be shown on the form, next to the field. Not as a toast.

Also if the beta status actually prevents us from using such basic features (I assume it actually doesn't), we should reconsider if it's something we want to use in this soon-to-be-production project.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+    if (errors.teamName) {
+      console.log(errors);
+      toast.addToast({
+        variant: "danger",
+        message: errors.teamName.message,
+        duration: 30000,
+      });
+    }
+  }, [errors]);
```
anttimaki, line 50: I can't be sure but it's possible this objects gets recreated on each render and thus the effect also runs each time. Anyways the function would actually only need `errors.teamName?.message`, so just use it as the dependency. As a general rule prefer primitive values in the dependency array if feasible, that way you don't have to worry about this.

But I assume we don't really need the useEffect at all since I assume we don't want to show field errors on the toast component.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+      onError={(response) => {
+        if (isErrorResponse(response)) {
+          toast.addToast({
+            variant: "danger",
+            message: response.error.message,
+            duration: 30000,
+          });
+        } else {
+          // TODO: Add sentry error here
+          console.log("TODO: Sentry error logging missing!");
```
anttimaki, line 73: Not a fan of logging TODO items.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          toast.addToast({
+            variant: "danger",
+            message: "Unhandled form response error",
```
anttimaki, line 76: Could you think something more user friendly for the message?

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+      }}
+      validateStatus={(status) => status === 200}
+      className={styles.root}
+    >
+      <div className={styles.dialog}>
+        <div className={styles.dialogText}>
+          Enter the name of the team you wish to create. Team names can contain
+          the characters a-z A-Z 0-9 _ and must not start or end with an _
+        </div>
+        <TextInput
```
anttimaki, line 89: Not sure if it has been discussed yet design-wise, but do we want to denote required fields somehow?

packages/cyberstorm/src/components/Layout/Settings/Profile/Profile.tsx
```diff
@@ -43,12 +43,12 @@ export function Profile(props: ProfileProps) {
         <SettingItem
           title="Profile Summary"
           description="A short introduction of a max of 160 characters"
-          content={<TextInput placeHolder={userData.description} />}
+          content={<TextInput placeholder={userData.description} />}
```
anttimaki, line 46: Is this actually a placeholder? Looks more like the proper value, prefilled into the form? Comment applies to other changes in this file as well.

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamDetails/TeamDetails.tsx
```diff
@@ -76,12 +76,12 @@ export function TeamDetails(props: Props) {
         <SettingItem
           title="Profile Summary"
           description="A short description shown in header and profile cards"
-          content={<TextInput placeHolder={teamData.description} />}
+          content={<TextInput placeholder={teamData.description} />}
```
anttimaki, line 79: Is this actually a placeholder? Looks more like the proper value, prefilled into the form? Comment applies to other changes in this file as well.

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -4,64 +4,60 @@ import styles from "./TextInput.module.css";
 import { Icon } from "../Icon/Icon";
 import { classnames } from "../../utils/utils";
 
-export interface TextInputProps {
-  placeHolder?: string;
+export interface TextInputProps
+  extends React.ComponentPropsWithoutRef<"input"> {
+  value?: string;
```
anttimaki, line 9: This should probably accept numbers as well?

packages/cyberstorm/src/components/TextInput/TextInput.tsx
```diff
@@ -4,64 +4,60 @@ import styles from "./TextInput.module.css";
...
+        ) : null}
+        <input
+          {...elementProps}
+          ref={ref}
+          className={classnames(
+            styles.input,
+            leftIcon ? styles.hasLeftIcon : null
+          )}
+          value={value}
+          onKeyDown={(e) => onEnter(e)}
```
anttimaki, line 50: Nitpick but `onKeyDown={onEnter}` would do the same

packages/cyberstorm/src/components/Toast/Provider.tsx
```diff
@@ -37,16 +42,16 @@ const toastReducer = (
 };
 
 interface ContextInterface {
-  addToast: (props: Omit<ToastProps, "id">) => void;
+  addToast: ({ ...props }: Omit<ToastProps, "id">) => void;
   remove: (id: string) => void;
 }
 
-const ToastContext = createContext<ContextInterface | null>(null);
+export const ToastContext = createContext<ContextInterface | null>(null);
```
anttimaki, line 49: Seems `ToastContext` is used only within this file so why export it? I'd assume we'd like the developers to use the `useToast` hook so not providing the context might be a good thing. 🤷🏻

anttimaki (approved):
I'll leave it up to you to decide how dire the need to fix each comment is. Personally I think the error-in-toast UX is something that we'd want to fix asap. Also check one comment from earlier review that's not yet fully fixed.

packages/cyberstorm/src/components/Forms/CreateTeamForm/CreateTeamForm.tsx
Oksamies: Didn't see anything in designs. I'll ask around and add markers (or whatever we'll use) in a another PR 👍

packages/cyberstorm/src/components/Button/Button.tsx
Oksamies: The plain button needs more than adding the disabled attribute to the element, since it's a div.

PR #909, opened 2023-11-15
Versions missing alert

Cyanideankst:
[Refs TS-1874 - Package - Versions tab missing alert](https://app.clickup.com/t/20413243/TS-1874)


Oksamies (approved):

anttimaki:
Could the "Thunderstore Mod Manager" in the alert link to the download page, like the banner on top of the current website does?

PR #910, opened 2023-11-16
Avatar component

Oksamies:


packages/cyberstorm/src/components/Avatar/Avatar.module.css
```diff
@@ -1,19 +1,31 @@
 .root {
+  display: flex;
   align-items: center;
   justify-content: center;
-  border-radius: var(--size) !important;
-  background-color: #fff;
+  border-radius: calc(var(--size) * 4.92) !important;
```
anttimaki, line 5: What's the purpose of oddly specific 4.92? Is it something we want to use here or everywhere with round elements? Should it be (global) variable?

packages/cyberstorm/src/components/Avatar/Avatar.module.css
```diff
@@ -1,19 +1,31 @@
...
-  min-height: var(--size);
-  max-height: var(--size);
-  border-radius: var(--size);
+  width: var(--size);
+  height: var(--size);
+  border-radius: calc(var(--size) * 4.92);
   transition: opacity 150ms ease-in-out;
 }
 
+.image:where(.placeholder) {
```
anttimaki, line 16: IDK what's the status with global CSS buit should we use it for `#39e9aa` and `line-height` here?

packages/cyberstorm/src/components/Avatar/AvatarButton.tsx
```diff
@@ -27,11 +34,17 @@ export const AvatarButton = React.forwardRef<
       {...forwardedProps}
       ref={ref}
     >
-      <img
-        src={src}
-        className={classnames(styles.image, getSize(size))}
-        alt=""
-      />
+      {src ? (
```
anttimaki, line 37: Could we just use Avatar component here to avoid duplication?

anttimaki (approved):
The CSS comments are just "was this intentional" level stuff, you know better the current state and target state of all that stuff.

The DRY comment would be a nice-to-have.

packages/cyberstorm/src/components/Avatar/AvatarButton.tsx
Oksamies: We could. We'd just a have a double wrapper problem. I'll add a separate task for combining the components.

packages/cyberstorm/src/components/Avatar/Avatar.module.css
Oksamies: It's a common multiplier between the different sizes border-radius'. A bit rough, but will do.

PR #917, opened 2023-11-22
Fix CreateTeam forms UI problems and add field error parsing to Cyberstorm forms

Oksamies:


packages/thunderstore-api/src/errors.ts
```diff
@@ -3,17 +3,22 @@ export function isApiError(e: Error | ApiError | unknown): e is ApiError {
...
+    this.response = response;
   }
 
-  static createFromResponse(response: Response): ApiError {
+  static async createFromResponse(response: Response): Promise<ApiError> {
     // TODO: Implement response parsing for known error scenarios
-    return new ApiError(`${response.status}: ${response.statusText}`);
+    return new ApiError(
+      `${response.status}: ${response.statusText}`,
+      await response.json()
```
MythicManiac, line 17: You need to try-catch this, it might not be json

packages/thunderstore-api/src/errors.ts
```diff
@@ -3,17 +3,22 @@ export function isApiError(e: Error | ApiError | unknown): e is ApiError {
...
+    );
   }
 
   getFieldErrors(): { [key: string]: string[] } {
-    // TODO: Implement
-    return {};
+    return { ...this.response };
```
MythicManiac, line 22: Not guaranteed to exist if parsing is done properly

packages/thunderstore-api/src/errors.ts
```diff
@@ -3,17 +3,22 @@ export function isApiError(e: Error | ApiError | unknown): e is ApiError {
 }
 
 export class ApiError extends Error {
-  constructor(message?: string) {
+  response: { [key: string]: string[] } | undefined;
```
MythicManiac, line 6: Calling this `responseData` or such would be more appropriate. `response` would be appropriate if you stored the actual response object itself.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -16,15 +16,18 @@ export async function apiFetch2(args: apiFetchArgs) {
...
   const response = await fetch(url, {
+    method: args.method ? args.method : "GET",
     headers: {
       ...BASE_HEADERS,
       ...getAuthHeaders(args.config),
     },
-    body: args.body,
+    ...(args.method === "GET" || args.method === "HEAD" || args.method === null
+      ? {}
+      : { body: args.body }),
```
MythicManiac, line 26: This is wrong, whether a request body should be sent has nothing to do with the request method. It's just more common for that to be done with POST requests.

If having `undefined` in the body was causing issues, check if `null` works instead:

```suggestion
    body: body ?? null,
```

According to the typescript type hints at least, that should be valid.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -16,15 +16,18 @@ export async function apiFetch2(args: apiFetchArgs) {
   const url = getUrl(args.config, args.path, args.query);
 
   const response = await fetch(url, {
+    method: args.method ? args.method : "GET",
```
MythicManiac, line 19: Don't do this, change the type hint for `apiFetchArgs` to require the type to be provided instead if you want to enforce it.

The best option would probably be to move both the `method` and `body` arguments under the same object and let the caller control any of the normal fetch parameters like so:

```ts
export type apiFetchArgs = {
  config: RequestConfig;
  path: string;
  query?: string;
  request?: Omit<RequestInit, "headers">;
};
```

Then you can just merge it to the fetch call as-is:

```ts
  const response = await fetch(url, {
    ...(args.request ?? {}),
    headers: {
      ...BASE_HEADERS,
      ...getAuthHeaders(args.config),
    },
  });
```

It should be the caller's responsibility to call the function correctly, and typescript's is used to make sure the caller knows what "correctly" means.

packages/thunderstore-api/src/apiFetch.ts
MythicManiac: This suggestion isn't necessary if you follow the larger change suggestion, leaving it as a note anyway.

MythicManiac (approved):
Good to go now as far as I'm concerned, but do check that it works for the use case you tested it for

PR #918, opened 2023-11-22
Prepare PackageDetailLayout for Dapper implementation

anttimaki:
@thunderstore/cyberstorm: light cleanup for PackageDetailLayout

Reorder things for better readability.

Refs TS-1979

@thunderstore/cyberstorm: adjust PackageDetailLayout props

- Use the same prop names as the other layouts that accept the same
  props, like PackageDependantsLayout
- Don't export the prop type since it's not necessary
- Remove managementDialogIsOpen prop from the layout. This was used to
  control Storybook stories, but by and large makes little sense on a
  layout component. Considered removing the prop from the Dialog
  component altogether but decided to leave it for now to avoid scope
  creep
- Remove the aforementioned Storybook story and cleanup the related
  code

Refs TS-1979

@thunderstore/dapper: remove fields from Package interface

- Remove shortDescription somce backend only has one description to
  return, and it's already included in the Package interface via
  PackagePreview interface
- Remove additionalImages since we don't have any and they aren't
  used anyway
- Remove discordLink since package doesn't have one. Community might
  have one, but I don't think it makes sense to show it on package's
  detail page, since the package or its author might have nothing to
  do with that Discord server

Refs TS-1979

@thunderstore/dapper: rename Package interface fields

- Rename githubLink to website_url. Package has one link, but it
  doesn't necessarily point to GitHub. The link is guaranteed to exist,
  since it's a required field in the package's manifest.json. Rendering
  imitates how the old website renders these links
- Rename firstUploaded to datetime_created. This is consistent with
  other fields returned by Dapper (for some reason backend has named this
  field date_created, although it's a datatime field). The field is
  guaranteed to exist, obviously. Fix the label shown in the UI. Fix the
  data returned by dapper-fake to match the format backend actually
  returns
- Rename dependencyString to full_version_name to match backend naming.
  Fix the data returned by dapper-fake to something more authentic
- Rename dependantCount to dependant_count to match backend naming

Refs TS-1979

@thunderstore/dapper: rename Package.versions fields

- Rename version to version_number and downloadCount to download_count
  to match the naming in backend
- Rename uploadDate to datetime_created and and change the format in
  dapper-fake to match the data returned by backend
- Since the versions are shown on a separate tab, the versions might
  end up being fetched with a separate dedicated dapper call in the
  future

Refs TS-1979

@thunderstore/dapper: rename PackageDependency fields

- Rename fields to match backend naming
- Change dapper-fake to return data matchign the actual data
- Remove dependencies field from Package interface since it's really
  not used there. Dependencies have their own (unimplemented) Dapper
  method

Refs TS-1979

### DIFF
--- a/apps/cyberstorm-nextjs/app/c/[community]/p/[namespace]/[package]/page.tsx
+++ b/apps/cyberstorm-nextjs/app/c/[community]/p/[namespace]/[package]/page.tsx
@@ -7,8 +7,8 @@ export default function Page({
 }) {
   return (
     <PackageDetailLayout
-      community={params.community}
-      namespace={params.namespace}
+      communityId={params.community}
+      namespaceId={params.namespace}
       packageName={params.package}
     />
   );
--- a/apps/cyberstorm-storybook/stories/layouts/PackageDetailLayout.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/layouts/PackageDetailLayout.stories.tsx
@@ -8,28 +8,14 @@ const meta = {
 } as Meta<typeof PackageDetailLayout>;
 
 const Template: StoryFn<typeof PackageDetailLayout> = (args) => (
-  <div>
-    <PackageDetailLayout {...args} />
-  </div>
+  <PackageDetailLayout {...args} />
 );
 
-const defaultArgs = {
-  community: "Train City 2042",
-  namespace: "Mechanics",
-  packagepackageNameId: "Thomas the Dankiest Engine",
-};
-
 const ReferencePackageDetailLayout = Template.bind({});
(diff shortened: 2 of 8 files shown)


anttimaki:
* master
  * **PR #918** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/918?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/918?utm_source=stack-comment).

Oksamies (approved):

PR #919, opened 2023-11-22
Add MultiSelectSearch component

Oksamies:
TS-1968

### DIFF
--- a/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
@@ -30,7 +30,9 @@ const defaultArgs = {
 };
 
 const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
-  const [selected, setSelected] = useState(options);
+  const [selected, setSelected] = useState<{ label: string; value: string }[]>(
+    []
+  );
   const defaultProps = {
     ...args,
     onChange: (x: { label: string; value: string }[]) => setSelected(x),
@@ -39,12 +41,7 @@ const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
   return (
     <div>
       <div style={{ color: "white" }}>
-        Value in state:{" "}
-        {selected
-          .map((s) => {
-            return s.label;
-          })
-          .join(", ")}
+        Value in state: {selected.map((s) => s.label).join(", ")}
       </div>
       <MultiSelectSearch {...defaultProps} />
     </div>
--- a/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
@@ -34,14 +34,17 @@ const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
   const defaultProps = {
     ...args,
     onChange: (x: { label: string; value: string }[]) => setSelected(x),
+    value: selected,
   };
   return (
     <div>
       <div style={{ color: "white" }}>
         Value in state:{" "}
-        {selected.map((s) => {
-          return s.label;
-        })}
+        {selected
+          .map((s) => {
+            return s.label;
+          })
+          .join(", ")}
       </div>
       <MultiSelectSearch {...defaultProps} />
     </div>
--- a/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
@@ -1,0 +1,63 @@
+import { StoryFn, Meta } from "@storybook/react";
+import React, { useState } from "react";
+import { MultiSelectSearch } from "@thunderstore/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch";
+
+const meta = {
+  title: "Cyberstorm/Components/MultiSelectSearch",
+  component: MultiSelectSearch,
+} as Meta<typeof MultiSelectSearch>;
+
+const options = [
+  { label: "Team 1", value: "Team 1" },
+  { label: "Team 2", value: "Team 2" },
+  { label: "Team 3", value: "Team 3" },
+  { label: "Team 4", value: "Team 4" },
+  { label: "Team 5", value: "Team 5" },
+  { label: "Team 6", value: "Team 6" },
+  { label: "Team 7", value: "Team 7" },
+  { label: "Team 8", value: "Team 8" },
+  { label: "Team 9", value: "Team 9" },
+  { label: "Team 10", value: "Team 10" },
+  { label: "Team 11", value: "Team 11" },
+  { label: "Team 12", value: "Team 12" },
+  { label: "Team 13", value: "Team 13" },
+  { label: "Team 14", value: "Team 14" },
+];
+
+const defaultArgs = {
+  placeholder: "Select something",
+  options: options,
+};
+
+const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
+  const [selected, setSelected] = useState(options);
+  const defaultProps = {
+    ...args,
+    onChange: (x: { label: string; value: string }[]) => setSelected(x),
+  };
+  return (
+    <div>
+      <div style={{ color: "white" }}>
+        Value in state:{" "}
+        {selected.map((s) => {
+          return s.label;
+        })}
+      </div>
+      <MultiSelectSearch {...defaultProps} />
+    </div>
+  );
+};
+
+const GreenMultiSelectSearch = Template.bind({});
+GreenMultiSelectSearch.args = {
+  ...defaultArgs,
+  color: "green",
+};
+
+const RedMultiSelectSearch = Template.bind({});
+RedMultiSelectSearch.args = {
+  ...defaultArgs,
+  color: "red",
+};
+
+export { meta as default, GreenMultiSelectSearch, RedMultiSelectSearch };
--- a/packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.module.css
+++ b/packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.module.css
@@ -1,0 +1,141 @@
+.root {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  gap: var(--gap--16);
+  justify-content: flex-end;
+  width: auto;
+  min-height: 6rem;
+  color: var(--color-text--tertiary);
+}
+
+.selected {
+  display: flex;
+  flex-flow: row wrap;
+  gap: var(--space--8);
+}
+
+.search {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  width: auto;
+  height: 2.75rem;
+  color: var(--color-text--tertiary);
+}
+
+.inputContainer {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  width: 100%;
+  border: var(--border-width--2) solid var(--border-color);
+
+  border-radius: var(--border-radius--8);
+  background-color: var(--color-surface--4);
+
+  --border-color: transparent;
+
+  transition: ease-out 300ms;
+}
+
+.input {
+  width: 100%;
+  margin: var(--space--10) var(--space--14);
+  font-weight: var(--font-weight-medium);
+  font-size: var(--font-size--l);
+  line-height: normal;
+  background-color: transparent;
+}
+
+.inputContainer:hover {
+  --border-color: var(--color-border--highlight);
+}
+
+.inputContainer:focus-within {
+  color: var(--color-text--default);
+  background-color: var(--color-black);
+
+  --border-color: var(--color-border--highlight);
+}
+
+.input::placeholder {
+  color: var(--color-text--tertiary);
+}
+
+.inputContainer[data-color="red"] {
+  --border-color: var(--color-red--5);
+}
+
+.inputContainer[data-color="red"]:hover {
+  --border-color: var(--color-red--3);
+}
+
+.inputContainer[data-color="green"] {
+  --border-color: var(--color-cyber-green--50);
+}
+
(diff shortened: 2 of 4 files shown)


Oksamies:
Note that this component is being improved in #920 . I'm looking for tips and opinions

apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
+  const [selected, setSelected] = useState(options);
+  const defaultProps = {
+    ...args,
+    onChange: (x: { label: string; value: string }[]) => setSelected(x),
+  };
+  return (
+    <div>
+      <div style={{ color: "white" }}>
```
anttimaki, line 40: Not sure this is needed given that the component itself shows the state too. But then again the implementation currently double-tracks the state (more about that on other comments), so maybe this can be used for debugging.

apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+        Value in state:{" "}
+        {selected.map((s) => {
+          return s.label;
+        })}
```
anttimaki, line 44: `        }).join(", ")}` would be a bit nicer

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.module.css
```diff
@@ -0,0 +1,141 @@
...
+}
+
+.inputContainer[data-color="green"]:hover {
+  --border-color: var(--color-cyber-green--80);
+}
+
+.removeAllButton {
+  width: 3rem;
+  height: 100%;
+  color: #c6c3ff;
```
anttimaki, line 85: Just a thought but could the button have a different color when selections have been made? To make it more noticeable?

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { Icon } from "../Icon/Icon";
+import { classnames } from "../../utils/utils";
+import { Button } from "../../index";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import {
+  faCaretDown,
+  faCircleXmark,
+  faXmark,
+} from "@fortawesome/pro-solid-svg-icons";
+import { useRef } from "react";
```
anttimaki, line 13: Would be clearer if combined with L2

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <div
+            className={classnames(
+              styles.menu,
+              isVisible ? styles.visible : null
+            )}
+          >
+            {filteredOptions.map((option, key) => {
+              return (
+                <MultiSelectItem
+                  key={key}
```
anttimaki, line 170: Ideally [the index shouldn't be used as the key](https://react.dev/learn/rendering-lists#why-does-react-need-keys). If we can guarantee that `option.value` can be unique within this `.map`, it would make a nice key.

anttimaki (changes requested):
Most of the stuff is not that important, but I would consider fixing at least the ones that are straightforward and quick to fix.

By far the most important comment is the one regarding the double state management. If there's no need to duplicate it, I would change the implementation. Marking as request changes because of this issue, but I can probably be swayed with good arguments.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.module.css
Oksamies: Implementing as in designs, so no.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+type Option = {
+  label: string;
+  value: string;
+};
+type Props = {
+  options: Option[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onChange: (v: any) => void;
+  onBlur: () => void;
+  disabled?: boolean;
```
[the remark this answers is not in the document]
Oksamies: It's TODO. Added a TODO comment.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+  name: string;
```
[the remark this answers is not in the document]
Oksamies: It's in use now

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+ * Cyberstorm MultiSelectSearch component
+ *  TODO: Better keyboard navigation
+ */
+export const MultiSelectSearch = React.forwardRef<HTMLInputElement, Props>(
+  function MultiSelectSearch(props, ref) {
+    const { options, onChange, onBlur, placeholder, color } = props;
+    const menuRef = useRef<HTMLDivElement | null>(null);
+    const inputRef = useRef<HTMLInputElement | null>(null);
+
+    const [isVisible, setIsVisible] = React.useState(false);
```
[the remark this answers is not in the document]
Oksamies: I think isVisible is better, as the boolean switches the visibility on and off.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const [search, setSearch] = React.useState("");
+    const [selected, setList] = React.useState<Option[]>([]);
```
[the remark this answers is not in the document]
Oksamies: Removed this useState

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+    React.useEffect(() => {
+      document.addEventListener("mousedown", closeOpenMenu);
+      document.addEventListener("touchstart", closeOpenMenu);
+      return () => {
+        document.removeEventListener("mousedown", closeOpenMenu);
+        document.removeEventListener("touchstart", closeOpenMenu);
+      };
+    });
+
+    React.useEffect(() => {
```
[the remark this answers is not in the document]
Oksamies: Removed the onChange

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+      onChange(selected);
+      const updatedOptions = options.filter(
+        (option) => !selected.includes(option) && option.label.includes(search)
```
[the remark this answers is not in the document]
Oksamies: Good point

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+              className={styles.input}
+              value={search}
+              data-color={color}
+              onFocus={() => setIsVisible(true)}
+              onChange={(e) => setSearch(e.currentTarget.value)}
+              ref={inputRef}
+              placeholder={placeholder}
+            />
+            <button
+              onClick={(e) => handleParentClick(e, () => setList([]))}
```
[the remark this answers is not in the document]
Oksamies: Yeah, should just clear input, changed to just clear input.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,205 @@
...
+                handleParentClick(e, () => setIsVisible(!isVisible))
+              }
+              className={styles.openMenuButton}
+            >
+              <Icon inline>
+                <FontAwesomeIcon icon={faCaretDown} />
+              </Icon>
+            </button>
+          </div>
+          <div
```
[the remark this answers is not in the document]
Oksamies: Decided to go with showing an empty dropdown instead of hiding it. Found a little less confusing to have an empty list, instead of nothing happening on screen. In the end designs decide though.

apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  { label: "Team 14", value: "Team 14" },
+];
+
+const defaultArgs = {
+  placeholder: "Select something",
+  options: options,
+};
+
+const Template: StoryFn<typeof MultiSelectSearch> = (args) => {
+  const [selected, setSelected] = useState(options);
```
anttimaki, line 33: This now sets all options selected initially, so the user has nothing to choose from before they unselect something. Not the best demo case.

apps/cyberstorm-storybook/stories/components/MultiSelectSearch.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    ...args,
+    onChange: (x: { label: string; value: string }[]) => setSelected(x),
+    value: selected,
+  };
+  return (
+    <div>
+      <div style={{ color: "white" }}>
+        Value in state:{" "}
+        {selected
+          .map((s) => {
```
anttimaki, line 44: Nitpick: `.map((s) => s.label)`

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,191 @@
...
+          !value.includes(option) &&
+          option.label.toLowerCase().includes(search.toLowerCase())
+      );
+      setFilteredOptions(updatedOptions);
+    }, [value, search, options, setFilteredOptions]);
+
+    return (
+      <div className={styles.root} ref={ref}>
+        <div className={styles.selected}>
+          {value.map((option) => {
```
anttimaki, line 89: Currently the options are listed in the order they are selected. IDK if it would be more user friendly if they were ordered alphabetically? Something to think about e.g. when fixing the other TODOs later on?

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,191 @@
...
+          onFocus={(e) => {
+            inputRef.current && inputRef.current.focus();
+            e.stopPropagation();
+          }}
+          role="button"
+          tabIndex={0}
+          ref={menuRef}
+        >
+          <div className={styles.inputContainer} data-color={color}>
+            <input
```
anttimaki, line 116: On Storybook, at least with Firefox, the tab navigation will now get stuck into this input, not moving forward no matter how many times I tab.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
```diff
@@ -0,0 +1,191 @@
...
+              }}
+              className={styles.clearSearch}
+            >
+              <Icon inline>
+                <FontAwesomeIcon icon={faCircleXmark} />
+              </Icon>
+            </button>
+            <div className={styles.inputButtonDivider}></div>
+            <button
+              onClick={(e) => {
```
anttimaki, line 138: On Storybook, at least with Firefox, when the menu is invisible, I'd expect clicking this would make it visible. But the menu stays visible only for a fraction of a second, then it closes and the focus moves into the input.

packages/cyberstorm/src/utils/type_guards.ts
```diff
@@ -3,3 +3,10 @@ export const isRecord = (obj: unknown): obj is Record<string, unknown> =>
 
 export const isStringArray = (arr: unknown): arr is string[] =>
   Array.isArray(arr) && arr.every((s) => typeof s === "string");
+
+export const isNode = (e: EventTarget | null): e is Node => {
+  if (!e || !("nodeType" in e)) {
```
anttimaki, line 8: Nitpick. I suspect you've fallen victim to one of the classic blunders, namely only fixing the line mentioned in a comment and disregarding the context. Now that this no longer throws an error, it would make more sense to be just
```
export const isNode = (e: EventTarget | null): e is Node =>
  e !== null && "nodeType" in e;
```

anttimaki (approved):
I've understood there's some pressure to get this merged so I'll mark this as approved, but I think the two usability problems reproducable on Storybook/FF should be fixed ASAP.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
Oksamies: Added sorting 👍

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
Oksamies: Fixed this issue, but I'll keep the keyboard navigation todo there, because I'm not 100% what is the best practice, so we need to fully fix the keyboard nav later.

packages/cyberstorm/src/components/MultiSelectSearch/MultiSelectSearch.tsx
Oksamies: Probably was broken because of the focus fuckery in the last comment you made, should be fixed now 👍

PR #924, opened 2023-11-24
@thunderstore/cyberstorm: add NewTabs

anttimaki:
@thunderstore/cyberstorm: add NewTabs

Add an improved version Tabs component. The old Tabs component is
retained for now to avoid scope creep. Ideally eventually all of the
old components will be replaced, the old implementation removed and
NewTabs renamed to Tabs.

The main motivation for the new implementation is an usecase from
PackageDetailLayout. Dapper will initally load only content for the
default tab. Additionally it will load flags defining whether other
tabs have content. E.g. if the changelog tab has no content, the tab is
disabled. If changelog exists, it's loaded with a separate Dapper call.

Additional improvements include:

- Moving away from "children as props" approach
- Ease of use when developer doesn't have to provide the tab buttons
  and contents separately
- Ease of use when the Tabs component will internally handle switching
  between the tabs
- Probably cleaner support for Suspenses

Refs TS-1979

@thunderstore/cyberstorm-nextjs: add stories for NewTabs

Refs TS-1979

@thunderstore/cyberstorm: use NewTabs in PackageDetailLayout

This is in preparation for how the data will be fetched to different
Tabs with Dapper.

Refs TS-1979

### DIFF
--- a/apps/cyberstorm-storybook/stories/components/NewTabs.stories.tsx
+++ b/apps/cyberstorm-storybook/stories/components/NewTabs.stories.tsx
@@ -1,0 +1,48 @@
+import { StoryFn, Meta } from "@storybook/react";
+import { NewTabs } from "@thunderstore/cyberstorm";
+import { usePromise } from "@thunderstore/use-promise";
+import { Suspense } from "react";
+import { faCog, faFlag, faThumbsUp } from "@fortawesome/pro-solid-svg-icons";
+
+const meta = {
+  title: "Cyberstorm/Components/Tabs",
+  component: NewTabs,
+} as Meta<typeof NewTabs>;
+
+const promiseLoader = () =>
+  new Promise<void>((resolve) => setTimeout(() => resolve(), 2000)).then(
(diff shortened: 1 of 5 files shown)


anttimaki:
* master
  * **PR #924** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/924?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/924?utm_source=stack-comment).

anttimaki:
@Oksamies this btw uses the "namespaced components" approach. IIRC you mentioned there was some problems with it, what were they again?

Oksamies (approved):
Realized I forgot to use the `export default` and `import {default as X}`, when I tried the namespaced components, so that's probably why those didn't work for me.

Anyways this looks good to me 👍

PR #934, opened 2023-11-27
Minor but big UI effects

Oksamies:
NewTabs bottom border to be full width
Make package listings pages filter sidebar sticky
Fix WrapperCard padding

refs -

### DIFF
--- a/packages/cyberstorm/src/components/Layout/BaseLayout/BaseLayout.module.css
+++ b/packages/cyberstorm/src/components/Layout/BaseLayout/BaseLayout.module.css
@@ -14,7 +14,7 @@
   --w: 56vw;
 
   position: absolute;
-  top: calc(var(--header-height) - var(--header-height) *2);
+  top: calc(-1 * var(--header-height));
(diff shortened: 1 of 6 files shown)


packages/cyberstorm/src/components/Layout/BaseLayout/BaseLayout.module.css
```diff
@@ -14,7 +14,7 @@
   --w: 56vw;
 
   position: absolute;
-  top: -3.5rem;
+  top: calc(var(--header-height) - var(--header-height) *2);
```
anttimaki, line 17: Why not just `calc(-1 * var(--header-height))` ?

anttimaki (approved):

packages/cyberstorm/src/components/Layout/BaseLayout/BaseLayout.module.css
Oksamies: That's advanced math

PR #938, opened 2023-11-30
Prerequisites for upcoming forms and actions

Oksamies:


packages/cyberstorm-forms/src/components/FormSelect.module.css
```diff
@@ -0,0 +1,64 @@
+.SelectTrigger {
```
anttimaki, line 1: This should use camelCase, not PascalCase, to be consistent with the Cyberstorm package. Applies to other selectors in this files as well.

EDIT: to think of it, these are all used in Select component, so I need no reason to use the `Select` prefix in the names at all.

packages/cyberstorm-forms/src/components/FormSelect.module.css
```diff
@@ -0,0 +1,64 @@
...
+  display: inline-flex;
```
anttimaki, line 2: Also generally speaking I feel this file doesn't adhere to all the style things we've set up so far and I'm not sure why you're actively creating code debt by ignoring them. I'm talking about variables for colors, sizes, spaces etc.

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
+"use client";
+
+import { z, ZodObject } from "zod";
+import { ZodRawShape } from "zod/lib/types";
```
anttimaki, line 4: Per my earlier review, this should be imported from "zod".

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+export type FormSelectProps<
+  Schema extends ZodObject<Z>,
+  Z extends ZodRawShape
+> = {
+  // The schema is required to allow TS to infer valid values for the name field
+  schema: Schema;
+  name: Path<z.infer<Schema>>;
+  options: { value: string; label: string }[];
```
anttimaki, line 20: Nice to see you've chosen to use the value/label approach instead of a single string this time around. Although it does make the `SelectSear` stick out like a sore thumb.

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+  placeholder?: string;
```
anttimaki, line 21: This is never passed to `TempSelect`?

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+  defaultValue?: string;
+  ariaLabel?: string;
```
anttimaki, line 23: Should we make this required?

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+  options,
+  ariaLabel,
+}: FormSelectProps<Schema, Z>) {
+  const {
+    field,
+    fieldState: { error },
+    formState: { isSubmitting, disabled },
+  } = useController({ name });
+
+  if (defaultValue) {
```
anttimaki, line 37: I'm not familiar with the whole form contraption we've got going on, but I feel that setting a default value as an effect, field by field, in a possibly large form, is not the approach we want to take. On a quick glance RHF supports setting defaults once per form. If possible, I think that approach should be used. @MythicManiac feel free to chip in.

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+    React.useEffect(() => {
+      field.onChange(defaultValue);
+    }, [defaultValue]);
+  }
+
+  return (
+    <>
```
anttimaki, line 44: Per my earlier review, this should probably be `div` to keep the field and error together. Don't know which approach you ended up taking in the earlier PR.

packages/cyberstorm-forms/src/components/FormSelect.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <TempSelect
```
anttimaki, line 45: Nitpick: not fan of the name since there's nothing temporary about this.

Main point: is the wrapper even needed? I appreciate the idea of breaking down a large component to subcomponents, but in this particular case I feel it goes too far, and adding the TempSelect-in-the-middle component just makes things unnecessarily complicated.

packages/cyberstorm-forms/src/components/FormSubmitButton.tsx
```diff
@@ -23,16 +23,44 @@ const SubmitButtonContent = React.memo(
...
 SubmitButtonContent.displayName = "SubmitButtonContent";
 
-export function FormSubmitButton({ text }: { text: string }) {
+export function FormSubmitButton({
+  text,
+  colorScheme = "accent",
+  icon,
+}: {
+  text: string;
+  colorScheme?:
```
anttimaki, line 32: DRY: `colorScheme?: React.ComponentProps<typeof Button.Root>["colorScheme"];`

packages/cyberstorm-forms/src/components/FormTextInput.tsx
```diff
@@ -13,17 +14,24 @@ export type FormTextInputProps<
...
   Z extends ZodRawShape
->({ name, placeholder }: FormTextInputProps<Schema, Z>) {
+>({ name, placeholder, existingValue }: FormTextInputProps<Schema, Z>) {
   const {
     field,
     fieldState: { isDirty, invalid, error },
     formState: { isSubmitting, disabled },
   } = useController({ name });
 
+  if (existingValue) {
```
anttimaki, line 29: As stated in a comment for `FormSelect`, I'd prefer this would be handled via RHF rather than manually in each component. If that's not viable, at least rename the prop to defaultValue to be consistent with prop name of `FormSelect`

packages/cyberstorm/package.json
```diff
@@ -28,7 +28,7 @@
     "@radix-ui/react-dialog": "^1.0.2",
     "@radix-ui/react-dropdown-menu": "^2.0.1",
     "@radix-ui/react-radio-group": "^1.1.3",
-    "@radix-ui/react-select": "^1.2.0",
+    "@radix-ui/react-select": "^2.0.0",
```
anttimaki, line 31: IDK how Radix uses semver but is it wise to mix components from v2 with components from v1?

packages/cyberstorm/src/components/Dialog/Dialog.module.css
```diff
@@ -18,8 +17,10 @@
   border: var(--border);
   border-radius: var(--border-radius--8);
   background-color: var(--color-surface--2);
-  box-shadow: var(--box-shadow-default);
-  transform: translate(calc(var(--left) * -1), calc(var(--top) * -1));
+
+  box-shadow: hsl(206deg 22% 7% / 0.35) 0 10px 38px -10px, hsl(206deg 22% 7% / 0.2) 0 10px 20px -15px;
+  transform: translate(-50%, -50%);
```
anttimaki, line 22: A bit offtopic but is this still the preferred way of centering stuff? I would have thought there's something better by now. @gkurck

packages/cyberstorm/src/components/Header/Header.module.css
```diff
@@ -1,7 +1,7 @@
 .root {
   position: sticky;
   top: 0;
-  z-index: 999;
+  z-index: auto;
```
anttimaki, line 4: Isn't `auto` the default value? Is this actually needed at all?

packages/ts-api-react-forms/src/ApiAction.tsx
```diff
@@ -0,0 +1,52 @@
+"use client";
```
anttimaki, line 1: This whole thing confuses me. According to commit message, the idea is to "use API without forms". So why is this located in `ts-api-react-forms` package? Why not `ts-api-react`, or better yet directly in `thunderstore-api`?

Secondly, what seems to me to basically be a "onChange" function is for some reason formatted as a React component? Why?

This uses zod, which **might** be a good idea, but on the other hand **might** be overkill. It was used on the forms since RHF supports it and helps reduce boilerplate(?). Now were not using RHF. Would regular TS interfaces do the same job?

Also I'm not a fan of the `metaData`, as a variable name nor as a concept. YMMV but if you compare this to Dapper, everything above `thunderstore-api` methods treats data as "everything that is needed to make the request, defined by the ts-api method arguments". It's then up to the method to decide what parts of information goes to URL and what goes to query parameters (or request body, although no such use case has been created by me). This way Cyberstorm components don't (need to) know how the data is used, they're just responsible to providing it all.

In summary: I would cut out out `ts-api-react-forms`/`ts-api-react` and go either directly to `thunderstore-api` or possibly through a separate `ts-api-actions` package. That being said, I feel a lot of the above might be due to me not comprehending the inner workings of `ts-api-react-forms`/`ts-api-react` so you might want a second opinion from @MythicManiac

packages/ts-api-react-forms/src/ApiAction.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useApiAction } from "./useApiAction";
+
+export interface ApiActionProps<
+  Schema extends ZodObject<Z>,
+  Result extends object,
+  Z extends ZodRawShape
+> {
+  schema: Schema;
+  endpoint: ApiEndpoint<z.infer<Schema>, Result>;
+  metaData: any;
```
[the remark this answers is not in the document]
anttimaki: What ever happens to this whole file, as per my earlier reviews I would prefer a) better typing b) silencing the warning and adding a comment describing why it was necessary. I like this approach since writing the comment usually (not necessarily in this case) works as rubber ducking, and makes it obvious to others that the choice to use `any` was deliberate and necessary.

Needless to say, this comment applies to all these warnings.

anttimaki (commented):
Leaving this as a comment since I think we should discuss the two main points of the review before moving forward:

- The ApiAction as a concept
- Setting default values to form components

packages/ts-api-react-forms/src/ApiAction.tsx
anttimaki: Forgot to elaborate on the naming of `metaData`: It's one word, so `metadata`. Even then I'm not sure if it's really "data about the rest of the data". It's additional data, and in this care very much required to form the complete set of "data".

packages/cyberstorm-forms/src/components/FormSelect.tsx
Oksamies: We'll need a way to update the value in the field later on, when we are not reloading the page. But that is then, so I'll remove this.

There was a problem with using the defaultValue solely, which I can't remember now, but I'll figure that one out.

packages/cyberstorm-forms/src/components/FormSelect.tsx
Oksamies: Ended up going with those "no element" way, so that the elements can be easily controlled in the form level.

If you have some example case where having these wrapped in a div would be useful; let me know.

anttimaki: Well I don't really have one, I guess it depends on the CSS used in the forms themselves. Just keep an eye out: if you end up wrapping all the elements in div, it should probably be done in the component rather than repeated on the outside of it.

packages/cyberstorm/package.json
Oksamies: From what I could find, it's alright to mix them.

packages/cyberstorm/src/components/Dialog/Dialog.module.css
gkurck: Yes, flex is preferred here, no need to take the element out of the document flow.

PR #940, opened 2023-11-30
Add AddTeamMemberForm

Oksamies:


packages/cyberstorm-forms/src/forms/AddTeamMemberForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+  useFormToaster,
+} from "@thunderstore/cyberstorm-forms";
+import { FormSelect } from "../components/FormSelect";
+
+export function AddTeamMemberForm(props: {
+  dialogOnChange: (v: boolean) => void;
+  teamName: string;
+}) {
+  const { onSubmitSuccess, onSubmitError } = useFormToaster({
+    successMessage: `Member TODO added to ${props.teamName}`,
```
anttimaki, line 21: Any reason this is left as TODO? Why can't we read the member name from the form data?

packages/cyberstorm-forms/src/forms/AddTeamMemberForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+              schema={teamAddMemberFormSchema}
+              name={"user"}
+              placeholder={"Enter username..."}
+            />
+          </div>
+          <FormSelect
+            schema={teamAddMemberFormSchema}
+            name={"role"}
+            options={roleOptions}
+            defaultValue="member"
```
anttimaki, line 58: Could use `roleOptions[0]["value"]` instead of a magic string.

packages/cyberstorm/src/components/Layout/Teams/TeamSettings/TeamMembers/TeamMembers.tsx
```diff
@@ -24,47 +18,33 @@ export function TeamMembers(props: Props) {
...
-        </div>
-        <div className={styles.dialogFooter}>
-          <Button.Root colorScheme="success">
-            <Button.ButtonLabel>Add Member</Button.ButtonLabel>
-          </Button.Root>
-        </div>
-      </>
-    </Dialog.Root>
-  );
+  const [dialogOpen, setOpenDialog] = useState(false);
```
anttimaki, line 21: By convention `setOpenDialog` should be `seDialogOpen`.

packages/thunderstore-api/src/fetch/teamAddMember.ts
```diff
@@ -0,0 +1,28 @@
+import { RequestConfig } from "../index";
```
anttimaki, line 1: This comes a bit late since `teamCreate.ts` already exists, but my original idea was that the `fetch` directory here would be for stuff that fetches data (not stuff that uses `fetch` function). Now the files like `teamAddMember` and `teamMembers` are a bit confusingly named. I wouldn't mind splitting the files to e.g. `fetch`, `submit` and `action` directories. Not a blocker now since refactoring should be rather easy as long as everything is exported via `src/index.ts`.

packages/thunderstore-api/src/fetch/teamAddMember.ts
```diff
@@ -0,0 +1,28 @@
...
+import { apiFetch2 } from "../apiFetch";
+
+export type teamAddMemberMetaData = {
```
anttimaki, line 4: No reason export this or `teamAddMemberApiArgs`.

packages/thunderstore-api/src/fetch/teamAddMember.ts
```diff
@@ -0,0 +1,28 @@
...
+  teamIdentifier: string;
+};
+
+export type teamAddMemberApiArgs = {
+  user: string;
```
anttimaki, line 9: Could be `username` to make it more clear?

packages/thunderstore-api/src/fetch/teamAddMember.ts
```diff
@@ -0,0 +1,28 @@
...
+  role: string;
```
anttimaki, line 10: Could be typed as `member` | `owner`

packages/thunderstore-api/src/fetch/teamAddMember.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export function teamAddMember(
+  config: RequestConfig,
+  data: teamAddMemberApiArgs,
```
anttimaki, line 15: As mentioned in my review for #938 I'm not sure I like the split between "data" and "metadata". Especially for these small forms, might be a different story for a form with 10 fields, in which case having `data` (or `formData`) as a separate object might make sense. Even then I would pass whatever is in "metadata" as separate args.

packages/ts-api-react-forms/src/schema.ts
```diff
@@ -5,3 +5,12 @@ export const createTeamFormSchema = z.object({
     .string({ required_error: "Team name is required" })
     .min(1, { message: "Team name is required" }),
 });
+
+export const teamAddMemberFormSchema = z.object({
+  user: z
+    .string({ required_error: "Username is required" })
+    .min(1, { message: "Username is required" }),
+  role: z
```
anttimaki, line 13: Checking the length for a select makes little sense. And I think we have a default value so even the required-check is probably not needed (although I guess a completionist might leave the latter just to be sure).

PR #942, opened 2023-11-30
Add TeamDetailsEdit form

Oksamies:


anttimaki (approved):
I did not repeat the comments from the earlier reviews for the copy-pasted code here, but I expect you to do any relevant fixes here too.

No hard blockers, but please consider fixing them anyway.

PR #962, opened 2023-12-05
Add metaData prop to ApiForm

Oksamies:
metaData props purpose is to carry data that
should not be inserted into the payload,
to the fetch function

First use-case is carryin the teamName to
fetch function so that it can be inserted into
the url

### DIFF
--- a/packages/cyberstorm-forms/src/forms/CreateTeamForm.tsx
+++ b/packages/cyberstorm-forms/src/forms/CreateTeamForm.tsx
@@ -21,6 +21,7 @@ export function CreateTeamForm() {
     <ApiForm
       {...toaster}
       schema={createTeamFormSchema}
+      meta={{}}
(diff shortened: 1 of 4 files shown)


Oksamies:
* master
  * **PR #962** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/962?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/962?utm_source=stack-comment).

anttimaki (commented):
I've not much to add to my earlier review. Passing around an undocumented "meta" parameter may be confusing to developers. If that's the approach we want to take anyway, then I guess this PR seems okay.

PR #989, opened 2024-01-02
Add CommunityList skeleton component

Oksamies:
refs TS-1822

PR #991, opened 2024-01-02
Add Skeletons for components in Community view

Oksamies:


Oksamies:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/991?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `master`
  * **PR #989** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #990** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/990?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #991** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/991?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/991?utm_source=stack-comment).

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,81 +24,133 @@ interface Props {
...
- *       adding another wrapping component.
- */
-export function PackageList(props: Props) {
+export function PackageList(props: PackageListProps) {
   const { categories, deprecated, listingType, nsfw, searchQuery, section } =
     props;
 
   const [order, setOrder] = useState(PackageOrderOptions.Updated);
   const [page, setPage] = useState(1);
+  const [packageCount, setPackageCount] = useState<number | null>(null);
```
anttimaki, line 33: This implementation doesn't seem to work if:

1. The search returns no results (`PackageCount` and `Pagination` components are not rendered, but keep showing the skeleton components)
2. The values are returned from the cache of `usePromise`, at least that's what I think causes it. The issue goes something like this:
    1. Load the initial package list, everything works
    2. Toggle a category filter "include" -> "exclude" -> "off"
    3. Packages are shown as expected, but `PackageCount` and `Pagination` are not

(And TBH feels hacky to start with.)

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,81 +24,133 @@ interface Props {
...
 
-  const deferredCategories = useDeferredValue(categories);
```
anttimaki, line 39: 1. FYI, the deferred values were used originally because there were errors that the slow fetching of packages blocked the rest of the UI without them. I didn't see the errors now, so I guess something had changed and they might not be needed anymore.
2. Getting rid off them to get the Suspense fallback show again might be ok, but from what I understood that's frowned upon and the best practice would be to use the fallbacks only when loading the initial data. After that updates should be indicated in a more subtle manner. See https://react.dev/reference/react/Suspense#showing-stale-content-while-fresh-content-is-loading and https://react.dev/reference/react/useDeferredValue for more info. Again, if we truly want to fallback to the skeleton components, I guess it's okay to do this from the technical perspective, as long as it doesn't block the rest of the UI.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,81 +24,133 @@ interface Props {
...
 
         <PackageOrder order={order} setOrder={setOrder} />
       </div>
 
       <div className={styles.packages}>
-        {packages.results.map((p) => (
-          <PackageCard key={`${p.namespace}-${p.name}`} package={p} />
-        ))}
+        <Suspense
+          fallback={range(1, 13).map((x) => (
```
anttimaki, line 67: Why 13? Could we maybe use `PER_PAGE` here to avoid page jumping around when user uses the pagination at the bottom? (Obviously it still happens on the last page but we can't know how many items there will be so it can't be helped.)

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,81 +24,133 @@ interface Props {
...
+            <SkeletonBox key={x} className={styles.packageCardSkeleton} />
+          ))}
+        >
+          <PackageListResults
+            listingType={listingType}
```
anttimaki, line 72: Could use `{...props}` to avoid some boilerplate.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,81 +24,133 @@ interface Props {
...
+      ) : (
+        <SkeletonBox className={styles.packageCountSkeletonBar} />
+      )}
     </div>
   );
 }
 
 PackageList.displayName = "PackageList";
+
+interface PackageListResultsProps {
```
anttimaki, line 103: Could extend `PackageListProps`

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
```diff
@@ -53,20 +55,29 @@ export function PackageSearch(props: Props) {
 
       <div className={styles.contentWrapper}>
         <div className={styles.sidebar}>
-          <SectionMenu
-            allSections={allSections}
-            selected={section}
-            setSelected={setSection}
-          />
+          <Suspense
```
anttimaki, line 58: This still suffers from the same problem as the previous version of this PR. `allSections` and `categories` are loaded outside the suspense, so the fallback is never triggered.

BUT it's also a bit different case, since it *seems to me* the values are *currently* not loaded on the client side at all, only the SSR content is used. Which may or may not be what we actually want (better performance with cache vs. new categories taking time to be shown on the site). *If* we are happy with the SSR-only, this `Suspense` is probably not needed at all.

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
```diff
@@ -53,20 +55,29 @@ export function PackageSearch(props: Props) {
...
+            fallback={range(1, 6).map((x) => (
```
anttimaki, line 59: This "render x SkeletonBoxes" seems to be a recurring pattern. Could be nice to have something like `<SkeletonList items="6" className={styles.sideBarSkeletonBar} />` that would wrap the range-stuff?

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
```diff
@@ -53,20 +55,29 @@ export function PackageSearch(props: Props) {
...
+              setCategories={setCategories}
+            />
 
-          <OthersMenu
-            deprecated={deprecated}
-            setDeprecated={setDeprecated}
-            nsfw={nsfw}
-            setNsfw={setNsfw}
-          />
+            <OthersMenu
```
anttimaki, line 74: Regardless of whether `SectionMenu` and `CategoryMenu` remains inside the `Suspense`, `OthersMenu` doesn't depend on any fetched data, so it doesn't need to be wrapped inside the `Suspense`. Unless we deliberately want to hide it until the other menus are loaded, that is. Could be worthwhile to have a discussion and come up with a site-wide practice for this.

anttimaki (changes requested):
Overall I feel the changes to `PackageList.tsx` are overcomplicating things and unnecessary. Unless there's a specific reason, I would urge you to consider just wrapping the whole `<PackageList>` component in `PackageSearch` with a Suspense. This means the fallback component needs to be a bit more complicated but overall I think it's the simpler and better approach.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
Oksamies: We don't have it in the designs, but I guess covering the stale data with some mist and a loading icon, would make more sense.

Oksamies: Decided to just add some opacity

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
Oksamies: I've went with the thought of; Suspense it if it loads something.

So with that, you're right in that it can be left outside. We can revisit this, when reviewing the page 👍

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
Oksamies: Ah, truuue. I just checked the designs and added skeletons to where ever they were in designs.

IMO the sidebar doesn't need a skeleton.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -22,13 +23,6 @@ interface Props {
 
 const PER_PAGE = 20;
 
-/**
- * Fetches packages based on props and shows them as a list.
```
anttimaki, line 26: Nitpick: I'm not saying this comment was very helpful, but I'd argue it was more helpful than the generic copy pasted "Cyberstorm X component" comments we're allowing to exists :D

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -69,32 +63,44 @@ export function PackageList(props: Props) {
     deferredDeprecated,
   ]);
 
+  const isNotStale =
```
anttimaki, line 66: Nitpick: I would change this to `const isStale = deferredOrder !== order || [...]`

Using `&&` requires checking all the values, while `||` can exit early when the first changed value is detected.

It would also change `isStale={!isNotStale}` to `isStale={isStale}` which is more readable.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
```diff
@@ -69,32 +63,44 @@ export function PackageList(props: Props) {
...
+    deferredOrder === order &&
+    deferredPage === page &&
+    deferredSearchQuery === searchQuery &&
+    deferredCategories === categories &&
+    deferredSection === section &&
+    deferredNsfw === nsfw &&
+    deferredDeprecated === deprecated;
+
   return (
     <div className={styles.root}>
```
anttimaki, line 76: Nitpick: I would consider using the `StaleCover` over the whole component instead of individual subcomponents.

* It would be less repetition (not a huge problem here but can be on future components with more subcomponents, and it would be nice to future proof this be consistent site-wide)
* If we add some spinner icon oslt it will be easier to position e.g. to center of the area. This ofc is irrelevant if we won't add one.
* It would cover the PackageOrder - I personally can't decide if I like it more or less but it's a matter of opinion I think

packages/cyberstorm/src/components/PackageSearch/PackageSearch.tsx
```diff
@@ -17,6 +17,8 @@ import styles from "./PackageSearch.module.css";
 import { CategorySelection } from "./types";
 import { PackageList } from "../PackageList/PackageList";
 import { TextInput } from "../TextInput/TextInput";
+import { PackageListSkeleton } from "../PackageList/PackageListSkeleton";
+import { SkeletonBoxes } from "../SkeletonBoxes/SkeletonBoxes";
```
anttimaki, line 21: Highlighting the defined but unused linting failure.

packages/cyberstorm/src/components/SkeletonBoxes/SkeletonBoxes.tsx
```diff
@@ -0,0 +1,23 @@
+import { range } from "../../utils/utils";
+import { SkeletonBox } from "../SkeletonBox/SkeletonBox";
+
+/**
+ * Cyberstorm SkeletonList component
```
anttimaki, line 5: Nitpick: misleading (and frankly useless) comment.

packages/cyberstorm/src/components/StaleCover/StaleCover.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { classnames } from "../../utils/utils";
+
+interface Props {
+  children: ReactNode;
+  className?: string;
+  isStale: boolean;
+}
+
+/**
+ * Component for covering children while their content is stale
```
anttimaki, line 12: :+1: for actually writing something usefull here.

packages/cyberstorm/src/components/StaleCover/StaleCover.tsx
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function StaleCover(props: Props) {
```
anttimaki, line 14: Nitpick: technically this doesn't cover the component (like e.g. a modal background layer would) since it's rendered below the component. `StaleIndicator` or `StalenessIndicator` oslt could be more descriptive.

packages/cyberstorm/src/components/PackageList/PackageList.tsx
Oksamies: Didn't mean to remove this 🤦 I'll add it back

packages/cyberstorm/src/components/PackageList/PackageList.tsx
Oksamies: Yeah, true! It was so late when I was doing this, that I couldn't figure out quickly on how to do this better, so I just made it "isXobject" styled check and flipped the variable name.

This is nice, now it seems obvious but back then just couldn't gather it from my thoughts :D

packages/cyberstorm/src/components/PackageList/PackageList.tsx
anttimaki: > `PackageOrder` has to be available, even if the other stuff is loading.

Does it though?

Oksamies:
### Merge activity

* **Jan 8, 7:52 AM**: @Oksamies started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/991).

PR #994, opened 2024-01-03
@thunderstore/dapper: support deactivated dependencies

anttimaki:
@thunderstore/dapper: support deactivated dependencies

Since deactivated dependencies are still required to use a mod, backend
will return them with the rest of the dependencies from now on. New
field is_active is returned with the data. Since the reason for
deactivation might be a gross icon or description text, null and a
placeholder text is returned instead.

Refs TS-2013

@thunderstore/cyberstorm: support deactivated dependencies

To accommodate for the fact that deactivated dependencies will return
a null instead of icon URL:

- Use ImageWithFallback component instead of img element
- Add support for package icon type to ImageWithFallback
- Add support for square images to ImageWithFallback

These changes were done since it's the minimum required to prevent the
build from breaking. The following changes were considered out of scope
for this task:

- Components currently try to link to the deactivated packages, but
  their name should be shown as plain text instead
- Deactivated dependencies should have "Removed" tag on the sidebar
  of the PackageDependencyList and PackageDependencyDialog

The description for the deactivated dependencies is also censored by
the backend, but that requires no action from the client.

Refs TS-2013

### DIFF
--- a/packages/cyberstorm/src/components/ImageWithFallback/ImageWithFallback.module.css
+++ b/packages/cyberstorm/src/components/ImageWithFallback/ImageWithFallback.module.css
@@ -1,3 +1,17 @@
+.isSquare {
+  --aspect-ratio: 1;
+  --image-width: 50%;
+}
+
+.is3By4 {
+  --aspect-ratio: 3 / 4;
+  --image-width: 30%;
+}
+
+.fullWidth {
+  --image-width: 100%;
+}
+
 .imageWrapper {
   display: flex;
   align-items: center;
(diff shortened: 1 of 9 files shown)


anttimaki:
* `master`
  * **PR #994** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/994?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/994?utm_source=stack-comment).

Oksamies (approved):

PR #998, opened 2024-01-09
WIP nextjs deploy testing

Oksamies:


Oksamies:
* `master`
  * **PR #998** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/998?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/998?utm_source=stack-comment).

apps/cyberstorm-nextjs/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN chown node:node -R /app
 
 USER node
 RUN --mount=type=secret,id=npmrc,target=/home/node/.npmrc,uid=1000 yarn install --frozen-lockfile
- RUN yarn workspace @thunderstore/cyberstorm-nextjs run build
+ RUN yarn workspace @thunderstore/cyberstorm-nextjs run build:dev
```
anttimaki, line 16: Even if this is "just get something deployed" attempt, I would think we don't want to deploy to production using the local development env variable file, but rather the actual environment variables. This is the reason I mentioned resorting to hacky pre-building in Discord.

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -24,8 +24,10 @@ export async function middleware(request: NextRequest) {
   const req = request;
   const response = NextResponse.next();
 
-  if (request.nextUrl.pathname === "/communities") {
-    return NextResponse.redirect(new URL("/", request.url));
+  if (req.nextUrl.href.includes("/cyberstorm/_next")) {
```
anttimaki, line 27: What is this needed for? Asking because it would be cleaner if we could just handle the routing shenanigans in the config files of next.js and nginx without resorting to touching the code.

packages/use-promise/package.json
```diff
@@ -13,6 +13,10 @@
     "build": "tsc",
     "dev": "tsc --watch"
   },
-  "dependencies": {},
-  "devDependencies": {}
+  "dependencies": {
```
anttimaki, line 16: I would prefer you'd approve[ my PR](https://github.com/thunderstore-io/thunderstore-ui/pull/997) instead, given that this is a separate issue, my PR has a bit more content, and I asked first :)

apps/cyberstorm-nextjs/middleware.ts
```diff
@@ -24,8 +24,12 @@ export async function middleware(request: NextRequest) {
   const req = request;
   const response = NextResponse.next();
 
+  // What we do here is the following:
```
anttimaki, line 27: Yes, I can see WHAT it does quite easily in this case. It would be much more valuable to know WHY we do it. Something like "The project root is set to `/cyberstorm` while the NextJs site is served from the same domain as the old Django site. Communities page acts as the front page until we have a proper landing page."

apps/cyberstorm-nextjs/next.config.js
```diff
@@ -2,6 +2,10 @@ const withPreconstruct = require("@preconstruct/next");
 
 /** @type {import('next').NextConfig} */
 const nextConfig = withPreconstruct({
+  basePath: "/cyberstorm",
+  images: {
+    path: "/cyberstorm/_next/image",
```
anttimaki, line 7: What does this actually do? I couldn't find `images.path` in [the documentation](https://nextjs.org/docs/app/api-reference/next-config-js/images), and the MM screenshot seems to load fine without it. Config files are probably the last place where we want to have leftover obsolete code, because no one dares to touch them unnecessarily. If this indeed is an undocumented feature that serves some purpose, maybe add a comment about it.

apps/cyberstorm-nextjs/package.json
```diff
@@ -23,7 +23,7 @@
     "eslint": "8.38.0",
     "eslint-config-next": "13.3.0",
     "jose": "^4.14.4",
-    "next": "^13.4.2",
+    "next": "^13.5.2",
```
anttimaki, line 26: Oo, the dev server output is way prettier with this version.

packages/cyberstorm/src/components/Footer/Footer.tsx
```diff
@@ -23,7 +16,7 @@ import { Tooltip } from "../Tooltip/Tooltip";
 import { Icon } from "../Icon/Icon";
 import { classnames } from "../../utils/utils";
 
-const AD_IMAGE_SRC = "/images/tsmm_screenshot.png";
+const AD_IMAGE_SRC = "/cyberstorm/images/tsmm_screenshot.png";
```
anttimaki, line 19: Not important here but if we use more static images before the old website is completely replaced, we might want to have the `/cyberstorm/images` path in a global static variable so we can edit the path in one place when the cyberstorm-prefix is dropped.

packages/cyberstorm/src/components/Header/Header.tsx
```diff
@@ -84,10 +67,20 @@ export function Header() {
                   <FontAwesomeIcon icon={faUpload} />
                 </Button.ButtonIcon>
               </Button.Root>
-            </PackageUploadLink>
+            </a>
           </li>
           <li>
-            <UserDropDown />
+            <Suspense
```
anttimaki, line 73: 👍🏻

packages/cyberstorm/src/components/Header/UserDropDown.tsx
```diff
@@ -17,27 +11,17 @@ import { AvatarButton } from "../Avatar/AvatarButton";
 import * as Button from "../Button/";
 import { DropDown, DropDownDivider, DropDownItem } from "../DropDown/DropDown";
 import { DropDownLink } from "../DropDown/DropDownLink";
-import { SettingsLink, TeamsLink, UserLink } from "../Links/Links";
 
-/**
```
anttimaki, line 22: Nice job removing the comments that would've been outdated.

packages/cyberstorm/src/components/Header/UserDropDown.tsx
```diff
@@ -17,27 +11,17 @@ import { AvatarButton } from "../Avatar/AvatarButton";
...
- * TODO: show fallback if user is not authenticated.
- * TODO: check if all the features in dropdown are actually supported
- *       on beta launch.
- */
 export const UserDropDown = () => {
   const dapper = useDapper();
   const user = usePromise(dapper.getCurrentUser, []);
   const avatar = user.connections.find((c) => c.avatar !== null)?.avatar;
 
-  if (!user.username) {
```
anttimaki, line 32: Irrelevant now, but as more stuff gets added the if-else approach below gets quite messy. Would probably be cleaner to have two subcomponents that both return a DropDown and then just to something like `return user.username ? <AuthenticatedDropDown user={user} /> : <UnauthenticatedUser />`

anttimaki (approved):
None of the comments are blockers. I'd say the ones regarding the config and the routing comment are the most important ones, but even they can be fixed in a separate PR.

The PR title, branch name and commit message are outdated/bad. FWIW I'm not fan of the graphite's auto-generated branch names since the date-prefix makes tab-completion a bit painful. Doesn't mean you can't use them, just letting you know my thoughts.

apps/cyberstorm-nextjs/middleware.ts
Oksamies: I'll literally take your word 👍

apps/cyberstorm-nextjs/next.config.js
Oksamies: Huh, yeah, I'm not sure what docs I've been looking at since I couldn't find it either now. Since it works without it just fine, I'll remove it 🤔

packages/cyberstorm/src/components/Footer/Footer.tsx
Oksamies: Hmm, I'm not sure right now how to do separation between component library and nextjs right now, so I'll add it to be figured out later. But yes this is a thing that needs to be sorted out.

packages/cyberstorm/src/components/Header/UserDropDown.tsx
Oksamies: Yea, I've tried to improve this component structure in a later PR, but this definitely needs more work in the future. Someone might call this code debt for when the other pages are added, but I call it iterative development 👓

Oksamies:
### Merge activity

* **Jan 17, 7:32 AM**: @Oksamies started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/998).

PR #1023, opened 2024-01-29
Remove /communities to /cyberstorm forwarding

Oksamies:

PR #1024, opened 2024-01-30
Move Communities layouts to be under the NextJS app

Oksamies:


Oksamies:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1024?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `master`
  * **PR #1023** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1024** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1024?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1024?utm_source=stack-comment).

apps/cyberstorm-nextjs/app/communities/CommunitiesLayout.module.css
anttimaki, line 1: I'm not sure the name of this file makes sense anymore since the counterpart is now called `layout.tsx`

apps/cyberstorm-nextjs/app/communities/CommunityList.tsx
anttimaki, line 1: Why is CommunityList moved from components to the app? What defines if something should be a component? I think this is an important question and whatever the decision ends up being, it should be documented/discussed in a manner that the whole team is aware of it.

apps/cyberstorm-nextjs/app/communities/CommunityList.tsx
```diff
@@ -5,9 +5,14 @@ import { useDapper } from "@thunderstore/dapper";
 import { usePromise } from "@thunderstore/use-promise";
 
 import styles from "./CommunityList.module.css";
-import { SortOptions } from "./CommunityListLayout";
-import { CommunityCard } from "../../CommunityCard/CommunityCard";
-import * as EmptyState from "../../EmptyState";
+import { CommunityCard } from "@thunderstore/cyberstorm/src/components/CommunityCard/CommunityCard";
```
anttimaki, line 8: Imports from other packages shouldn't contain a path like this.

apps/cyberstorm-nextjs/app/communities/CommunityList.tsx
```diff
@@ -5,9 +5,14 @@ import { useDapper } from "@thunderstore/dapper";
...
+import { EmptyState } from "@thunderstore/cyberstorm";
+
+enum SortOptions {
```
anttimaki, line 11: Why duplicate the options here?

apps/cyberstorm-nextjs/app/communities/CommunityListSkeleton.tsx
```diff
@@ -1,6 +1,6 @@
+import { range } from "@thunderstore/cyberstorm/src/utils/utils";
```
anttimaki, line 1: Import from other package with path information.

apps/cyberstorm-nextjs/app/communities/CommunityListSkeleton.tsx
```diff
@@ -1,6 +1,6 @@
...
 import styles from "./CommunityList.module.css";
-import { CommunityCardSkeleton } from "../../CommunityCard/CommunityCardSkeleton";
-import { range } from "../../../utils/utils";
+import { CommunityCardSkeleton } from "@thunderstore/cyberstorm/src/components/CommunityCard/CommunityCardSkeleton";
```
anttimaki, line 3: Import from other package with path information.

apps/cyberstorm-nextjs/app/communities/layout.tsx
```diff
@@ -0,0 +1,19 @@
+import "@thunderstore/cyberstorm-styles";
+import rootStyles from "../RootLayout.module.css";
+import { BreadCrumbs } from "@thunderstore/cyberstorm";
+import React from "react";
+import { PageHeader } from "@thunderstore/cyberstorm/src/components/Layout/BaseLayout/PageHeader/PageHeader";
```
anttimaki, line 5: Import from other package with path information.

apps/cyberstorm-nextjs/app/communities/layout.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export default function CommunitiesLayout(props: React.PropsWithChildren) {
```
anttimaki, line 7: I gave a quick glance to [the documentation](https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts#layout-pattern) and it seems to me the approach there is a bit different from the approach used here. How much though have you given to this, to ensure we have as DRY implementation as possible while still retaining necessary flexibility? (This is an unloaded question, honestly IDK how this should be done.)

apps/cyberstorm-storybook/.storybook/preview.js
```diff
@@ -1,11 +1,17 @@
...
+  CyberstormProviders,
+  Footer,
+  Header,
+} from "@thunderstore/cyberstorm";
 import "@thunderstore/cyberstorm-styles";
 import { DapperProvider } from "@thunderstore/dapper";
 import { DapperFake } from "@thunderstore/dapper-fake";
 import { LinkLibrary } from "../LinkLibrary";
 import { SessionProvider } from "../SessionContext";
+import styles from "../../cyberstorm-nextjs/app/RootLayout.module.css";
```
anttimaki, line 14: Relative imports between packages/apps should be no-go IMO.

apps/cyberstorm-storybook/.storybook/preview.js
```diff
@@ -30,7 +36,11 @@ export const decorators = [
           <SessionProvider>
             <DapperProvider dapperConstructor={() => new DapperFake()}>
               <LinkingProvider value={LinkLibrary}>
-                <Story />
+                <div className={styles.root}>
+                  <Header />
```
anttimaki, line 40: I don't quite see why we'd want to show header/footer on every showcase. Doesn't it just introduce noise making it hard to notice issues on the component itself? It also triggers changes for every Story everytime the header or footer is changed.

apps/cyberstorm-storybook/stories/pages/Communities.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import { StoryFn, Meta } from "@storybook/react";
+import CommunitiesLayout from "../../../cyberstorm-nextjs/app/communities/layout";
```
anttimaki, line 2: Relative imports between apps.

apps/cyberstorm-storybook/stories/pages/Communities.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Page from "../../../cyberstorm-nextjs/app/communities/page";
```
anttimaki, line 3: Relative imports between apps.

apps/cyberstorm-storybook/stories/pages/Communities.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+import React from "react";
+
+const meta = {
+  title: "Cyberstorm/Pages/Communities",
+  component: CommunitiesLayout,
+} as Meta<typeof CommunitiesLayout>;
+
+const Template: StoryFn<typeof CommunitiesLayout> = () => (
+  <div>
```
anttimaki, line 12: Is the `div` needed? Methinks `<SomethingLayout>` should work as expected without it.

anttimaki (commented):
Any idea why the test are not passing?

apps/cyberstorm-nextjs/app/communities/CommunitiesLayout.module.css
Oksamies: Trying to avoid the situation where we have 300 `layouts.module.css` files. Going with the old css file names for now.

apps/cyberstorm-nextjs/app/communities/CommunityList.tsx
Oksamies: That's a good question. IMO the line is and will always be too fuzzy, but in this case `CommunityList` should probably be a slot https://nextjs.org/docs/app/building-your-application/routing/parallel-routes#slots

I'm still looking into slots. In the meanwhile everything that is under the layout folder in components folder, will be moved under NextJS app. After that we'll split the layouts to more granular entities and move "components" back under cyberstorm components.

apps/cyberstorm-nextjs/app/communities/CommunityList.tsx
Oksamies: It wouldn't work with imports, because of NextJS. NextJS somehow can't handle imports between pages and layouts. Still thinking about a solution for this. Might just make a separate file for constants in the nextjs app

apps/cyberstorm-nextjs/app/communities/layout.tsx
Oksamies: To have as DRY implementation as possible, we'll have to change designs, which is not something I want to get into right now.

Currently I've just focused on getting everything moved under NextJS and after that figure out the best structure and separation.

So in practice the implementation/convertion will be something like this:
BaseLayout -> RootLayout
Cyberstorm Layouts -> Route specific layouts
Components/"Sub-cyberstorm-layouts" inside Cyberstorm layouts -> NextJS Slots or Pages. Depending on the case.

With this setup we should not have too much repetition and still have all the flexibility we need.

TLDR; Have thought about this. Don't have a clear answer. Best to go forward and figure it out as we go 👍

apps/cyberstorm-storybook/.storybook/preview.js
Oksamies: It is a no-go. I'm not sure what to do with storybook at this moment. I'll remove this, but storybook will be borked.

apps/cyberstorm-storybook/.storybook/preview.js
Oksamies: True, but we do want to mirror the NextJS layouts to storybook. It's a work in progress.

Oksamies:
Test failed because of request timeout. Dunno network lag?

PR #1082, opened 2024-04-09
Remove "Try again" button from error rendering

MythicManiac:
Remove the "Try again" button from the error rendering component as it does not seem to do anything in the place it currently turns up on.

In other words, either it's misleading & pointless or it's broken.


Oksamies:
The reason you've were experiencing "it not doing anything"; is because it tries to reload the slot, but fails every time, because of Dapper being broken on some cases. (not being init). I'll close this out, but if you have some way to reproduce the button "not doing anything" I'd be willing to take a look at it 👍

MythicManiac:
It didn't generate any console logs nor network traffic though, are you sure that's what's going on?

PR #1351, opened 2025-02-24
Improve ApiAction and fix hook mismatch in PackageSearch re-renders

Oksamies:
In the last tag the ability to like packages was added to PackageSearch.
This caused React issues because there were different amount of hooks
between re-renders. The solution here is to generate the hook ones and
then just change the input data to that hook.

This solution though warranted some changes to how ApiAction accepts
data. Corresponding changes are done to the PackageLikeAction.

On the same go this commit improves success and error message displaying
via toasts. The changes to useFormToaster are only done to
PackageLikeAction at this moment, but considering all the other actions
and forms are un-used right now, the same changes to other "action" and
"form" functions can be done, when they are taken into use.

Additionally UseApiAction now has return data types.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/PackageSearch/PackageSearch.tsx
+++ b/apps/cyberstorm-remix/app/commonComponents/PackageSearch/PackageSearch.tsx
@@ -407,6 +407,12 @@ export function PackageSearch(props: Props) {
     }
   }, [currentUser]);
 
+  const likeAction = PackageLikeAction({
+    isLoggedIn: Boolean(currentUser?.username),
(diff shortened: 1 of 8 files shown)


Oksamies:
* **#1351** <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1351?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/thunderstore-io/thunderstore-ui/1351?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #1535, opened 2025-09-16
Add forced currentUser update to settings and teams pages

Oksamies:
## Summary by CodeRabbit

- New Features
  - Settings is now available with Connections and Account views.
  - Teams section added, including team pages with tabs: Profile, Members, Service Accounts, and Settings.
  - Mobile user menu now includes the Settings option.

- Refactor
  - Unified internal navigation to use app-native links for Settings, Teams, and Upload across header, desktop dropdown, and mobile menus for more consistent routing.

### DIFF
--- a/apps/cyberstorm-remix/app/root.tsx
+++ b/apps/cyberstorm-remix/app/root.tsx
@@ -43,7 +43,7 @@ import {
   getSessionContext,
   getSessionStale,
   SESSION_STORAGE_KEY,
-  sessionValid,
+  runSessionValidationCheck,
 } from "@thunderstore/ts-api-react/src/SessionContext";
 import {
   getPublicEnvVariables,
@@ -152,12 +152,12 @@ export async function clientLoader({ request }: Route.ClientLoaderArgs) {
   } else {
     // In all other cases check if actually need to fetch
     // current-user data. Ideally we shouldn't need to do
-    // this sessionValid check again, but for some reason
+    // this runSessionValidationCheck check again, but for some reason
     // we need to run this here too in addition to the,
     // shouldRevalidate function, cause for some reason
     // the commits to localStorage are not done before
     // the clientLoader is run.
-    sessionTools.sessionValid(
+    sessionTools.runSessionValidationCheck(
       publicEnvVariables.VITE_API_URL,
       publicEnvVariables.VITE_COOKIE_DOMAIN
     );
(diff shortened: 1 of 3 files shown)


apps/cyberstorm-remix/app/root.tsx
```diff
@@ -142,13 +143,28 @@ export async function clientLoader() {
...
   );
 
-  // We need to run this here too in addition to the, shouldRevalidate function,
-  // as for some reason the commtis to localStorage are not done before the the clientLoader is run
-  sessionTools.sessionValid(
-    publicEnvVariables.VITE_API_URL,
-    publicEnvVariables.VITE_COOKIE_DOMAIN
+  let forceUpdateCurrentUser = false;
+  if (
+    request.url.startsWith(`${publicEnvVariables.VITE_BETA_SITE_URL}/teams`) ||
```
anttimaki, line 148: Why do we force this for these two pages? Or rather, why don't we force it on all pages? Why show any page to whoever is browsing as if they're a specific user, if we can't be sure they actually are that user? Or the other way around, why not let user know immediate when we detect that they're no longer authenticated?

apps/cyberstorm-remix/app/root.tsx
```diff
@@ -142,13 +143,28 @@ export async function clientLoader() {
...
+    forceUpdateCurrentUser = true;
+  } else {
+    // In all other cases check if actually need to fetch
+    // current-user data. Ideally we shouldn't need to do
+    // this sessionValid check again, but for some reason
+    // we need to run this here too in addition to the,
+    // shouldRevalidate function, cause for some reason
+    // the commits to localStorage are not done before
+    // the clientLoader is run.
+    sessionTools.sessionValid(
```
anttimaki, line 160: I feel this is bad pattern. Based on the naming I'd assume the function return a boolean. Well, it does, but the value is ignored. So why did we call it? Because the function had some side effects that are completely unknown to me until I read that function.

~~I think this should be split to two functions, isSessionValid that returns a boolean without having side effects, and another with a descriptive name for the side effects it has.~~ Scratch that, looks the boolean value is never used anywhere, so we can drop the return value and just have function that updates the staleness. (Assuming this is still needed after the earlier comment is resolved.)

anttimaki (changes requested):

Oksamies:
ESLint and CodeQL checks are failing because Graphite created a temporary branch when I was fixing and submitting code in each PR in order. (someone has limited the maximum amount of pushes to 2 in a single command and graphite normally tries to do the whole stack at once to prevent needing to create temporary branches)

PR #1538, opened 2025-09-17
Fix setSsion stale value set

Oksamies:
Herp derp, it's not a boolean so we should not check for a boolean, but
just set the value

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected session staleness handling to store the exact “yes”/“no” value, preventing sessions from being incorrectly marked as stale.
  * Users should see more consistent session behavior, avoiding unexpected stale-state prompts or refreshes.
  * No changes to the public API; no action required by users.


anttimaki (approved):
Herp derp, not too thrilled with the commit message on this one either.

PR #1547, opened 2025-09-17
Add Version Detail api endpoint to thunderstore-api

Oksamies:
Add Version Detail api endpoint to thunderstore-api

Add getPackageVersionDetails to Dapper

Add packageVersion pages


VilppeRiskidev (changes requested):
Please address coderabbi/graphite comments (either take the suggestions into account or reply to the comment with a reason why not, or just close/delete the conversations which are pure hallucinaition).

apps/cyberstorm-remix/app/p/tabs/Versions/PackageVersionVersions.tsx
```diff
@@ -0,0 +1,175 @@
...
+  if (isSemver(String(a[0].sortValue)) && isSemver(String(b[0].sortValue))) {
+    if (semverLt(String(a[0].sortValue), String(b[0].sortValue))) {
+      return columnMeta.direction;
+    }
+    if (semverGt(String(a[0].sortValue), String(b[0].sortValue))) {
+      return -columnMeta.direction;
+    }
+  }
+  return 0;
+}
```
VilppeRiskidev, line 87: Would this work? or do we not have semverCompare?
Also I think this function should probably be a more generic helper the comparison itself isn't really PackageVersion specific.

```
function rowSemverCompare(
  a: TableRow,
  b: TableRow,
  columnMeta: TableCompareColumnMeta
) {
  const av = String(a[0].sortValue);
  const bv = String(b[0].sortValue);

  if (isSemver(av) && isSemver(bv)) {
    return semverCompare(av, bv) * columnMeta.direction;
  }

  return 0;
}
```

Or even just call `semverCompare(av, bv) * columnMeta.direction;` direclty inline without the wrapper function if that works?

apps/cyberstorm-remix/app/p/tabs/Versions/PackageVersionWithoutCommunityVersions.tsx
```diff
@@ -0,0 +1,173 @@
...
+              />
+            </div>
+          </div>
+        )}
+      </Await>
+    </Suspense>
+  );
+}
+
+const columns: NewTableLabels = [
```
VilppeRiskidev, line 156: This object seems to be the same as in `PackageVersionVersion.tsx`. Could it be declared once somewhere?
How much differences are there between these two components, should one just extend the other to reduce duplicate code?

apps/cyberstorm-remix/app/p/packageVersion.tsx
```diff
@@ -0,0 +1,545 @@
...
+            >
+              {version.full_version_name}
+            </span>
+            <CopyButton text={version.full_version_name} />
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
```
VilppeRiskidev, line 545: This component and thepackageVersionWithoutCommunity should also be checked for unnecessary duplicate code. If it can't reasonably be reduced, please explain why.

PR #1585, opened 2025-10-16
Report Package implementation revision

Oksamies:
Bug fix: Add setting submission error to strongForm

strongForm wouldn't set it's submission error value, if the error is not directly related to strongForm internal logic

Bring ReportPackage Nimbus component closer to how other components are done

The existing implementation is very well done, but it's not aligned to how other similar components are done in the project. This commit introduces an implementation that is a bit simpler, with the added negative effect of having to have two usage instances of the component in the same page (packageListing route).

Performance effects are mostly unknown, but no noticable difference was discovered in local dev env testing.


apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -147,22 +168,91 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
           />
         </div>
-        {error && (
+        {strongForm.inputErrors ||
+        strongForm.refineError ||
+        strongForm.submitError ? (
           <div className="report-package__block">
-            <NewAlert csVariant="danger">{error}</NewAlert>
+            {strongForm.inputErrors ? (
+              <NewAlert csVariant="danger">
```
anttimaki: Input specific errors should be shown next to the relevant input if at all possible.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -147,22 +168,91 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
+                  <div key={key}>
+                    <strong>{key}:</strong>{" "}
+                    {Array.isArray(value) ? value.join(", ") : value}
+                  </div>
+                ))}
+              </NewAlert>
+            ) : null}
+            {strongForm.refineError ? (
+              <NewAlert csVariant="danger">
+                Error while refining: {strongForm.refineError.message}
```
anttimaki: IDK what refining means and I assume neither does the users.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -147,22 +168,91 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
+              </NewAlert>
+            ) : null}
+            {strongForm.submitError ? (
+              <NewAlert csVariant="danger">
+                Error while submitting: {strongForm.submitError.message}
```
anttimaki: This now contains the less user-friendly error message when attempting to report a package while unauthenticated.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -147,22 +168,91 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
           </div>
-        )}
+        ) : null}
       </Modal.Body>
       <Modal.Footer>
-        <NewButton csVariant="secondary" onClick={() => onOpenChange(false)}>
-          Cancel
-        </NewButton>
+        <Modal.Close asChild>
+          <NewButton csVariant="secondary">Cancel</NewButton>
```
anttimaki: Cancel no longer resets the form state. At least I'd expect it to do so.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -147,22 +168,91 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
         <NewButton csVariant="accent" onClick={strongForm.submit}>
           Send report
         </NewButton>
       </Modal.Footer>
     </>
   );
+
+  const done = (
+    <>
+      <Modal.Body className="report-package-body report-package-body--centered">
```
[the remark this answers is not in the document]
anttimaki: This(?) breaks the centering in the "submitted" modal in a very obvious way.

apps/cyberstorm-remix/app/p/packageListing.tsx
```diff
@@ -426,22 +420,34 @@ export default function PackageListing() {
...
+                          likeUpdateTrigger={fetchAndSetRatedPackages}
+                          requestConfig={config}
+                        />
+                      )}
+                    </Await>
+                  </Suspense>
+                  <Suspense fallback={<ReportPackageButton />}>
+                    <Await resolve={listing}>
+                      {(resolvedValue) => (
+                        <ReportPackage
```
anttimaki: After the changes in this PR, it would probably make sense to move this back into Actions. It was extracted from there just because my implementation wasn't very compatible with it.

anttimaki (commented):
Running out of time so I'm not sure if this is the full review. Anyway, I think we should discuss some of the changes proposed here on big picture level.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -27,29 +54,20 @@ const reportOptions: SelectOption<PackageListingReportRequestData["reason"]>[] =
...
-export function ReportPackageForm(props: ReportPackageFormFullProps) {
-  const {
-    config,
-    onOpenChange,
-    setIsSubmitted,
-    error,
-    setError,
-    ...requestParams
-  } = props;
+  const toast = useToast();
```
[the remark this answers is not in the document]
Oksamies: There has been no design decision made on the "way of showing errors" so in efforts of keeping the implementations similar, I've switched it to using the Toast in addition to having Alerts in the form.

> Using toasts blinking in the corner of the eye is a bad way to do it.

The Toasts are meant to convey that something went wrong with submissions or actions, not that the form has something wrong. Hence the utility for grabbing field errors etc in the strongForm. This form just can't have field errors in practice, so the only errors that should show are the submissions related errors.

Your thinking is not wrong, but you've confused two different things together

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
Oksamies: Not possible currently, as that would need mapping of fields to errors and anchoring to elements on the form. Additionally there has not been a design decision made on how errors should be displayed, for anything, so by that the errors should not be shown there.

apps/cyberstorm-remix/app/p/packageListing.tsx
```diff
@@ -426,22 +420,34 @@ export default function PackageListing() {
...
-                        likeUpdateTrigger={fetchAndSetRatedPackages}
-                        requestConfig={config}
-                      />
-                    )}
-                  </Await>
-                </Suspense>
-
-                {ReportPackageButton}
+                <div className="package-listing__narrow-other-actions">
+                  <Suspense fallback={<p>Loading...</p>}>
```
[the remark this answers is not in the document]
Oksamies: The mobile designs do not include any indication on how fallbacks or skeleton components should be done. So I've left this as so on purpose, so it can be noticed by the UX testers.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.css
```diff
@@ -42,4 +46,10 @@
     font-size: var(--font-size-body-md);
     line-height: var(--line-height-md);
   }
+
+  @media (width <= 650px) {
```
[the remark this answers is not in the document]
anttimaki: Regardless, how much time do you wager it would take to do it now? If it's not a big change, it could make sense to do it sooner than later to avoid having this conversation over and over again for each modal. (Note the earlier comment about using 100% width paired with maxWidth affects this one as well.)

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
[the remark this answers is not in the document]
anttimaki: > There has been no design decision made on the "way of showing errors" so in efforts of keeping the implementations similar, I've switched it to using the Toast in addition to having Alerts in the form.

Design decision has now been made. Generally speaking the defaults are:
- Errors related to modal should be shown on the modal, not on toast
- Errors related to a form should be shown on the form, not on toast
- Form submission error is an error related to the form
- What to do with "actions" (meaning e.g. like button) remains to be decided

Note that I'm not asking you to change everything everywhere all at once, just to not actively undo the changes already merged with the changes in this PR.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
```diff
@@ -95,21 +113,24 @@ export function ReportPackageForm(props: ReportPackageFormFullProps) {
...
         message = "You must be logged in to report a package.";
       }
-      setError(message);
+      toast.addToast({
+        csVariant: "danger",
+        children: message,
+        duration: 8000,
+      });
     },
   });
```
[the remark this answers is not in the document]
anttimaki: The toast will not be reintroduced here

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
anttimaki: > Refining the form data is done when the input data doesn't match the wanted data format of the endpoint. 

Thanks for the clarification, but telling this to *me* doesn't help our less tech-savvy users when they encounter the error message in production. What sort of error messages can we expect from `strongForm.refineError.message`? If they're human friendly we can probably drop the whole "Error while refining:" part? If not, would something like "Form contains invalid values:" work in combination the `strongForm.refineError.message`?

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
anttimaki: > This component is the wrong place to re-define the error message, it should be done in the function that throws the error or at very minimum some a bit more shared function like strongForm.

User doesn't care where the redefinition gets done as long as it gets done. Consequently I don't care where it gets done as long as it gets done, but now it seems it doesn't get done at all, so the change in this PR makes things worse.

apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackage.tsx
anttimaki: Currently the poll in our Discord has 100% (n=3) in favor of Cancel resetting the form contents.

A LLM also agrees, adding a note that if you're worried about clicking in the modal background accidentally resetting the form for the user, it's behaviour can be different. But that leaves using the X-button or esc-key as a grey area, where closing the modal is unlikely to be accidental. I'd say it's clearer to just reset the form.

apps/cyberstorm-remix/app/p/packageListing.tsx
anttimaki: What I tried to say was that [before my PRs](https://github.com/thunderstore-io/thunderstore-ui/pull/1582/files#diff-caa284e2f1e08c53540f6f4d8ab5a0ba0b5559beed976f69ba6a0d1a21b74705L1169), PackageReportButton, as well as `<div className="package-listing-sidebar__actions">` were contained in the Actions component, which made sense as they're somewhat a logical unit. I only extracted them because my implementation (custom hook) required it. If the custom hook is dropped, I'd say it makes sense to put them back in. Not a blocker but would make things cleaner IMO.

Oksamies:
Closing as too stale

PR #1605, opened 2025-11-07
Harden error handling across Thunderstore UI

Oksamies:
- document the end-to-end error-handling initiative, plans, and audit outcomes under docs
- add user-facing error primitives to @thunderstore/thunderstore-api, sanitize server detail, and cover the new behaviour with unit tests
- streamline @thunderstore/dapper-ts so its methods return raw Thunderstore API payloads without reshaping
- expose consistent UserFacingError handling through useApiCall, ApiAction, useFormToaster, and useStrongForm
- introduce the Nimbus error boundary toolkit plus shared loader helpers, then wire the root layout into the new fallbacks
- wrap community, package, settings, and upload routes with Nimbus boundaries, improve loader error mapping, and enhance Suspense fallbacks


docs/error-handling.md
```diff
@@ -0,0 +1,242 @@
...
+
+-   **`loader` (Server-side):** Errors are consistently processed through `handleLoaderError`. This utility catches errors from API calls (e.g., from Dapper), maps them to structured `UserFacingError` payloads, and throws a `Response`. This ensures that errors are standardized before they reach the UI, allowing for consistent presentation in the route's `ErrorBoundary`.
+
+-   **`clientLoader` (Client-side):** Error handling is less standardized. `clientLoader` functions often do not use `handleLoaderError`. Instead, they might:
+    -   Throw a generic `Error` object.
+    -   Have no explicit `try...catch` block, letting promise rejections from API calls be caught by the nearest Remix error boundary.
+
+This can lead to less specific error messages for the user. The recommendation is to apply the same `try...catch` and `handleLoaderError` pattern in `clientLoader` as is used in `loader` to ensure a consistent user experience.
+
+#### Bad Example (`clientLoader` without `handleLoaderError`)
```
anttimaki: I would probably drop the bad example and focus on good example(s).

docs/error-handling.md
```diff
@@ -0,0 +1,242 @@
...
+```typescript
+// This is not recommended
+export async function clientLoader({ params }: ClientLoaderFunctionArgs) {
+  const { dapper } = getLoaderTools();
+  try {
+    const team = await dapper.getTeam(params.teamId);
+    // Awaiting all follow-up data forces the UI to block on the clientLoader.
+    const members = await dapper.getTeamMembers(params.teamId);
+    return { team, members };
+  } catch (error) {
```
anttimaki: (Don't catch an error just to throw it again.)

docs/error-handling.md
```diff
@@ -0,0 +1,242 @@
...
+  try {
+    // Await the first fetch only because the follow-up request depends on it.
+    team = await dapper.getTeam(params.teamId);
+  } catch (error) {
+    // This provides a consistent, user-facing error.
+    handleLoaderError(error);
+  }
+
+  // Kick off dependent work without awaiting so Suspense can resolve it downstream.
+  const members = dapper.getTeamMembers(team.slug);
```
anttimaki: Should this be inside the try-block above? If it's intentionally left out, maybe use the comment above to describe what happens if it's erroneously placed inside the try-block.

docs/error-handling.md
```diff
@@ -0,0 +1,242 @@
...
+import { NimbusErrorBoundaryFallback } from "cyberstorm/utils/errors/NimbusErrorBoundary";
+
+export function ErrorBoundary() {
+  const error = useRouteError();
+  const payload = resolveRouteErrorPayload(error);
+
+  return (
+    <NimbusErrorBoundaryFallback
+      error={error}
+      title={payload.headline}
```
anttimaki: Could payload.headline be renamed to payload.title so it's in line with usage (and also shorter).

docs/error-handling.md
```diff
@@ -0,0 +1,242 @@
...
+  );
+}
+```
+
+### Usage with Suspense & `Await`
+
+When handing promises to the UI, wrap the consuming view with `Suspense`/`Await` and provide `NimbusAwaitErrorElement` as the `errorElement`. This ensures promise rejections render the same Nimbus fallback instead of crashing the tree.
+
+```tsx
+// apps/cyberstorm-remix/app/p/tabs/Readme/Readme.tsx
```
anttimaki: This is not what's done in the readme component (copy-paste comment?).

anttimaki (commented):
This review is only for the `error-handing.md` file. The main points are that I would attempt to improve the clarity by restructuring the document:

1. Reading the document with my level of knowledge of the project and the technologies used, I think structuring the doc based on the "core concepts" isn't the clearest way to do it. I would instead structure it to 1. Server side loader 2. Client side clientLoader 3. Error boundaries

2. For error boundaries, it's not perfectly clear to me what's the point of separate sections "## React Router Related Error Handling" and "## `NimbusErrorBoundary`"? Could we have just one section that tells us how to do things in Cyberstorm Remix app? I.e. combine the two parts to a structure something like 1. Root level error boundary ("be aware it exist, you don't need to worry about it") 2. Route-level Error boundary (including using suspense?) in your component 3. Granular Component Boundary in your component

Oksamies:
Closing as too stale

PR #1614, opened 2025-11-19
(thunderstore-api) Implement user-facing error handling improvements and update tests for authentication errors

Oksamies:
(thunderstore-api) Implement user-facing error handling improvements and update tests for authentication errors

(thunderstore-api) Refactor API fetch functions to remove async/await where unnecessary for improved performance

### DIFF
--- a/packages/thunderstore-api/src/apiFetch.ts
+++ b/packages/thunderstore-api/src/apiFetch.ts
@@ -44,29 +44,33 @@ function sleep(delay: number) {
   return new Promise((resolve) => setTimeout(resolve, delay));
 }
 
-export type apiFetchArgs<B, QP> = {
+type SchemaOrUndefined<Schema extends z.ZodSchema | undefined> =
+  Schema extends z.ZodSchema ? z.infer<Schema> : undefined;
+
+type apiFetchArgs<
+  RequestSchema extends z.ZodSchema | undefined,
+  QueryParamsSchema extends z.ZodSchema | undefined,
+  ResponseSchema extends z.ZodSchema | undefined,
+> = {
   config: () => RequestConfig;
   path: string;
-  queryParams?: QP;
+  queryParams?: SchemaOrUndefined<QueryParamsSchema>;
   request?: Omit<RequestInit, "headers" | "body"> & { body?: string };
   useSession?: boolean;
-  bodyRaw?: B;
+  bodyRaw?: SchemaOrUndefined<RequestSchema>;
+  requestSchema: RequestSchema;
+  queryParamsSchema: QueryParamsSchema;
+  responseSchema: ResponseSchema;
 };
 
-type schemaOrUndefined<A> = A extends z.ZodSchema
-  ? z.infer<A>
-  : never | undefined;
-
-export async function apiFetch(props: {
-  args: apiFetchArgs<
-    schemaOrUndefined<typeof props.requestSchema>,
-    schemaOrUndefined<typeof props.queryParamsSchema>
-  >;
-  requestSchema: z.ZodSchema | undefined;
-  queryParamsSchema: z.ZodSchema | undefined;
-  responseSchema: z.ZodSchema | undefined;
-}): Promise<schemaOrUndefined<typeof props.responseSchema>> {
-  const { args, requestSchema, queryParamsSchema, responseSchema } = props;
+export async function apiFetch<
+  RequestSchema extends z.ZodSchema | undefined,
+  QueryParamsSchema extends z.ZodSchema | undefined,
+  ResponseSchema extends z.ZodSchema | undefined,
+>(
+  args: apiFetchArgs<RequestSchema, QueryParamsSchema, ResponseSchema>
+): Promise<SchemaOrUndefined<ResponseSchema>> {
+  const { requestSchema, queryParamsSchema, responseSchema } = args;
 
   if (requestSchema && args.bodyRaw) {
     const parsedRequestBody = requestSchema.safeParse(args.bodyRaw);
@@ -74,6 +78,7 @@ export async function apiFetch(props: {
       throw new RequestBodyParseError(parsedRequestBody.error);
     }
   }
+
   if (queryParamsSchema && args.queryParams) {
     const parsedQueryParams = queryParamsSchema.safeParse(args.queryParams);
     if (!parsedQueryParams.success) {
@@ -88,8 +93,7 @@ export async function apiFetch(props: {
         apiHost: config().apiHost,
         sessionId: undefined,
       };
-  // TODO: Query params have stronger types, but they are not just shown here.
-  // Look into furthering the ensuring of passing proper query params.
+  const sessionWasUsed = Boolean(usedConfig.sessionId);
   const url = getUrl(usedConfig, path, queryParams);
 
   const response = await fetchRetry(url, {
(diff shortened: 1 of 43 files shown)


VilppeRiskidev (approved):

packages/thunderstore-api/src/get/__tests__/teamMembers.test.ts
```diff
@@ -10,5 +10,5 @@ it("ensures accessing team members requires authentication", async () => {
       data: {},
       queryParams: {},
     })
-  ).rejects.toThrowError("401: Unauthorized");
+  ).rejects.toThrowError("Authentication required. Please sign in.");
```
anttimaki: If the thrown error gives us clean access to the status code, it could be less brittle to check that instead of the message. Unless the message itself is the very behaviour we are interested in.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -39,32 +42,44 @@ async function fetchRetry(
...
 
+/**
+ * Simple timeout helper used by the retry loop.
+ */
 function sleep(delay: number) {
   return new Promise((resolve) => setTimeout(resolve, delay));
 }
 
-export type apiFetchArgs<B, QP> = {
+/**
```
anttimaki: This comment is actively confusing me. It seems it would be more suited for `apiFetchArgs` than `SchemaOrUndefined`. And if it would be a comment for `apiFetchArgs` I would nitpick that it provides no value.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -39,32 +42,44 @@ async function fetchRetry(
...
+ * Arguments supplied to `apiFetch` describing the HTTP request and validation schemas.
+ */
+type SchemaOrUndefined<Schema extends z.ZodSchema | undefined> =
+  Schema extends z.ZodSchema ? z.infer<Schema> : undefined;
+
+export type apiFetchArgs<
```
anttimaki: I've noticed in the few last PRs that especially types seems to be exported even when they're used only locally. I would rather not export stuff just in case. It's easy to add export later if needed, removing it is more effort if it's used somewhere outside already. And I think it also increases the odds of IDEs autoimporting stuff from these files directly, instead of the root of the package where the public interface of the package should be defined.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -39,32 +42,44 @@ async function fetchRetry(
...
+ */
+export async function apiFetch<
+  RequestSchema extends z.ZodSchema | undefined,
+  QueryParamsSchema extends z.ZodSchema | undefined,
+  ResponseSchema extends z.ZodSchema | undefined,
+>(props: {
+  args: apiFetchArgs<RequestSchema, QueryParamsSchema>;
+  requestSchema: RequestSchema;
+  queryParamsSchema: QueryParamsSchema;
+  responseSchema: ResponseSchema;
```
anttimaki: Why are these three defined separately? If we have function `apiFetch` and a dedicated type `apiFetchArgs`, I'd expect the latter to define the actual arguments the former accepts, not just some subset of them. Nothing in the function body seems to benefit from having them defined separately either (e.g. passing `args` ahead to a helper function). Including these three in `args` would also reduce boilerplate code where undefined values are passed to the function.

packages/thunderstore-api/src/delete/packageWiki.ts
```diff
@@ -6,7 +6,7 @@ import {
   PackageWikiPageDeleteRequestParams,
 } from "../schemas/requestSchemas";
 
-export async function deletePackageWikiPage(
+export function deletePackageWikiPage(
```
anttimaki: What is the reasoning for removing these async/awaits? Does it change the behaviour anywhere when the apiFetch itself still awaits the request? I'm asking since if the behaviour doesn't change, there's a chance this change was done under the assumption that it would change something, which in part could cause issues elsewhere.

If this is just "removing unnecessary stuff", I guess the changes themself are fine (although a bit pointless), but they definitely shouldn't be done as part of this huge PR stack. Here they just bloat the amount of review, rebase, etc. (Again, assuming I'm not missing something.)

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -81,14 +97,14 @@ export async function apiFetch(props: {
   }
 
   const { config, path, request, queryParams, useSession = false } = args;
+  const configSnapshot = config();
```
anttimaki: What's the idea behind adding this variable? Seems the config is used only on the next line so there doesn't seem to a possibility of this function using different values if the config is changed outside? And if the point is to protect from changes during the whole fetch process, shouldn't the snapshot be taken earlier than this point? Again, introducing an extra variable is just a nitpick thing, but it seems this might be done based on incorrect assumptions.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -100,19 +116,27 @@ export async function apiFetch(props: {
   });
 
   if (!response.ok) {
-    throw await ApiError.createFromResponse(response);
+    const apiError = await ApiError.createFromResponse(response, {
```
anttimaki: Nitpick: speaking of extra variables, I think this is one and it would be clearer to just throw on this line.

packages/thunderstore-api/src/apiFetch.ts
```diff
@@ -121,6 +145,9 @@ function getAuthHeaders(config: RequestConfig): RequestInit["headers"] {
     : {};
 }
 
+/**
+ * Builds the request URL with optional query parameters.
```
anttimaki: Nitpick: low value comment.

packages/thunderstore-api/src/errors.ts
```diff
@@ -7,37 +9,101 @@ type JSONValue =
   | { [x: string]: JSONValue }
   | JSONValue[];
 
+export type ApiErrorContext = {
```
anttimaki: (Another example of a type that is used only internally by the file being exported for some reason.)

packages/thunderstore-api/src/delete/packageWiki.ts
Oksamies: Behavior is not changed as it's still the promise that is returned. Fair point that these could have been in a separate PR, not included in the stack.

packages/thunderstore-api/src/apiFetch.ts
Oksamies: Good catch, that variable was actually a remnant from a thing I tried 👍

packages/thunderstore-api/src/errors.ts
```diff
@@ -7,37 +9,101 @@ type JSONValue =
...
+      : responseText
+        ? sanitizeServerDetail(responseText)
+        : undefined;
+
     return new ApiError({
-      message: `${response.status}: ${response.statusText}`,
+      message: buildApiErrorMessage({ response, responseJson, context }),
       response: response,
       responseJson: responseJson,
+      responseText,
```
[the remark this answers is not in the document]
Oksamies: Mm, I guess it could be handy if it's there, but I'm not sure to which extent? Definitely something that can be added if the need arises 👍

packages/thunderstore-api/src/errors.ts
```diff
@@ -61,22 +127,161 @@ export class ApiError extends Error {
...
+  for (const key of JSON_MESSAGE_PRIORITY_KEYS) {
+    if (key in objectValue) {
+      const message = extractMessage(objectValue[key]);
+      if (message) {
+        return sanitizeServerDetail(message);
+      }
+    }
   }
+
+  for (const entry of Object.values(objectValue)) {
```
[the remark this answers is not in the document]
Oksamies: The function has a very good chance of getting something out of the given input, but the where it's used, I'd say it's a very little risk of showing random garbage. Though en pistä päätä pantiks, it's been a long day so can't think straight anymore

Oksamies:
Closing as too stale

PR #1615, opened 2025-11-19
Refactor API methods to remove unnecessary async/await and improve return handling

Oksamies:
Refactor API methods to remove unnecessary async/await and improve return handling

Note: Some of the functions return shapes differ as of this commit, so the places that use these functions should change in future commits

Update dapper-fake functions to match the changed dapper methods return shapes (pagination)

### DIFF
--- a/packages/dapper-fake/src/fakers/community.ts
+++ b/packages/dapper-fake/src/fakers/community.ts
@@ -83,11 +83,28 @@ export const getFakeCommunities: GetCommunities = async (
   }
 
   const count = communities.length;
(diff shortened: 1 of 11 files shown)


VilppeRiskidev (approved):

anttimaki (commented):
Nothing to comment here except that the [comment from previous PR's review](https://github.com/thunderstore-io/thunderstore-ui/pull/1614#discussion_r2570750682) applies here too.

Oksamies:
Closing as too stale

PR #1752, opened 2026-03-15
Use ssrLoader HoF with all server side loader() funcs

anttimaki:
Technically this provides no benefits in loaders that don't use Dapper as that's the only source expected source for ApiErrors. It was decided to use it in all loaders for consistency's sake, and to avoid situations where it ends up missing when a developer adds Dapper calls to an existing loader.


VilppeRiskidev (approved):

anttimaki:
At this point it's easier to redo than rebase this.

PR #1777, opened 2026-03-24
Re-rendering performance improvements

Oksamies:


Oksamies:
Closing as it's better to start over after educating myself

PR #1789, opened 2026-03-26
Update Node.js engine version to >=24.0.0 across all packages

Oksamies:


Roffenlund (approved):

anttimaki (changes requested):
Was there any specific reason not version pinning the exact version, or at least the major version, as suggested by the task description and how this was attempted at #1559 ? If we want to move towards exact version pinning (in this world of supply chain attacks and whatnot), this might be a good starting point?

Also not all docker files in the repo updated to use the same Node version, and I would pin the alpine version as well. Unless there's a reason not to?

Oksamies:
> Was there any specific reason not version pinning the exact version, or at least the major version, as suggested by the task description and how this was attempted at #1559 ? If we want to move towards exact version pinning (in this world of supply chain attacks and whatnot), this might be a good starting point?
> 
> Also not all docker files in the repo updated to use the same Node version, and I would pin the alpine version as well. Unless there's a reason not to?

Good questions/point-outs, I had forgot that you already worked on this. I'll rebase and continue on your PR as that has more cohesive changes than this.

PR #1810, opened 2026-05-06
Improve form field validation checks(TS-3357)

Roffenlund:
* Update validators with new validators
* Refactor validators and add function for getting field level error messages
* Simplify useStrongForm
* Update tests for validators and useStrongForm
* Add field-level validation to forms

The following forms were updated:
* Create Team
* Add Service Account
* Donation Link (Team Profile)
* Add Team Member
* Wiki Create & Edit
* Report Package

### DIFF
--- a/apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackageForm.tsx
+++ b/apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackageForm.tsx
@@ -98,7 +98,7 @@ export function ReportPackageForm(
   >({
     inputs: formInputs,
     validators: {
-      description: { maxLength: 12288 },
+      description: {},
     },
     submitor,
     onSubmitSuccess: () => {
--- a/apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackageForm.tsx
+++ b/apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackageForm.tsx
@@ -97,6 +97,9 @@ export function ReportPackageForm(
     InputErrors
   >({
     inputs: formInputs,
+    validators: {
+      description: { maxLength: 12288 },
+    },
     submitor,
     onSubmitSuccess: () => {
       setIsSubmitted(true);
@@ -116,6 +119,7 @@ export function ReportPackageForm(
   const handleCancel = () => {
     resetFormInputs();
     setError(null);
+    strongForm.resetFormState();
     onOpenChange(false);
   };
 
@@ -154,7 +158,13 @@ export function ReportPackageForm(
             placeholder="Invalid submission"
             csSize="textarea"
             rootClasses="report-package__textarea"
+            {...strongForm.getFieldComponentProps("description")}
           />
+          {strongForm.getFieldError("description") && (
+            <NewAlert csVariant="danger" csSize="small">
+              {strongForm.getFieldError("description")}
+            </NewAlert>
+          )}
         </div>
         {error && (
           <div className="report-package__block">
--- a/apps/cyberstorm-remix/app/p/tabs/Wiki/WikiNewPage.tsx
+++ b/apps/cyberstorm-remix/app/p/tabs/Wiki/WikiNewPage.tsx
@@ -14,6 +14,7 @@ import { type OutletContextShape } from "~/root";
 
 import {
   Heading,
+  NewAlert,
   NewButton,
   NewTextInput,
   SkeletonBox,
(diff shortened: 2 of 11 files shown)


apps/cyberstorm-remix/app/p/components/ReportPackage/ReportPackageForm.tsx
```diff
@@ -97,6 +97,9 @@ export function ReportPackageForm(
     InputErrors
   >({
     inputs: formInputs,
+    validators: {
+      description: { maxLength: 12288 },
```
Oksamies: This does not trigger on my local dev, when I test it with 12289 characters (or even more). As per task description the field should go red after 12288 characters, right?

Oksamies: Never mind, my dev env had not updated to this branch. The form submit validation works, but the the length is not detected onBlur if form has not been submitted.
e.g.
Write long text -> No error
Submit too long text -> Error
Submit long text -> Error -> Remove characters -> Error goes away -> Add more characters so that limit is passed -> Error re-appears. No submission was needed for the error to reappear.

Ideally the limit would be checked on keystroke/timer/onBlur, even when the form is not submitted.

apps/cyberstorm-remix/app/p/tabs/Wiki/WikiNewPage.tsx
```diff
@@ -179,6 +180,10 @@ export default function Wiki() {
     InputErrors
   >({
     inputs: formInputs,
+    validators: {
+      title: { required: true, maxLength: 512 },
```
Oksamies, line 184: onBlur doesn't work. When form is submitted this is checked and prevents submission if too long.

apps/cyberstorm-remix/app/p/tabs/Wiki/WikiNewPage.tsx
```diff
@@ -179,6 +180,10 @@ export default function Wiki() {
...
+      markdown_content: { required: true, maxLength: 100000 },
```
Oksamies: Required check works, length check doesn't work in anyway. Also It looks like the Django field itself doesn't the 10k character limit. Should it?

apps/cyberstorm-remix/app/p/tabs/Wiki/WikiPageEdit.tsx
```diff
@@ -248,6 +249,10 @@ export default function WikiEdit() {
     InputErrors
   >({
     inputs: formInputs,
+    validators: {
+      title: { required: true, maxLength: 512 },
+      markdown_content: { required: true, maxLength: 100000 },
```
Oksamies: Same here as with the creation of new wiki.

apps/cyberstorm-remix/app/p/tabs/Wiki/WikiNewPage.tsx
Oksamies: Also to clarify the onBlur does work when I remove characters to be under the 512 limit, so it's only not working when going over.

apps/cyberstorm-remix/app/settings/teams/team/tabs/ServiceAccounts/ServiceAccounts.tsx
```diff
@@ -139,7 +139,16 @@ function AddServiceAccountForm(props: {
     InputErrors
   >({
     inputs: formInputs,
-    validators: { nickname: { required: true } },
+    validators: {
+      nickname: {
+        required: true,
+        maxLength: 32,
+        pattern: {
+          regex: /^[A-Za-z0-9_]+$/,
```
Oksamies, line 147: This is not checked onBlur before submission.

apps/cyberstorm-remix/app/settings/teams/Teams.tsx
```diff
@@ -203,11 +203,22 @@ function CreateTeamForm(props: { config: () => RequestConfig }) {
     InputErrors
   >({
     inputs: formInputs,
-    validators: { name: { required: true } },
+    validators: {
+      name: {
+        required: true,
+        maxLength: 64,
+        pattern: {
+          regex: /^[a-zA-Z0-9]+([a-zA-Z0-9_]+[a-zA-Z0-9])?$/,
```
Oksamies, line 211: Same here not checked onBlur before submission is made.

Oksamies (approved):
I'm a bit unsure what the intention was with the onBlur behavior, so I'm going to accept this on the basis of; I think it works fine and requirements written in task might be invalid.

Only valid request change is the wiki markdown content limit not working, but that also has it's caviat.

Roffenlund:
@Oksamies the maxLength checks do work as intended. I generated strings manually and the errors fired as intended when submitting the forms. I did however remove redundant maxLength validators for fields that use TextField in the Django backend.

Oksamies (approved):

PR #1811, opened 2026-05-07
Misc fixes

Oksamies:


apps/cyberstorm-remix/app/communities/communities.tsx
```diff
@@ -235,6 +235,9 @@ const CommunitiesList = memo(function CommunitiesList(props: {
               new Date(community.datetime_created).getTime() >
               new Date().getTime() - 1000 * 60 * 60 * 336
             }
+            imageLoading={index < 28 ? "eager" : "lazy"}
+            imagePreload={index < 11}
+            imageFetchPriority={index < 11 ? "high" : "auto"}
```
Roffenlund: These numbers should probably be explicitly declared somewhere?

apps/cyberstorm-remix/app/c/Community.tsx
```diff
@@ -168,52 +168,50 @@ export default function Community() {
...
-                        </NewLink>
-                      ) : null
-                    }
-                  </Await>
-                </Suspense>
-                <Suspense fallback={<SkeletonBox />}>
-                  <Await resolve={community}>
-                    {(resolvedValue) =>
+
+              <Suspense fallback={<SkeletonBox />}>
```
Roffenlund, line 197: This `<Suspense>` block would probably benefit from being broken out into a separate helper function as it seems repetitive.

Roffenlund (changes requested):

apps/cyberstorm-remix/app/c/Community.tsx
Oksamies: Naah, I'd say it's fine this way. If it was a more complicated set of if statements then perhaps

Roffenlund (approved):

PR #1817, opened 2026-05-15
Error pages (TS-3396)

Roffenlund:
* Export ErrorBoundary on layout/parent routes
* Export ErrorBoundary on standalone routes
* Use `isApiError `instead of `instanceof ApiError`
* Use `ssrLoader `for component SSR `loader()` functions


Oksamies (approved):
LGTM :shipit:

PR #1819, opened 2026-05-18
Move package management into a listing-page modal

VilppeRiskidev:
* Replace the standalone /edit route with a Manage Package modal on the package listing page, matching the Review Package pattern.
* Includes permission-gated sections (categories, deprecate/undeprecate, unlist with confirmation)
* Removes the old edit URL and PackageEdit links
* Extracts PackageListingManagement for loader/Await wiring

* Restore Manage Package for moderators (`can_manage || can_moderate`); keep per-action permission gates inside the modal.
* Load community filters only when the modal opens and the user can manage categories.
* Fetch listing status only after permissions, and only for users who need moderation/manage/admin data.
* Split Suspense so review notes and Manage Package are not blocked by unrelated requests.
* Lazy-mount modal content and add in-flight guards for unlist, deprecate, and category save.


Roffenlund (dismissed):
Is this still WIP? The modal seems to be missing features.

Oksamies (approved):
:shipit: (after checking out Copilot comments. I ran it after reviewing, just in case)

PR #1842, opened 2026-05-26
Use text-underline-position:from-font in Links

VilppeRiskidev:


packages/cyberstorm/src/newComponents/Link/Link.css
```diff
@@ -2,6 +2,7 @@
   .link {
     color: var(--link-color);
     text-decoration: none;
+    text-underline-position: from-font;
```
Roffenlund: I don't understand the effect of this line. Isn't it effectively dead code since text-decoration is already set to none in the line above or am I misunderstanding the use case of this?

Roffenlund (changes requested):

Roffenlund (approved):

PR #1846, opened 2026-05-27
Upload page fixes

VilppeRiskidev:


VilppeRiskidev:
Unit tests might be a bit bad, I can improve on those

apps/cyberstorm-remix/app/styles/layout.css
```diff
@@ -25,7 +25,7 @@
     align-self: flex-start;
     justify-content: flex-start;
     max-width: var(--nimbus-layout-content-max-width, 120rem);
-    padding: var(--space-16);
+    padding: var(--space-16) var(--space-16) 4rem;
```
Oksamies, line 29: This change should have been in it's own PR

apps/cyberstorm-remix/app/upload/uploadHooks.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  useEffect(() => {
+    if (file) {
+      void startUpload();
+    }
+  }, [file, startUpload]);
+
+  const clearFile = useCallback(() => {
+    setFile(null);
+    if (fileInputRef.current) {
```
Oksamies, line 68: I'm not 100% sure, but maybe the fileInputRef should be added to the inputs list of the useCallback?

apps/cyberstorm-remix/vitest.config.ts
```diff
@@ -1,11 +1,20 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { defineProject } from "vitest/config";
 
-const cyberstormRoot = new URL("./cyberstorm", import.meta.url).pathname;
+const projectDir = path.dirname(fileURLToPath(import.meta.url));
+const cyberstormRoot = path.join(projectDir, "cyberstorm");
+// package.json exports point at dist/, which CI does not build before `yarn coverage`.
```
Oksamies: Wouldn't fixing the package.json exports be the right course of action?

Oksamies (approved):
Good refactor and good use of StrongForm 👍

I would have preferred the changes to other routes than upload.tsx be in a separate follow PR, but no need to do that now, just for future reference.

I'd recommend re-doing the commits in a more sensible way. Instead of 15 commits I think there should only be few.

apps/cyberstorm-remix/app/styles/layout.css
VilppeRiskidev: Good point. This'll be made obsolete soon with all of the whitespace fixes.

apps/cyberstorm-remix/app/commonComponents/FormSection/FormSection.css
```diff
@@ -0,0 +1,132 @@
...
+    .form-sections,
+    .settings-items {
+      --form-section-meta-width: 13rem;
+    }
+  }
+
+  @media (width <= 48rem) {
+    .form-sections {
+      gap: var(--gap-3xl);
+    }
```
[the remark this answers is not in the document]
VilppeRiskidev: This'll be made obsolete with the upcoming witespace fixes.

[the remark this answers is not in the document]
VilppeRiskidev: This'll be made obsolete with the upcoming whitespace fixes.

apps/cyberstorm-remix/app/upload/uploadHooks.ts
VilppeRiskidev: You’re right to check it, but in this case fileInputRef should not be added.

* fileInputRef is a useRef(...) object, and the ref object identity is stable across renders.
* useCallback deps are about values whose identity can change and should trigger a new callback.
* Here, clearFile reads fileInputRef.current at call time, so it always sees the latest node without needing a dep.

PR #1852, opened 2026-06-03
Nimbus SSR loader caching (TS-3735)

Roffenlund:
Implement cache control headers in selected Nimbus views. Implement a util for creating cache-control headers and integrate this into the ssrLoader which wraps the SSR loader() functions. The default behavior ignores caching, so caching has to be switched on if caching is rquired in selected route.

* Add util for building Cache-Control headers
* Integrate Cache-Control util in ssrLoader to enable cache
* Add cache to selected views

### DIFF
--- a/apps/cyberstorm-remix/app/communities/communities.tsx
+++ b/apps/cyberstorm-remix/app/communities/communities.tsx
@@ -60,46 +60,53 @@ const selectOptions = [
   },
 ];
 
-export const loader = ssrLoader(async ({ request }: Route.LoaderArgs) => {
-  const url = new URL(request.url);
-  const searchParams = url.searchParams;
-  const order = searchParams.get("order") ?? SortOptions.Popular;
-  const search = searchParams.get("search");
-  const page = undefined;
-  const dapper = new DapperTs(() => {
+export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
+  return loaderHeaders;
+};
+
+export const loader = ssrLoader(
+  async ({ request }: Route.LoaderArgs) => {
+    const url = new URL(request.url);
+    const searchParams = url.searchParams;
+    const order = searchParams.get("order") ?? SortOptions.Popular;
+    const search = searchParams.get("search");
+    const page = undefined;
+    const dapper = new DapperTs(() => {
+      return {
+        apiHost: getApiHostForSsr(),
+        sessionId: undefined,
+      };
+    });
     return {
-      apiHost: getApiHostForSsr(),
-      sessionId: undefined,
+      communities: await dapper.getCommunities(
+        page,
+        order === null ? undefined : order,
+        search === null ? undefined : search
+      ),
+      seo: createSeo({
+        descriptors: [
+          { title: "Communities | Thunderstore" },
+          {
+            name: "description",
+            content: "Browse all communities on Thunderstore",
+          },
+          { property: "og:type", content: "website" },
+          {
+            property: "og:url",
+            content: `${url.origin}/communities`,
+          },
+          { property: "og:title", content: "Communities | Thunderstore" },
+          {
+            property: "og:description",
+            content: "Browse all communities on Thunderstore",
+          },
+          { property: "og:site_name", content: "Thunderstore" },
+        ],
+      }),
     };
-  });
-  return {
-    communities: await dapper.getCommunities(
-      page,
-      order === null ? undefined : order,
-      search === null ? undefined : search
-    ),
-    seo: createSeo({
-      descriptors: [
-        { title: "Communities | Thunderstore" },
-        {
-          name: "description",
-          content: "Browse all communities on Thunderstore",
-        },
-        { property: "og:type", content: "website" },
-        {
-          property: "og:url",
-          content: `${url.origin}/communities`,
-        },
-        { property: "og:title", content: "Communities | Thunderstore" },
-        {
-          property: "og:description",
-          content: "Browse all communities on Thunderstore",
-        },
-        { property: "og:site_name", content: "Thunderstore" },
-      ],
-    }),
-  };
-});
+  },
+  { cache: true }
+);
 
 export async function clientLoader({ request }: Route.ClientLoaderArgs) {
   const tools = getSessionTools();
--- a/apps/cyberstorm-remix/app/communities/communities.tsx
+++ b/apps/cyberstorm-remix/app/communities/communities.tsx
@@ -60,9 +60,7 @@ const selectOptions = [
   },
 ];
 
-export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
-  return loaderHeaders;
-};
+export { forwardLoaderHeaders as headers } from "cyberstorm/utils/ssrLoader";
 
 export const loader = ssrLoader(
   async ({ request }: Route.LoaderArgs) => {
--- a/apps/cyberstorm-remix/app/p/packageListing.tsx
+++ b/apps/cyberstorm-remix/app/p/packageListing.tsx
@@ -155,8 +155,13 @@ export const loader = ssrLoader(
         ],
       }),
     };
-  }
+  },
+  { cache: true }
 );
+
+export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
+  return loaderHeaders;
+};
 
 export async function clientLoader({
   params,
--- a/apps/cyberstorm-remix/app/p/packageListing.tsx
+++ b/apps/cyberstorm-remix/app/p/packageListing.tsx
@@ -159,9 +159,7 @@ export const loader = ssrLoader(
   { cache: true }
 );
 
-export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
-  return loaderHeaders;
-};
+export { forwardLoaderHeaders as headers } from "cyberstorm/utils/ssrLoader";
 
 export async function clientLoader({
   params,
--- a/apps/cyberstorm-remix/app/p/tabs/Wiki/Wiki.tsx
+++ b/apps/cyberstorm-remix/app/p/tabs/Wiki/Wiki.tsx
@@ -31,50 +31,60 @@ import "./Wiki.css";
 
 export { RouteErrorBoundary as ErrorBoundary } from "app/commonComponents/ErrorBoundary";
 
-export const loader = ssrLoader(async ({ params }: Route.LoaderArgs) => {
-  if (params.communityId && params.namespaceId && params.packageId) {
-    const dapper = new DapperTs(() => {
-      return {
-        apiHost: getApiHostForSsr(),
-        sessionId: undefined,
-      };
-    });
+export const loader = ssrLoader(
+  async ({ params }: Route.LoaderArgs) => {
+    if (params.communityId && params.namespaceId && params.packageId) {
+      const dapper = new DapperTs(() => {
+        return {
+          apiHost: getApiHostForSsr(),
+          sessionId: undefined,
+        };
+      });
 
-    let wiki: Awaited<ReturnType<typeof getPackageWiki>> | undefined;
+      let wiki: Awaited<ReturnType<typeof getPackageWiki>> | undefined;
 
-    try {
-      wiki = await dapper.getPackageWiki(params.namespaceId, params.packageId);
-    } catch (error) {
-      if (isApiError(error) && error.response.status === 404) {
-        wiki = undefined;
-      } else {
-        throw error;
+      try {
+        wiki = await dapper.getPackageWiki(
+          params.namespaceId,
+          params.packageId
+        );
+      } catch (error) {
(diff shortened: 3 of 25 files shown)


apps/cyberstorm-remix/cyberstorm/utils/ssrLoader.ts
```diff
@@ -1,35 +1,70 @@
...
+      const result = await fn(args);
+
+      if (cache === false) {
+        return result;
+      }
+
+      const cacheOptions = cache === true ? {} : cache;
+      return data(result, {
+        headers: cacheControl(cacheOptions),
+      }) as unknown as T;
```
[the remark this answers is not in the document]
Oksamies: AIs take on this AI review comment:

The review comment is not valid for this codebase. Here's why:

The mechanism is accurately described — cache: true → cacheControl({}) → Cache-Control: public, s-maxage=300, ..., and cacheControl never adds Vary. All of that is correct.

But the premise doesn't apply here. Every SSR loader that uses caching (all 14 routes) hardcodes sessionId: undefined in the DapperTs constructor — they explicitly never send auth credentials. Session/auth-specific data is fetched exclusively in clientLoader, which runs browser-side only and whose response is never CDN-cached.

The architecture enforces the separation at the data-fetching level rather than at the cache-header level. There is no auth-varying data in the cached SSR responses to begin with, so Vary: Cookie or isPrivate would add no safety — they'd just unnecessarily bypass the CDN for responses that are genuinely identical across all users.

Verdict: false positive. The review comment would be valid if SSR loaders used session-based auth, but this codebase intentionally doesn't.

Oksamies (approved):
A soft approval that assumes all the AI review comments will be addressed / fixes will be implemented.

apps/cyberstorm-remix/cyberstorm/utils/ssrLoader.ts
```diff
@@ -42,14 +74,12 @@ export function ssrLoader<A extends LoaderFunctionArgs, T>(
         // Include only handpicked attributes to avoid sensitive information
         // ending up e.g. on logs.
         const payload = { status, statusText, url };
-
         throw new Response(JSON.stringify(payload), {
```
Oksamies, line 162: **5xx error documents can end up publicly CDN-cached.**

The Response thrown here carries no `Cache-Control`, but on a document request the thrown error doesn't control the final headers. The package tab routes (Readme/Changelog/Versions/Required, and the packageListingVersion children) export no `ErrorBoundary`, so a thrown 5xx bubbles to `packageListing.tsx` / `packageListingVersion.tsx` / the wiki layout — whose own loaders succeeded. React Router resolves document headers from the matches up to the boundary route (`getDocumentHeadersImpl`), calling the boundary's `headers` function with its own *successful* `loaderHeaders`; `errorHeaders` is available as an argument but `({ loaderHeaders }) => loaderHeaders` ignores it.

Net effect: a transient backend 5xx on e.g. `/c/x/p/ns/pkg/changelog` renders the boundary error page with `public, max-age=60, s-maxage=300, stale-while-revalidate=600` — caches are explicitly permitted to pin that error document for up to 5 minutes (whether a CDN actually stores 5xx is config-dependent, but the header invites it).

Suggestion: add `"Cache-Control": "no-store"` to the headers of the Response thrown here (at least for `status >= 500`), and/or make the shared headers function prefer `errorHeaders` / return `no-store` when an error is present.

apps/cyberstorm-remix/app/p/tabs/Wiki/Wiki.tsx
```diff
@@ -31,50 +31,60 @@ import "./Wiki.css";
...
+              content: `Wiki for ${params.namespaceId}-${params.packageId}`,
+            },
+          ],
+        }),
+      };
+    } else {
+      throw new Error("Namespace ID or Package ID is missing");
     }
+  },
+  { cache: true }
```
Oksamies, line 82: **The wiki create/edit exclusion doesn't hold for document requests.**

`WikiNewPage.tsx` and `WikiPageEdit.tsx` are deliberately left uncached, but they export no `headers` function — and in RR v7 a matched route without one inherits the accumulated parent headers (`getDocumentHeadersImpl`: `headersFn == null → new Headers(parentHeaders)`). Since this layout (and `packageListing`/`Community` above it) is cached, document requests to `/wiki/new` and `/wiki/:slug/edit` are served with the full `public, max-age=60, s-maxage=300, stale-while-revalidate=600` — including WikiPageEdit's SSR-rendered page content in the edit form. Not a data leak (the SSR output is anonymous), but it contradicts the intended exclusion and makes authoring pages CDN-cacheable for ~5 min.

Note that dropping `{ cache: true }` from this layout alone wouldn't fix it — `packageListing.tsx` and `Community.tsx` would still bubble their public header down. The robust fix is an explicit override on the editor leaves, e.g. in `WikiNewPage.tsx` and `WikiPageEdit.tsx`:

```ts
export const headers: Route.HeadersFunction = () => ({ "Cache-Control": "no-store" });
```

apps/cyberstorm-remix/app/p/packageListing.tsx
```diff
@@ -130,9 +130,14 @@ export const loader = ssrLoader(
         ],
       }),
     };
-  }
+  },
+  { cache: true }
 );
 
+export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
```
Oksamies: Same header-inheritance mechanism as on the wiki layout: child routes that didn't opt in but also don't export `headers` inherit this public `Cache-Control` on their document responses — `Source.tsx` here, and `PackageSearch.tsx` (the community landing index) under the cached `Community.tsx`. Probably harmless for Source (the SSR document is a shell), but the community landing page gets ~5 min of public caching it never opted into.

Worth deciding per leaf: opt them in explicitly, or give them an overriding `headers` export. It would also be worth a line in the ssrLoader docstring noting that a cached parent will cache an uncached child's document unless the child exports its own `headers` — the current docstring only covers routes *with* caching enabled.

apps/cyberstorm-remix/cyberstorm/utils/ssrLoader.ts
```diff
@@ -1,27 +1,59 @@
...
     try {
-      return await fn(args);
+      const result = await fn(args);
+
+      if (cache === false || result instanceof Response) {
+        return result;
+      }
+
+      const cacheOptions = cache === true ? {} : cache;
+      return data(result, {
```
Oksamies, line 139: Two notes on the scope of these headers:

1. They apply not only to document HTML but also to the single-fetch `.data` responses (RR calls `getDocumentHeaders` inside `singleFetchLoaders` as well) — and those are requests browsers send *with* the session cookie.
2. The emitted header is `public` with no `Vary`.

Both are safe today only because of the invariant that cache-enabled loaders are fully anonymous (`sessionId: undefined`, no cookie/header reads, per-user data deferred to `clientLoader`). That invariant is load-bearing but currently undocumented — if a future change makes any cached loader render session-dependent output, the first response would be cross-served to all users from the CDN. Suggest stating it explicitly in this docstring ("loaders wrapped with `cache` MUST remain fully anonymous"); adding `Vary: Cookie` to the public branch in `cacheControl()` would be cheap defense-in-depth.

apps/cyberstorm-remix/cyberstorm/utils/ssrLoader.ts
```diff
@@ -1,27 +1,59 @@
...
- * 5xx ApiErrors are reported to Sentry before conversion because
- * sentryHandleError (handleError) filters out Response throws, which means
- * server-side API failures would otherwise be invisible in Sentry.
- * 4xx errors are intentional user-facing responses and are not captured.
+ * Caching is disabled by default. Pass `cache: true` for standard CDN settings,
+ * or provide a CacheControlOptions object to customize durations.
+ * Routes with caching enabled must also export a `headers` function to forward
+ * loader headers on document requests:
+ *
+ *   export const headers: Route.HeadersFunction = ({ loaderHeaders }) => loaderHeaders;
```
Oksamies: Worth noting the failure mode of this contract: `cache: true` without the matching `headers` export compiles, type-checks, and runs cleanly — the header is just silently dropped from the document response, so caching no-ops with no signal anywhere. It fails open in the safe direction, and all 14 routes in this PR pair the two correctly, but nothing enforces the pairing for the next route.

One way to shrink the footgun and the copy-paste: export a ready-made helper next to ssrLoader, e.g. `export const forwardLoaderHeaders = ({ loaderHeaders }: HeadersArgs) => loaderHeaders;` so routes write `export const headers = forwardLoaderHeaders;`. That would also give a single place to later add error-awareness (see the comment on the thrown-Response path).

apps/cyberstorm-remix/cyberstorm/utils/cache.ts
```diff
@@ -0,0 +1,39 @@
...
+  "Cache-Control": string;
+} {
+  const {
+    browserMaxAge = 60,
+    cdnMaxAge = 300,
+    staleWhileRevalidate = 600,
+    isPrivate = false,
+  } = options ?? {};
+
+  if (isPrivate) {
```
Oksamies, line 48: A few small notes on the util:

- The private branch silently ignores `cdnMaxAge` and `staleWhileRevalidate`. Dropping `s-maxage` is correct by design (it's a shared-cache directive), but the JSDoc lists all options generically — worth a note that those two only apply to the public branch (per RFC 5861, `stale-while-revalidate` would also be valid for private caches if you ever want it there).
- The JSDoc describes `staleWhileRevalidate` as CDN-scoped, but on a `public` response it applies to browser caches too — the effective browser staleness ceiling is `browserMaxAge + staleWhileRevalidate` (660s with defaults), not 60s. Doc wording only; the directive set itself is coherent.
- Values are interpolated unvalidated, so `NaN`/negative/non-integer would emit a malformed directive (RFC 7234 delta-seconds must be a non-negative integer). Harmless today since nothing passes custom numbers, but a `Math.max(0, Math.floor(...))` clamp would keep the exported util safe for future callers.

apps/cyberstorm-remix/app/communities/communities.tsx
```diff
@@ -59,46 +59,53 @@ const selectOptions = [
...
 ];
 
-export const loader = ssrLoader(async ({ request }: Route.LoaderArgs) => {
-  const url = new URL(request.url);
-  const searchParams = url.searchParams;
-  const order = searchParams.get("order") ?? SortOptions.Popular;
-  const search = searchParams.get("search");
-  const page = undefined;
-  const dapper = new DapperTs(() => {
+export const headers: Route.HeadersFunction = ({ loaderHeaders }) => {
```
Oksamies: Nit: this is the only route of the 14 where the `headers` export is placed above the loader — the other 13 place it right after. Might as well match the convention.

Oksamies (commented):
Nice clean implementation overall — the opt-in design is sound, the loader-wrapping refactor is behavior-preserving across all 14 routes, errors/redirects are correctly excluded from cache-wrapping at the loader level, and every `{ cache: true }` route is correctly paired with a `headers` export. Since all SSR loaders run anonymously (`sessionId: undefined`), nothing per-user ends up in the cache.

The substantive findings are all about how React Router v7 resolves the document `Cache-Control` across **nested routes** (`getDocumentHeadersImpl`, react-router 7.12.0): the unconditional `({ loaderHeaders }) => loaderHeaders` pattern produces some unintended cache policies. Details inline — the two worth addressing are error-document caching and the wiki create/edit pages inheriting the public cache header.

A few non-blocking notes that don't fit a single line:

- **Staleness tradeoff** — with the defaults, user-editable content (wiki/readme/changelog/versions) can be stale for ~5 min guaranteed and up to ~15 min during `stale-while-revalidate` after an edit, with no CDN purge hook. Client-side re-fetch masks this on hydrated navigation, so it mainly affects document loads (crawlers, hard refresh, pre-hydration flash). Worth confirming that window is acceptable for editor workflows.
- **Tests** — `cacheControl()` is a pure function and the new ssrLoader branch is easily unit-testable, but neither is covered (the existing `ssrLoader.test.ts` only exercises the no-cache and error paths). High-value asserts: the default header string; `isPrivate: true` output contains no `s-maxage`/`public`; `cache: true` attaches the header; a *returned* (not thrown) `Response` passes through unwrapped; a thrown ApiError Response carries no `Cache-Control`.
- **CDN config note** — `/communities` and `/dependants` vary by query params and are now `public`-cached. Standard CDNs key on the full query string, so junk params create unbounded distinct cache entries (self-bounded by the 300s TTL, so minor). Worth confirming cache-key normalization/param whitelisting on the CDN side when this ships.

Also verified fine, in case anyone else wonders about it: the `as unknown as T` cast in ssrLoader is sound — RR unwraps `DataWithResponseInit` at the single-fetch boundary on both server and client, and the existing `(await serverLoader()).seo` call in `PackageVersionVersions.tsx` keeps working correctly with `cache: true`.

PR #1861, opened 2026-06-13
Pass through data() results in ssrLoader instead of re-wrapping

Oksamies:
A cached route whose loader returns a react-router data() value — notably
gatedSsr404(), which carries its own 404 status — was being wrapped a
second time. ssrLoader's success branch only short-circuits on
`cache === false` or `result instanceof Response`, and a
DataWithResponseInit is neither, so it fell through to
`data(result, { headers: cacheControl() })`.

React Router unwraps a DataWithResponseInit exactly one level, so the
double wrap:
- dropped the inner 404 status (the outer init carried only headers), so
  the document and .data responses were served 200 and CDN-cached as a
  public, s-maxage "soft 404"; and
- nested the payload under .data, so useLoaderData() received the wrapper
  instead of the flat loader data (the gated sentinel and every field
  read as undefined).

This hit the three gated package routes (packageListing,
packageListingVersion, dependants), defeating gatedSsr's goal of
returning a real 404 document to anonymous/crawler direct entries for
hidden or nonexistent packages. The loaders stay anonymous
(sessionId: undefined) and the gated payload is public-only, so no user
data was cached — this is a status/SEO correctness regression, not a
data leak.

Fix: treat a DataWithResponseInit result like a Response in ssrLoader's
success branch and pass it through untouched. Such loaders own their own
status/headers and opt out of ssrLoader's cache wrapping, restoring the
pre-caching behavior (404 preserved, not publicly cached).

Tests:
- Add ssrLoader cases covering a data()/gatedSsr404-style result (status
  and flat shape preserved, not re-wrapped) and confirming a plain object
  result is still wrapped with public cache headers.

Refs. TS-3735

### DIFF
--- a/apps/cyberstorm-remix/cyberstorm/utils/__tests__/ssrLoader.test.ts
+++ b/apps/cyberstorm-remix/cyberstorm/utils/__tests__/ssrLoader.test.ts
@@ -1,4 +1,4 @@
-import type { HeadersArgs, LoaderFunctionArgs } from "react-router";
+import { type HeadersArgs, type LoaderFunctionArgs, data } from "react-router";
 import { describe, expect, it } from "vitest";
 
 import { ApiError } from "@thunderstore/thunderstore-api";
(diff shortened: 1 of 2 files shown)


Oksamies:
* **#1861** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1861?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1861?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Oksamies:
Changes merged as part of https://github.com/thunderstore-io/thunderstore-ui/pull/1860

ethangreen-dev (approved):

PR #1877, opened 2026-06-17
TS-3954: Disable bottom ad row, re-enable rail video ad

Oksamies:
Bottom banner ads have poor viewability, so gate the bottom-ads Island behind
a BOTTOM_ADS_ENABLED flag (default off) until we reassess; the slot config and
CSS stay intact for an easy revert, and with no container divs rendered the
in-view auction creation simply skips them.

Re-enable the rail video ad that was disabled in 5f54d944 — NitroPay has
blocked the rubiconproject requests that were causing the issue.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/Ads/nitroAds.ts
+++ b/apps/cyberstorm-remix/app/commonComponents/Ads/nitroAds.ts
@@ -264,31 +264,30 @@ export const RIGHT_COLUMN_SLOTS: RenderedAdSlot[] = [
     sizes: FISHSTICK_SIZES,
     mediaQuery: RIGHT_COLUMN_MQ,
   }),
-  // Video ad disabled — was causing issues. Re-add this slot to restore it;
-  // RAIL_VIDEO_ID, the onNavigateNimbusAds guard, and the layout.css rule are
-  // left in place so it's a one-spot revert.
-  // {
-  //   containerId: RAIL_VIDEO_ID,
-  //   sizeVariant: "video",
-  //   options: {
-  //     format: "video-nc",
-  //     demo: AD_DEMO_MODE,
-  //     refreshLimit: 0,
-  //     // 300s (5 min) auto-refresh: longer dwell for better viewability/CTR.
-  //     refreshTime: 300,
-  //     onNavigateMin: ON_NAVIGATE_MIN,
-  //     video: {
-  //       // The rail keeps the player on screen; never detach into the floating
-  //       // (docked) player, which otherwise covers the footer on tall layouts.
-  //       float: "never",
-  //     },
-  //     report: {
-  //       ...TOP_LEFT_REPORT,
-  //       icon: true,
-  //     },
-  //     mediaQuery: RIGHT_COLUMN_MQ,
-  //   },
-  // },
+  // In-content video player at the rail's bottom. Re-enabled (TS-3954) after
+  // NitroPay blocked the rubiconproject requests that were causing issues.
+  {
+    containerId: RAIL_VIDEO_ID,
+    sizeVariant: "video",
+    options: {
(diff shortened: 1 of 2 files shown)


Oksamies:
* **#1878** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1877** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

apps/cyberstorm-remix/app/root.tsx
```diff
@@ -355,7 +360,7 @@ export function Layout({ children }: { children: React.ReactNode }) {
                     </Island>
                   )}
                 </IslandContainer>
-                {shouldShowAds ? (
+                {shouldShowAds && BOTTOM_ADS_ENABLED ? (
```
Oksamies, line 363: The actual issue with bottom ads is their; visibility %, which is quite low. That's most likely because the ads take a moment to load and the user doesn't spend a lot of time at the bottom of the page. (mostly when changing the page)

Investigate if we can somehow have the ads be preloaded or warmed in a way, where they are immediately ready when the containers enter the viewport.

If we can't get the viewability up, then we remove the ad containers as it's low monetary value and keeps on taking space on the UX.

We might even want to try to reduce the height of the ad container in efforts to reduce amount of network traffic a bigger ad space might take up.

TLDR; bottom ads need balancing, not removal at this time.

(ps. would be best to test the load times on slower machines or CPU/network throttled browsers)

apps/cyberstorm-remix/app/root.tsx
```diff
@@ -64,6 +64,11 @@ import { Seo } from "./commonComponents/Seo/Seo";
 
 config.autoAddCss = false;
 
+// Bottom ad row disabled (TS-3954) due to poor viewability, pending
+// reassessment. Flip to true to bring the full-width bottom banner + its
+// companions back; the slot config and CSS are left intact for an easy revert.
+const BOTTOM_ADS_ENABLED = false;
+
```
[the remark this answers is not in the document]
Oksamies: I'm not sure if the effort and upkeep for this is valid for the gain in control. Let's just have it as hardcoded.

apps/cyberstorm-remix/app/commonComponents/Ads/nitroAds.ts
```diff
@@ -264,31 +264,30 @@ export const RIGHT_COLUMN_SLOTS: RenderedAdSlot[] = [
...
-  //     mediaQuery: RIGHT_COLUMN_MQ,
-  //   },
-  // },
+  // In-content video player at the rail's bottom. Re-enabled (TS-3954) after
+  // NitroPay blocked the rubiconproject requests that were causing issues.
+  {
+    containerId: RAIL_VIDEO_ID,
+    sizeVariant: "video",
+    options: {
+      format: "video-nc",
```
Oksamies, line 273: We should test if the floating display placement (which is a video ad to my knowledge), would be more performant. The video-nc player seems to hog quite a lot of CPU and network.

Oksamies (commented):
More work and testing needed.

apps/cyberstorm-remix/app/root.tsx
Oksamies: Let's try without bottom ads for a while and see what happens.

Roffenlund (approved):

PR #1880, opened 2026-06-17
TS-3952: Add a staff-only Django Admin link to the nav

Oksamies:
The backend already returns is_staff on the current-user payload, but zod
stripped it before the UI saw it. Carry is_staff through the user type and
both user schemas (optional + default false so cached sessions still parse),
the empty-user literal, and the fake user. Then render an "Admin" link to
/djangoadmin/ in the desktop dropdown and mobile user menu, gated on
user.is_staff — mirroring the legacy Django nav.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.tsx
+++ b/apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.tsx
@@ -7,6 +7,7 @@ import {
   faGamepad,
   faLongArrowLeft,
   faUpload,
+  faUserShield,
   faUsers,
 } from "@fortawesome/free-solid-svg-icons";
 import {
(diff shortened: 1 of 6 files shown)


Oksamies:
* **#1881** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1881?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1880** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Oksamies (commented):
Generally this is good 👍 But as noted by one copilot comment, we need to ensure that the changes don't cause errors in case the backend is updated after this PR is put into production. I'd recommend ensuring compatibility and leaving a comment for the removal of the "old" code when it's confirmed that the backend no longer uses to old fields/logic.

Edit: Comment is out-dated

VilppeRiskidev (approved):

PR #1881, opened 2026-06-17
WIP: TS-3946: Add Nimbus moderation review-queue page + bulk controls

Oksamies:
Build the /moderation/review-queue/packages page on top of the new Cyberstorm
moderation API (TS-3946 backend):
- thunderstore-api: fetchModerationReviewQueue + moderationReviewQueueBulkAction.
- A route + page listing packages awaiting review (icon, package, community,
  colour-coded review status), gated to logged-in users; non-moderators get a
  clear message (the API returns 403). Each row is selectable, with a status
  dropdown (approve / reject / unreviewed / keep in review queue) + Apply that
  calls the bulk endpoint and revalidates.
- A "Review queue" nav entry (desktop + mobile) for staff/global moderators,
  via a new ModerationReviewQueue link id.


Oksamies:
* **#1881** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1881?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1880** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1880?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.tsx
```diff
@@ -537,6 +538,22 @@ export function DesktopUserDropdown(props: {
           </NewLink>
         </NewDropDownItem>
       )}
+      {/* Moderation review queue, staff/global moderators (TS-3946). Community
```
Oksamies: Does `is_staff` include community moderators. This element should be available for community moderators (and above in permissions), not only global admins.

apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.tsx
```diff
@@ -537,6 +538,22 @@ export function DesktopUserDropdown(props: {
...
+          moderators reach it directly; the page/API enforce real permissions. */}
+      {user.is_staff ? (
+        <NewDropDownItem asChild>
+          <NewLink
```
Oksamies, line 546: Instead of linking to the review queue, we should have a "Moderation" page, which then have sub routes for the different moderation tools/topics. As there are no designs for this, try to keep the layout and visuals similar to other pages designs.

Now a important note that; sub-route does not mean the sub-routes should be in <Outlet>s. Basically just child paths. Kind of like with communities -> community.

apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.tsx
```diff
@@ -658,6 +675,19 @@ export function MobileUserMenu(props: {
         </section>
         <div className="mobile-navigation__divider" />
         <section>
+          {/* Moderation review queue, staff/global moderators (TS-3946). */}
+          {user.is_staff ? (
```
Oksamies: Same as the comment above, the /moderation route should be available for community moderators (and above) in permissions.

apps/cyberstorm-remix/app/moderation/reviewQueuePackages.tsx
```diff
@@ -0,0 +1,267 @@
...
+                checked={allSelected}
+                onChange={toggleAll}
+                aria-label="Select all"
+              />
+              <span />
+              <span>Package</span>
+              <span>Community</span>
+              <span>Status</span>
+            </div>
+            {results.map((item) => (
```
Oksamies: Instead of mapping all results in the same list, I'd recommend having a collapsible list for each status. That way the moderator can easily check the statuses they want.

apps/cyberstorm-remix/app/moderation/reviewQueuePackages.tsx
```diff
@@ -0,0 +1,267 @@
...
+      setApplying(false);
+    }
+  }
+
+  return (
+    <Page>
+      <PageHeader headingLevel="1" headingSize="2">
+        Package review queue
+      </PageHeader>
+
```
Oksamies: There should be searching and filtering of the results. The backend also needs to account for this. (e.g. we want the same capabilities as with searching for packages in the community route)

Oksamies (commented):
Let's adjust the spec a bit here. We should aim for similar usability as with PackageSearch. There needs to be a /moderation/ or /m/ route, where this "review" is a child route of.

There needs to be some consideration on the layout and implementation as we might later on add reviewing other entities like comments. This means that the /m/review page needs to probably have sub-path of "p". So "/m/review/p" or "/moderation/review/package".

Then there should also be a way to open the singular package that is for review, in a modal in the page. As the package might have internal comments or other critical information. The reason we shouldn't force users to open the normal package page is that, we want to nicely separate the moderation features from the main publicly available pages. (separation of concerns/implementation)

This needs quite a bit more work and thought so better do that before merge.

Also please keep in mind that the backend implementation needs to match the frontend and be secure. (e.g. no caching)

Oksamies:
Implementation and specs out of date, features included in this will be done as part of a bigger feature addition.

PR #1883, opened 2026-06-17
TS-3949: Keep modals within the viewport at large display scaling

Oksamies:
.modal__content was viewport-centered but had no max-height, and only the body
had a fixed 800px cap. At high Windows display/font scaling the title + body +
footer exceeded the screen and spilled equally above and below, pushing the
footer (e.g. a moderator's reject button) off-screen with no way to scroll.
Cap the content at the viewport height with a small margin and let the body
shrink (min-height: 0) so it, not the whole modal, scrolls.

### DIFF
--- a/packages/cyberstorm/src/newComponents/Modal/Modal.css
+++ b/packages/cyberstorm/src/newComponents/Modal/Modal.css
@@ -28,9 +28,15 @@
     min-width: 15rem;
     max-width: 39rem;
 
+    /* Never taller than the viewport (leave a vertical top/bottom margin), so
+       the title and footer stay on screen at any display/font scaling; the
+       body scrolls within (TS-3949). */
+    max-height: calc(100vh - 4rem);
+
     /* padding: var(--modal-content-padding); */
     border: var(--modal-border);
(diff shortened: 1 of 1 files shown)


Oksamies:
* **#1883** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1883?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Oksamies (commented):
Comment from @gkurck :
The way modal css is done is unnecessarily overcomplicated. Since everything is under a wrapper, you'll only need to set position:fixed for the wrapper since all children will be inside that stacking context now. Currently you also set position fixed for the modal itself which makes it not respect the wrapper boundaries, and also for the close button.

Let the wrapper handle positioning and sizing/padding constraints, the modal can just chill inside the wrapper

VilppeRiskidev (approved):

PR #1887, opened 2026-06-17
TS-3944: Add a fullscreen/expand control to the Analysis codebox

Oksamies:
The Analysis tab's code viewer (CodeBoxHTML) could only be read in a 600px
scroll box. Add an Expand button that opens the same code in a fullscreen modal
inset from the window edges by a uniform margin (matching the page gutter),
with a close button and Esc/overlay-click to dismiss. Wires up the previously
unimported CodeBoxHTML.css and reuses the shared Modal; no shared CodeBox
changes.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/CodeBoxHTML/CodeBoxHTML.css
+++ b/apps/cyberstorm-remix/app/commonComponents/CodeBoxHTML/CodeBoxHTML.css
@@ -4,3 +4,93 @@
   width: 100%;
   overflow: auto;
 }
+
+/* nimbus-components wins over the `components` layer that owns .modal__content,
+   so this fullscreen variant reliably overrides the centered modal sizing. */
+@layer nimbus-components {
+  .code-box-html__container {
+    position: relative;
+    width: 100%;
+  }
+
+  /* Toolbar overlaid on the box's top-right corner (inside the border). Lives on
+     the non-scrolling container so it stays put while the code scrolls. Buttons
+     sit in a row, copy first, all solid (non-ghost). */
+  .code-box-html__toolbar {
+    position: absolute;
+    top: var(--space-8);
+    right: var(--space-8);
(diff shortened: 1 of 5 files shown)


Oksamies:
* **#1887** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Oksamies (commented):
- The expand button should not be a ghost button.
- The copy button should be the same kind as the expand button. (and copy button should be the first button)
- The download button should should have the file ending instead of "Download" e.g. ".blob" 
- The component is missing siderail of line numbers
- There should be animation of the "expanded" modal sliding in from the bottom of the viewport and going back down when unexpanded.
- On non-mobile widths the "expanded" modal/element should have margin around 10% margin on both top and bottom. (e.g. so that it's not filling the whole screen and the user can dismiss by clicking on the area below or above the modal/card)
Image

Chaosyr:
> - The expand button should not be a ghost button.
> - The copy button should be the same kind as the expand button. (and copy button should be the first button)
> - The download button should should have the file ending instead of "Download" e.g. ".blob" 
> - The component is missing siderail of line numbers
> - There should be animation of the "expanded" modal sliding in from the bottom of the viewport and going back down when unexpanded.
> - On non-mobile widths the "expanded" modal/element should have margin around 10% margin on both top and bottom. (e.g. so that it's not filling the whole screen and the user can dismiss by clicking on the area below or above the modal/card)
> Image
> 

That needs a overflow-x: scroll :p 

I can tell because currently the code block is overflowing out of its container ^^

PR #1894, opened 2026-06-17
Remove unnecessary "Go to main content element"

Oksamies:


Oksamies:
* **#1894** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Oksamies (commented):
The Copilot comments are valid, but for this feature to work properly we would need to; structure the pages better so that the skip-to-content actually takes the user to the correct position in page. Right now on the communities page it takes the user to half way down the communities list visually and the keyboard navigation with tab still starts from the header navigation.

Either remove the feature as this PR does, or implement it properly.

VilppeRiskidev (approved):

Chaosyr:
> The Copilot comments are valid, but for this feature to work properly we would need to; structure the pages better so that the skip-to-content actually takes the user to the correct position in page. Right now on the communities page it takes the user to half way down the communities list visually and the keyboard navigation with tab still starts from the header navigation.
> 
> Either remove the feature as this PR does, or implement it properly.

Okay I was gonna say that this is a accessibility issue, but if thats the case it seems fine to remove for the time being.

PR #1896, opened 2026-06-18
Wiki and Analysis tabs are disabled by default and enabled once content exists

ebkr:
Note: I'm unable to test package analysis locally because it doesn't run in dev.

If we have a setup, please try to verify it.

# Screenshots

## User has no edit permission and wiki doesn't exist

image

## User has edit permissions and wiki does not exist

image

## User has no edit permissions and wiki exists

image

## User has edit permissions and wiki exists

image

### DIFF
--- a/apps/cyberstorm-remix/app/p/packageListing.tsx
+++ b/apps/cyberstorm-remix/app/p/packageListing.tsx
@@ -50,6 +50,7 @@ import {
 } from "@thunderstore/cyberstorm";
 import { PackageLikeAction } from "@thunderstore/cyberstorm-forms";
 import { DapperTs, type DapperTsInterface } from "@thunderstore/dapper-ts";
+import { isApiError } from "@thunderstore/thunderstore-api";
 
 import type { Route } from "./+types/packageListing";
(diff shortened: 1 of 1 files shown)


apps/cyberstorm-remix/app/p/packageListing.tsx
```diff
@@ -283,7 +286,42 @@ export default function PackageListing() {
...
+        setHasWiki(res.pages.length > 0);
+      })
+      .catch(() => setHasWiki(false));
+
+    dapper
+      .getPackageSource(listing.namespace, listing.name)
+      .then((res) => {
+        setHasSource((res?.decompilations?.length ?? 0) > 0);
+      })
+      .catch(() => setHasSource(false));
```
Oksamies: Do all of this logic in the clientLoader preferrably, as the permissions are already fetched there etc.

Oksamies (changes requested):
The goal of the fix is good, but the implementation is not inline with how other similar stuff is done in the project. Please see how similar checks, state management and passing of state are done elsewhere 👍

PR #1897, opened 2026-06-18
Pagination behaviour(TS-3378 )

Roffenlund:
- Refactor: replace the imperative buttons.push() array with declarative
  JSX and small PageButton/NavButton sub-components. No behavior change.
- Ellipsis: sibling buttons now always show their real page number, and
  "…" is a separate, non-clickable indicator shown only when pages are
  actually skipped — previously a button labeled "…" would jump to a
  specific adjacent page, and could appear even with no real gap.
- Scroll: page changes now jump to the top instantly instead of being
  animated by the global smooth-scroll.


Roffenlund:
Before:
image

After:
image

Oksamies (approved):
Other than AI comments, looks good to me 👍

PR #1900, opened 2026-06-18
Fix: Table Fix

Chaosyr:
This commit fixes the table issue seen in the Cyberstorm version of the site that causes tables to make the entire markdown render shift on horizontal scroll

image

image

This will make the tables look like the above rather than the current below

image

Do note its not perfect but it does fix major UX issues when looking at a packages README. Main thing I'd still want fixed regarding this is wider column text width. 

Reference Page: https://thunderstore.io/c/inscryption/p/creator/inscryption_augmented_to_base/

If this does not match conventions feel free to rebase or squash merge this into the repository.

### DIFF
--- a/apps/cyberstorm-remix/app/styles/markdown.css
+++ b/apps/cyberstorm-remix/app/styles/markdown.css
@@ -1,6 +1,5 @@
 @layer nimbus-utils {
   .markdown-wrapper {
-    display: flex;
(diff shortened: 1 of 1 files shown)


Oksamies:
Closed in favor of https://github.com/thunderstore-io/thunderstore-ui/pull/1902 which adds one more commit.

PR #1916, opened 2026-06-24
Provide a Valid Image in the Placeholder, and an Image that purposefully forces Alt-Text.

Chaosyr:
* Replaces the Image piece with a valid sample link, and a purposefully non-valid one explaining why its not valid.

Simple PR that just updates the sample to show valid image handling and invalid.

Prolly could be updated to be better still, not entirely sure how the one lined placeholder works as I'd usually use a multiline string for something like that (maintainability purposes) 

But it should work properly.

Utilizes the test catheim image lurking in this GitHub Repository.


Chaosyr:
This doesn't feel as needed anymore. Thus I'm closing this PR.

PR #1917, opened 2026-06-25
Fix upload page blocking valid packages over nested files

Oksamies:
The pre-upload zip validation raised "files are case-sensitive" and
"wrong file extension" errors whenever a manifest.json/icon.png/README.md
(or an icon.jpg/readme.txt) appeared in any case ANYWHERE in the archive
— e.g. a bundled dependency under plugins/ — even when the root files were
correct. This blocked legitimate uploads.

Cause: the port from the legacy Django upload page changed the casing,
wrong-extension and extension-less checks from root-only (full-path match)
to leaf-based, so they fired for nested entries. The server only reads the
root files via exact zipfile lookups (unzip.read("manifest.json"), etc.),
so nested files never affect it; the client was stricter than the server.

Gate the casing, wrong-extension and extension-less checks on the archive
root. Presence detection stays leaf-based so the "files not at root" and
"missing file" guidance still works.

Tests: replace the two cases that encoded the nested-detection behavior
with cases asserting nested copies are ignored; add table-driven coverage
for mixed-case names/extensions at the root and nested copies.

### DIFF
--- a/apps/cyberstorm-remix/app/upload/__tests__/uploadZipValidation.test.ts
+++ b/apps/cyberstorm-remix/app/upload/__tests__/uploadZipValidation.test.ts
@@ -174,29 +174,169 @@ describe("evaluateZipContents", () => {
     );
   });
 
-  it("detects wrong-case required files inside a subfolder", () => {
+  it("does not flag wrong-case copies nested in a subfolder when the root files are correct", () => {
+    // A BepInEx-style package: correctly cased files at the root, plus a
+    // bundled dependency under plugins/ whose own files use arbitrary casing.
+    const result = evaluateZipContents("package.zip", [
+      "manifest.json",
+      "icon.png",
+      "README.md",
+      "plugins/Manifest.json",
+      "plugins/Readme.md",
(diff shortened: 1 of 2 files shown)


Oksamies:
* **#1917** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

VilppeRiskidev (approved):
Ok 👍

Chaosyr (approved):

x753 (dismissed):
This works for nested files with the wrong case or extension, but we can also have files at the root with the wrong extension. The logic needs to check if the correct case and extension files exist in the correct place, and if they do we don't care if there are also incorrect files.

Chaosyr (approved):

PR #1919, opened 2026-06-26
Shrink Table Header Min Width

Chaosyr:
This makes tables with images behave better, and tables feel more compacted while still looking visually nice.

image

image

image

image

If users want longer headers or the column to take more space, they can give it a longer header name to do so.

If pages got a pinch more width it would prolly feel even nicer.


Chaosyr:
This one seems to be implemented not 100% how

image

But this does seem fine to close.

PR #1920, opened 2026-06-26
Add Overflow-Wrap Anywhere to CardPackage.css Description

Chaosyr:
* Add Overflow Wrap to Card Package Description

This appears to be the correct file for adding the given change based on the sites live css.

Should look like this live if all went well

image

and not

image


Chaosyr:
This was fixed, closing ^^

PR #1921, opened 2026-06-29
Make Wiki.css Respect Markdown.css

Chaosyr:
The Wiki Content which uses a markdown renderer was wrapped in a `Flex`, which broke the Wiki from following the `markdown.css` used site wide, now it works. (I made it utilize `Grid` specifically, as its the one that doesn't break overflow wrapping. `Block` does not work here.)

Before:
image

image

After:
image

image


Chaosyr:
This issue appears to have been fixed at some point, I'll mark this with a close.

PR #1928, opened 2026-07-17
Move the package rejection notice into the banner beside the notes

Oksamies:
The rejection reason rendered in the sidebar management block, floating above the moderation tools row — an awkward spot. Show it like the mod internal notes instead: a collapsible card overlaid on the community banner, laid out side by side with the notes (rejection first). A shared BannerNote card backs both.

Side by side, each card keeps its preview and clamps to the banner height, so neither an owner's single rejection card nor a mod's rejection + notes overhang the content below. Expanding a card grows it over the content, like the notes already did. On single-column widths the cards drop into normal flow, stacked.


ebkr (approved):
LGTM - Not tested locally but changes seem sound

PR #1932, opened 2026-07-27
Load the syntax highlighter lazily in CodeBox

Oksamies:
CodeBox is re-exported from the package barrel, so its static import of
react-syntax-highlighter landed in the shared cyberstorm chunk that every
route downloads. The component is only used on the package Source tab.

Splitting the highlighter into its own module and importing it after
mount drops the shared chunk from 342 KiB to 296 KiB raw (~103 KiB to
93 KiB gzipped); the highlighter becomes a 15 KiB gzipped chunk that only
the Source tab pays for.

Until it loads, the code renders as plain <pre> using the same inline
style, so the box geometry is identical and only the token colours
change. That also means SSR emits real code text rather than nothing.

### DIFF
--- a/packages/cyberstorm/src/components/CodeBox/CodeBox.tsx
+++ b/packages/cyberstorm/src/components/CodeBox/CodeBox.tsx
@@ -1,8 +1,13 @@
-import { LightAsync as SyntaxHighlighter } from "react-syntax-highlighter";
-import nightOwl from "react-syntax-highlighter/dist/esm/styles/hljs/night-owl";
+import {
+  type CSSProperties,
+  type ComponentType,
+  useEffect,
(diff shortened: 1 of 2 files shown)


Oksamies:
* **#1932** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1932?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1933, opened 2026-07-27
Inject the ad script during idle instead of on load

Oksamies:
ads-785.js pulls in the whole auction stack (GPT, prebid, Confiant,
Amazon, btloader) and measured ~1.7s of script execution. Injecting it
the instant `load` fired put all of that on top of hydration.

`load` is not late enough on its own: total blocking time is measured
until the main thread goes quiet, so the auction still ran inside that
window. Blocking the ad domains in Lighthouse took the community page
from 43 to 73 and TBT from 2625ms to 148ms, which is what this targets.

Slot creation already waited for idle; this applies the same pattern one
step earlier, to the script itself, and lifts the duplicated
requestIdleCallback/setTimeout/cleanup dance out of both call sites into
scheduleWhenIdle so the two cannot drift. The timeout bounds how long a
busy page can delay the first impressions.

Idle-gating only, deliberately: a fixed delay on top would move the lab
metric without reducing the work, while pushing the first impression out
of short visits and shifting every slot refresh by the same amount.

Ads will appear somewhat later than before, so this wants checking
against impression and viewability numbers on QA before it ships.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/Ads/__tests__/scheduleWhenIdle.test.ts
+++ b/apps/cyberstorm-remix/app/commonComponents/Ads/__tests__/scheduleWhenIdle.test.ts
@@ -1,0 +1,127 @@
+import { afterEach, describe, expect, test, vi } from "vitest";
+
+import { IDLE_FALLBACK_DELAY_MS, scheduleWhenIdle } from "../scheduleWhenIdle";
+
+// These run in a real browser, so requestIdleCallback exists; the Safari path is
+// covered by deleting it for the duration of a test. The DOM lib declares both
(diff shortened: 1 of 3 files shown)


Oksamies:
* **#1933** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1933?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1934, opened 2026-07-27
Load the first package card's icon eagerly at high priority

Oksamies:
The LCP element on listing pages is the first card's icon, and Image
defaults every image to loading="lazy" with no way to opt out, so the
browser only discovered it after layout. Lighthouse failed both of its
LCP checks on it: not eagerly loaded, and no priority hint.

Image gains a fetchPriority prop, CardPackage a `priority` flag, and the
listing grid sets it on index 0. Only the first card: fetchPriority is a
ranking hint, so marking a whole row would dilute it and put several
large icons in contention on mobile.

Confirmed against a local production build -- the first card renders
loading="eager" fetchPriority="high", the other 11 stay lazy. The LCP
element was verified with a PerformanceObserver rather than assumed:
community pages resolve to the card icon, package pages to a text span,
which is why the hero banners are deliberately left alone.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/PackageSearch/PackageSearch.tsx
+++ b/apps/cyberstorm-remix/app/commonComponents/PackageSearch/PackageSearch.tsx
@@ -525,10 +525,17 @@ export function PackageSearch(props: Props) {
                 <>
                   {resolvedValue.results.length > 0 ? (
                     <div className="package-search__grid">
-                      {resolvedValue.results.map((p) => (
+                      {resolvedValue.results.map((p, index) => (
                         <CardPackage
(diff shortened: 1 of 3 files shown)


Oksamies:
* **#1934** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1934?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1934?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1936, opened 2026-07-27
Keep the package tab title after hydration

Oksamies:
clientLoader.hydrate reruns the client loader on load and replaces the
match data the <Seo> head reads. The package clientLoaders returned no
`seo`, so once hydration swapped their data in, the title collapsed from
the SSR value to the root "Thunderstore" fallback.

Both loaders now build the title from the already-awaited `listing`
(name, namespace, community_name), so the client path emits the same
tags as SSR and the title no longer flips — at no extra request, since
`listing` is resolved either way. The format is also updated to the
requested "<package> by <author> | <community> | Thunderstore" (and the
versioned page's "<package> v<version> by <author> | ..."), shared
between the two loaders so they can't drift.

Community and package-search pages have the same underlying gap, but
their clientLoaders don't hold the community display name without
awaiting a deferred fetch, so they're left for a follow-up rather than
reintroducing that await.

### DIFF
--- a/apps/cyberstorm-remix/app/p/packageListing.tsx
+++ b/apps/cyberstorm-remix/app/p/packageListing.tsx
@@ -81,6 +81,44 @@ type PackageListingOutletContext = OutletContextShape & {
   packageDownloadUrl?: string;
 };
 
+type ResolvedListing = NonNullable<
+  Awaited<ReturnType<typeof getPublicListing>>
+>;
+
+// Browser-tab title + share metadata, shared by the SSR loader and the
+// clientLoader. clientLoader.hydrate reruns the client loader on load and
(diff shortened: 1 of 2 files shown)


Oksamies:
* **#1936** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1936?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1936?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1937, opened 2026-07-27
Render <details> collapsibles in wiki markdown

Oksamies:
The wiki renders raw markdown through react-markdown, which drops
embedded HTML unless rehype-raw is in the chain — so the <details>/
<summary> collapsibles the old site showed vanished. Add rehype-raw
before nimbusSanitize: raw parses the HTML into real nodes, then the
existing strict allowlist (which already permits details/summary/open
and strips <script>) cleans them, so raw HTML is parsed then sanitized,
never trusted.

rehype-raw parses whatever a wiki author wrote, which makes the sanitiser
the only thing between them and script execution — so cover it. The new
tests render the real component (not a reconstruction of its plugin
chain, which could drift from it) against 17 vectors: script tags, event
handlers, javascript: URLs, iframe srcdoc, svg onload, object/embed,
base/meta/link, template>script, unclosed tags and comment breakout. All
are neutralised; inverting the plugin order fails four of them.

Conversion also walks the document recursively, so an unusually
structured one can exhaust the call stack partway through and cost the
reader the page rather than just its formatting. MarkdownErrorBoundary
catches that and renders the source as plain text instead — more robust
than enumerating the shapes that cause it, since what matters is the
failure rather than its cause. It resets per document, so one that
cannot be converted doesn't affect the next.

react-markdown has to join optimizeDeps.include or Vite pre-bundles it
with a second copy of React and every render throws on a null dispatcher.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/Markdown/Markdown.css
+++ b/apps/cyberstorm-remix/app/commonComponents/Markdown/Markdown.css
@@ -5,6 +5,17 @@
     scrollbar-width: none;
   }
 
+  /* Shown when a document fails to convert (see MarkdownErrorBoundary). The
+     source is still readable, so wrap it rather than letting long lines force
+     the page wide. */
+  .markdown__fallback {
+    margin: 0;
+    color: var(--color-text-primary);
+    font-family: var(--font-family-monospace);
+    white-space: pre-wrap;
+    overflow-wrap: anywhere;
+  }
+
   .markdown {
     display: block;
     color: var(--color-text-primary);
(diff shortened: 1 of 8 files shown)


Oksamies:
* **#1937** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1938, opened 2026-07-27
Sort, cap, and filter the nav teams dropdown

Oksamies:
The user's teams arrived in the API's arbitrary order and the desktop
submenu had no height limit, so many teams ran off the screen with no
way to scan them.

- Sort alphabetically (case-insensitive), desktop and mobile.
- Cap the desktop submenu height and give the team list its own
  scrollbar, keeping the filter box and "All Teams" pinned.
- Past 8 teams, show a lightweight client-side filter. Its wrapper stops
  keydown propagation so Radix's menu typeahead doesn't hijack the input.

Pinning is load-bearing rather than cosmetic: the height cap makes every
child a shrinkable flex item, so without it flexbox squeezed the filter
box and "All Teams" alongside the list, and .dropdown's overflow:hidden
clipped the result instead of revealing it — the search field rendered at
24px against 42px of content and sat underneath "All Teams". Only the
list should absorb the cap.

The sort and filter live in teams.ts so they can be tested without
mounting the whole navigation tree.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.css
+++ b/apps/cyberstorm-remix/app/commonComponents/Navigation/Navigation.css
@@ -141,6 +141,69 @@
     min-width: 14rem;
   }
 
+  /* Teams submenu: cap the height so a user in many teams gets a scrollable
+     list instead of one that runs off the bottom of the screen. The filter box
+     and "All Teams" link stay pinned; only the team list scrolls. */
+  .navigation-header__teams-dropdown {
+    min-width: 14rem;
+    max-height: min(
+      24rem,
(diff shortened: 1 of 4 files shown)


Oksamies:
* **#1938** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1938?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1938?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):
Soft approve. Check AI comments.

PR #1939, opened 2026-07-27
Render the package source without re-highlighting it client-side

Oksamies:
Opening the Source tab on a package with large decompilations pinned the
CPU at 100% and left the page unusable. BepInExPack_Inscryption is the
reported case: 12 decompilations, 5.97MB of HTML carrying 0.89MB of
actual code, 229,651 spans.

Two things were wrong.

The API already returns Pygments-highlighted HTML — and CodeBoxHTML
stripped every span back out with a regex, then re-highlighted the plain
text client-side with highlight.js. That measured ~700ms of blocking work
per file, so ~9s for that package. The .highlight stylesheet for the
server's classes was already in the bundle and already imported; it was
just unused, because the markup it styles was being thrown away.

Everything was mounted at once, so even rendering the server's HTML
directly cost 229k DOM nodes, 2.3s to parse and 5.6s of layout — and kept
costing, since every later layout and paint had to walk it. That is the
part that made the CPU stay busy rather than merely spike.

So: render the server's highlighting, and only build DOM for the rows in
view. Measured on the same payload, all 12 files now render in 145ms with
4.1k nodes on the whole page, and a scroll step costs ~0.1ms. No size
gate and no click-to-reveal — files of any size open and scroll, with
line numbers and a sticky gutter.

The decompiled code derives from uploaded packages, so the markup is not
simply trusted: it is checked against the exact grammar Pygments emits
and falls back to plain text with a visible notice if anything else
appears. A full sanitiser over several MB would cost more than the
rendering does. The permitted class is pinned to the token set
Highlight.css actually styles rather than any word — a class this
component does not style has no reason to reach the DOM, and allowing
arbitrary ones would let package contents borrow the site's own class
names to restyle text inside the viewer. All 229,651 spans in the sampled
payload use classes from that set.

Line count is capped with a visible notice: past ~1.7M rows the surface
would exceed the browser's maximum element height and the tail would
silently stop being reachable. The API truncates long results well before
that, so it is a backstop against failing quietly.

### DIFF
--- a/apps/cyberstorm-remix/app/commonComponents/CodeBoxHTML/CodeBoxHTML.css
+++ b/apps/cyberstorm-remix/app/commonComponents/CodeBoxHTML/CodeBoxHTML.css
@@ -1,6 +1,63 @@
-.code-box-html {
-  position: relative;
-  align-self: stretch;
-  width: 100%;
-  overflow: auto;
+@layer nimbus-components {
+  .code-view {
+    display: flex;
+    flex-direction: column;
+    gap: var(--gap-sm);
+    align-self: stretch;
+    width: 100%;
+    min-width: 0;
+  }
+
+  /* The scroll port. Both axes live here: rows are absolutely positioned on the
+     surface inside, which is sized to the full line count and the widest line. */
+  .code-view__scroll {
+    position: relative;
+    width: 100%;
+    padding: var(--space-8) 0;
+    border: 1px solid var(--color-surface-8);
+    border-radius: var(--radius-md);
+    overflow: auto;
+    font-size: var(--font-size-body-md);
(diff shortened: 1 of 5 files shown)


Oksamies:
* **#1939** <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1939?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/thunderstore-io/thunderstore-ui/1939?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

Roffenlund (approved):

PR #1952, opened 2026-09-01
Editable package markdown

ethangreen-dev:
This PR adds the ability for users to edit README.md and CHANGELOG.md files without submitting a new package version.

Hard dependency on https://github.com/thunderstore-io/Thunderstore/pull/1239

image
The author view on the new last edited timestamp and edit button.

image
Mobile view

image
Versions list with minimal edited pill.

image
Upload page option to use the previous version's readme or download it.

image
The readme editor.

image
And the same editor for a failure state.

image
Mobile of the same view. Squashed but maybe usable.


Chaosyr:
Have a question, does the site edited version of the README/CHANGELOG get saved to that versions zip, or is it stored next to the package? /cur

ethangreen-dev:
> Have a question, does the site edited version of the README/CHANGELOG get saved to that versions zip, or is it stored next to the package? /cur

Stored in the DB, not in the package archive. We architecturally cannot easily edit package contents, so this is the best option for now.

Chaosyr:
> > Have a question, does the site edited version of the README/CHANGELOG get saved to that versions zip, or is it stored next to the package? /cur
> 
> Stored in the DB, not in the package archive. We architecturally cannot easily edit package contents, so this is the best option for now.

Got it, thanks for the clarification!!

packages/thunderstore-api/src/get/packageReadme.ts
```diff
@@ -20,6 +20,9 @@ export async function fetchPackageReadme(
     args: {
       config: config,
       path: path,
+      // Keep the browser HTTP cache out of the way so a site edit is
+      // visible on the next navigation. Shared caches still apply.
+      request: { cache: "no-store" },
```
Oksamies, line 25: Wouldn't this essentially make it so that all requests to these endpoints always hit the cdn? I'd prefer not adding these no-stores, so that we dont lose the benefits of browser caching.

It's not like the browser cache is that long either.

packages/thunderstore-api/src/get/packageVersionMarkdownRaw.ts
```diff
@@ -0,0 +1,57 @@
...
+  package: string;
+  version: string;
+}
+
+/**
+ * Fetches the resolved raw markdown of a version's README or CHANGELOG from
+ * the experimental API. Serves the override when one exists, the packaged
+ * file otherwise, which makes it the right source for prefilling the editor.
+ */
+export function fetchPackageVersionMarkdownRaw(
```
Oksamies, line 17: Let's not have one function for both readme and changelog. Separate these into their own functions "fetchPackageVersionReadmeMarkdownRaw" and "fetchPackageVersionChangelogMarkdownRaw".

Basically no reason to tie these two resources together at Nimbus' end.

packages/thunderstore-api/src/post/packageVersionMarkdown.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { ApiEndpointProps } from "../index";
+import {
+  type PackageVersionMarkdownRequestData,
+  type PackageVersionMarkdownRequestParams,
+  packageVersionMarkdownRequestDataSchema,
+} from "../schemas/requestSchemas";
+import { packageVersionMarkdownResponseDataSchema } from "../schemas/responseSchemas";
+import type { PackageVersionMarkdownResponseData } from "../schemas/responseSchemas";
+
+export function postPackageVersionMarkdown(
```
Oksamies, line 11: I'm assuming this fetches the readme? Correct the function naming and possibly the backend path too, so that it's not just "markdown". (ref. https://github.com/thunderstore-io/thunderstore-ui/pull/1952/changes#r4070508846)

packages/thunderstore-api/src/schemas/responseSchemas.ts
```diff
@@ -177,8 +179,37 @@ export type PackageSourceResponseData = z.infer<
...
+
+export type PackageVersionMarkdownResponseData = z.infer<
+  typeof packageVersionMarkdownResponseDataSchema
+>;
+
+// PackageVersionRawMarkdownResponse (experimental raw markdown endpoints)
+export const packageVersionRawMarkdownResponseDataSchema = z.object({
+  markdown: z.string().nullable(),
+  is_edited: z.boolean().optional().default(false),
+  edited_at: z.string().nullable().optional().default(null),
```
Oksamies, line 206: These edited_at fields should probably be datetimes?

apps/cyberstorm-remix/vitest.config.ts
```diff
@@ -9,6 +9,8 @@ export default defineProject({
   resolve: {
     alias: {
       cyberstorm: cyberstormRoot,
+      app: path.join(projectDir, "app"),
```
Oksamies, line 12: Why were these needed?

Oksamies (changes requested):
Answer questions at the very least

packages/thunderstore-api/src/get/packageReadme.ts
ethangreen-dev: Fair, cache only saved like 60 seconds so it's not necessary. Removed from everything but the editor queries

packages/thunderstore-api/src/get/packageVersionMarkdownRaw.ts
ethangreen-dev: splitty split split split

packages/thunderstore-api/src/post/packageVersionMarkdown.ts
ethangreen-dev: save endpoint, not fetch, but the name doesnt make that clear. fixed

packages/thunderstore-api/src/schemas/responseSchemas.ts
ethangreen-dev: ya, they all use datetime() now

apps/cyberstorm-remix/vitest.config.ts
ethangreen-dev: it was a hack, now the hack has been replaced.

